PR #19, opened 2012-08-16
Feature parameter validation via jsr

philipsommer:


raphaelbauer:
Awesome!
That is really really cool :)

PR #20, opened 2012-08-16
Content negotiation broken when Result has no content type set.

roughy:
The Resulthandler uses now the Accept Header to set the response content type


raphaelbauer:
very cool... :)
Thanks!
ra

PR #58, opened 2012-10-12
Feature/error objects in validation

philipsommer:
Some enhancements in validation :-)

### DIFF
--- a/ninja-core/src/main/java/ninja/validation/Validation.java
+++ b/ninja-core/src/main/java/ninja/validation/Validation.java
@@ -80,10 +80,27 @@ public interface Validation {
      */
     String getFieldViolationMessage(String field, String language);
 
+    /**
+     * Get a complete list of all field violations. This list DOES NOT contain general violations
+     * (use getGeneralViolations() instead).
+     * 
+     * @return A List of FieldViolation-objects
+     */
     List<FieldViolation> getFieldViolations();
 
+    /**
+     * Get all general constraint violations. This list does not contain any specific field
+     * violation (use getFieldViolations() instead).
+     * 
+     * @return The list of general violations.
+     */
     List<ConstraintViolation> getGeneralViolations();
 
+    /**
+     * Add a field violation to the list of filed violations.
+     * 
+     * @param fieldViolation
+     */
     void addFieldViolation(FieldViolation fieldViolation);
 
 }
--- a/ninja-core/src/main/java/ninja/validation/Validation.java
+++ b/ninja-core/src/main/java/ninja/validation/Validation.java
@@ -16,9 +16,9 @@
 
 package ninja.validation;
 
-import com.google.inject.ImplementedBy;
+import java.util.List;
 
-import java.util.Locale;
+import com.google.inject.ImplementedBy;
 
 /**
  * Validation context
@@ -67,7 +67,7 @@ public interface Validation {
      * @return The constraint violation, or null if no constraint violation was
      *         found
      */
-    ConstraintViolation getFieldViolation(String field);
+    ConstraintViolation getFieldConstraintViolation(String field);
 
     /**
      * Get the formatted violation message for the given field
@@ -79,5 +79,11 @@ public interface Validation {
      * @return The message, or null if there was no violation
      */
     String getFieldViolationMessage(String field, String language);
+
+    List<FieldViolation> getFieldViolations();
+
+    List<ConstraintViolation> getGeneralViolations();
+
+    void addFieldViolation(FieldViolation fieldViolation);
 
 }
--- a/ninja-core/src/main/java/ninja/validation/FieldViolation.java
+++ b/ninja-core/src/main/java/ninja/validation/FieldViolation.java
@@ -1,0 +1,11 @@
+package ninja.validation;
+
+public class FieldViolation {
+    public FieldViolation(String field, ConstraintViolation constraintViolation) {
+        this.field = field;
+        this.constraintViolation = constraintViolation;
+    }
+
+    public String field;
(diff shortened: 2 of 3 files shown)


ninja-core/src/main/java/ninja/validation/Validation.java
```diff
@@ -80,4 +80,27 @@
...
+     */
+    List<ConstraintViolation> getGeneralViolations();
+
+    /**
+     * Add a field violation to the list of filed violations.
+     * 
+     * @param fieldViolation
+     */
+    void addFieldViolation(FieldViolation fieldViolation);
+
```
christiangoudreau, line 105: remove line 105

ninja-core/src/main/java/ninja/validation/Validation.java
```diff
@@ -80,4 +80,27 @@
...
+    /**
+     * Get all general constraint violations. This list does not contain any specific field
+     * violation (use getFieldViolations() instead).
+     * 
+     * @return The list of general violations.
+     */
+    List<ConstraintViolation> getGeneralViolations();
+
+    /**
+     * Add a field violation to the list of filed violations.
```
christiangoudreau, line 100: The method itself is pretty much self explanatory.

ninja-core/src/main/java/ninja/validation/Validation.java
```diff
@@ -80,4 +80,27 @@
...
+    /**
+     * Get a complete list of all field violations. This list DOES NOT contain general violations
+     * (use getGeneralViolations() instead).
+     * 
+     * @return A List of FieldViolation-objects
+     */
+    List<FieldViolation> getFieldViolations();
+
+    /**
+     * Get all general constraint violations. This list does not contain any specific field
```
christiangoudreau, line 92: The second part of the comment add value to the method names which could be renamed getAllGeneralConstraintViolations.

christiangoudreau:
LGTM, just one general comment on the documentation and it's more about the orientation that this project wants to take concerning documentation.

Personnally I like to write self explanatory code and when I add comments, it's only to add information that the user may not understand by only reading the method name. Since this is a framework, more is better than less, but that also means that we need to maintain both the documentation and the code.

I'd like to see Raphael inputs on that since it will influence the way I review that kind of subject in the future.

raphaelbauer:
Hi @christiangoudreau and @philipsommer,

first of all thanks for the contribution :) It is really cool to see that the framework gets better step by step. And @christiangoudreau many thanks for reviewing the code.

Regarding comments, code style and many other things: I think Christian's comment points into the right direction: The whole human setup of the project is not yet right. Currently I am the single point of failure. Plus - there are not many "rules" how to do things. I tend to like that (not having rules), but at some points there should be some constraints and guidelines. 

Regarding the actual comments I totally agree that keeping documentation in two places (method + javadoc + ...) can be problematic. But to me it is even more problematic not having a human readable documentation in addition to method names. I really like it when I can use Eclipse to hoover over a method getting a nice description showing what the method actually does (I even like it more when there is an example showing something meaningful). But I agree with @christiangoudreau that there are cases where this might not be the best way to do it...

Therefore: The next big steps for ninja will be to bring more people into the boat and replace "Raphael" with something more meaningful ;) This will happen soon - and I'd really invite you both to join the party :) 

Cheers and many thanks again!

Raphael

philipsommer:
Hi Christian, hi Raphael,

thank you guys for your reviews and comments.
The reason I made some extra documentations on the methods is, that field violations and general violations don't have any intersection (the user could think, field violations are maybe included in general violations). For me it seems important for the user to know. Generally, Christian, you're completely right, code should be self-explanatory, I will take more care of that in the future.

@Rafael: I would like to implement a new feature to ninja in the near future. This feature should help us to unify http-error codes and what the api callee should see from the inner life of ninja (now, if an error occurs, the callee sees the exception stack trace). So, my thoughts went in the direction to implement (additional to the yet working filter-Annotation) an annotation for post-filters. Those filters should be able to catch exceptions and modify the result of the controller function. This also could be a first step to apply a simple caching mechanism. Please let me know about your thoughts.

Wish you both a nice day,

best, Philip

Am 15.10.2012 um 12:10 schrieb reyez:

> Hi @christiangoudreau and @philipsommer,
> 
> first of all thanks for the contribution :) It is really cool to see that the framework gets better step by step. And @christiangoudreau many thanks for reviewing the code.
> 
> Regarding comments, code style and many other things: I think Christian's comment points into the right direction: The whole human setup of the project is not yet right. Currently I am the single point of failure. Plus - there are not many "rules" how to do things. I tend to like that (not having rules), but at some points there should be some constraints and guidelines.
> 
> Regarding the actual comments I totally agree that keeping documentation in two places (method + javadoc + ...) can be problematic. But to me it is even more problematic not having a human readable documentation in addition to method names. I really like it when I can use Eclipse to hoover over a method getting a nice description showing what the method actually does (I even like it more when there is an example showing something meaningful). But I agree with @christiangoudreau that there are cases where this might not be the best way to do it...
> 
> Therefore: The next big steps for ninja will be to bring more people into the boat and replace "Raphael" with something more meaningful ;) This will happen soon - and I'd really invite you both to join the party :)
> 
> Cheers and many thanks again!
> 
> Raphael
> 
> —
> Reply to this email directly or view it on GitHub.

---

Philip Sommer | Software Developer
devbliss GmbH
Saarbrücker Straße 38 A, 10405 Berlin, Germany

Managing Director: Dr. Johann Kempe | Registered Office: Berlin, Germany | Registration Court: Amtsgericht Berlin Charlottenburg, HRB 141183 B

PR #60, opened 2012-11-29
ignore maven-enforcer-plugin in eclipse because its not supported by m2e

jbarop:
ignore maven-enforcer-plugin in eclipse because its not supported by m2e and causes noisy warnings in eclipse

btw: had problems with tabs as my editor is tab-free :)
do you relly want tabs because any other project is moving away from tabulators?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,6 +62,35 @@
 					</configuration>
 				</plugin>
 
+				<!--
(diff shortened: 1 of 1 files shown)


raphaelbauer:
perfect :) thanks!

PR #67, opened 2013-01-20
extract multiple languages from Accept-Language header

qubicllj:
Accept-Language header may contain several languages which are separated by comma

### DIFF
--- a/ninja-core/src/main/java/ninja/i18n/LangImpl.java
+++ b/ninja-core/src/main/java/ninja/i18n/LangImpl.java
@@ -216,6 +216,10 @@ public class LangImpl implements Lang {
         String[] languages = language.split(",");
         for (String l: languages){
             l = l.trim();
+            // Ignore the relative quality factor in Accept-Language header
+            if(l.contains(";")){
+                l = l.split(";")[0];
+            }
             Configuration configuration = langToKeyAndValuesMapping.get(l);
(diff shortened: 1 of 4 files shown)


raphaelbauer:
Hi qubic,

this is a really great fix :)
I have some comments:

1) Usually the header looks like: "Accept-Language: da, en-gb;q=0.8, en;q=0.7". Including some quality scores... What do you think about incorporating the q's as well? Apart from the fact the the original ninja code looks wrong - we would anyhow end up with gb;q=0.8 as country code. Or did I miss something?

2) There is a test called "LangImplTest" that would be more than happy getting some test to make sure your changes do not get annihilated by the next commit :) There is also a demo application acting as integration test that you could extend (Check out ninja-core-demo and then I18nControllerTest).

3) Please add your fix to ninja-core/src/site/markdown/developer/changelog.md (top under version 1.x). And also add yourself to ninja-core/src/site/markdown/team.md :)

Don't hesitate to contact me again - and thanks a lot :) Also let us know what you think and in what direction Ninja should evolve - we are going to organise some Ninja Hackathlons this year in Berlin - so development won't stop anytime  soon :)

Cheers,

Raphael for the Team

qubicllj:
Thanks for replay.
I think it is not necessary to consider the relative quality factor as the multiple values in Accept-Language header have been sorted.

PR #71, opened 2013-01-28
new body parser for post forms

zoza:
new body parser for post forms

### DIFF
--- a/ninja-core/src/test/java/ninja/ContextImplTest.java
+++ b/ninja-core/src/test/java/ninja/ContextImplTest.java
@@ -450,6 +450,28 @@ public class ContextImplTest {
         
         
     }
+    @Test
+    public void testParseBodyPostWorks() {
+
+        
+        when(httpServletRequest.getContentType()).thenReturn(ContentTypes.APPLICATION_POST_FORM);
+        
+        //init the context from a (mocked) servlet
+        context.init(httpServletRequest, httpServletResponse);
+        
+        when(bodyParserEngineManager.getBodyParserEngineForContentType(ContentTypes.APPLICATION_POST_FORM)).thenReturn(bodyParserEngine);
+        Dummy dummy = new Dummy();
+        dummy.name = "post";
+        dummy.count = 245L;
+        when(bodyParserEngine.invoke(context, Dummy.class)).thenReturn(dummy);
+        
+        Dummy o = context.parseBody(Dummy.class);
+        
+        verify(bodyParserEngineManager).getBodyParserEngineForContentType(ContentTypes.APPLICATION_POST_FORM);        
+        assertTrue(o instanceof Dummy);     
+        assertTrue(o.name.equals(dummy.name));     
+        assertTrue(o.count.equals(dummy.count));     
+    }
     
     /**
      * This is the default mode.
@@ -516,7 +538,8 @@ public class ContextImplTest {
     
     // Dummy class used for parseBody tests.
     class Dummy {
-        // intentionally left empty.
+       public String name;
+       public Long count; 
     }
     
     
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ bin
 .classpath
 .project
 .settings
+.metadata
 
 # when developing it is cool linking
 # the ninja framework in the main demo application
--- a/ninja-core-demo/src/main/java/conf/application.conf
+++ b/ninja-core-demo/src/main/java/conf/application.conf
@@ -23,7 +23,7 @@ application.secret=234523453tsfgdfgsdgfw345g
 
 application.session.expire_time_in_seconds=3600
 application.session.send_only_if_changed=true
-application.session.transferred_over_https_only=true
+application.session.transferred_over_https_only=false
 
 testproperty=without special mode we use this one... juchuuuu!!!!
 %test.testproperty=test testproperty!!!!
--- a/ninja-core-demo/src/main/java/views/ApplicationController/index.ftl.html
+++ b/ninja-core-demo/src/main/java/views/ApplicationController/index.ftl.html
@@ -7,7 +7,9 @@
 	<p>And developing large web applications becomes fun again.</p>
 </div>
 
-
+<#if flash_success??>
+        <p class="success">${flash_success}</p>
+</#if>
 <!-- Example row of columns -->
 <div class="row">
 	<div class="span4">
--- a/ninja-core/pom.xml
+++ b/ninja-core/pom.xml
@@ -216,6 +216,11 @@
 			<artifactId>jackson-dataformat-xml</artifactId>
 			<version>2.1.1</version>
 		</dependency>
+		<dependency>
+			<groupId>ognl</groupId>
+			<artifactId>ognl</artifactId>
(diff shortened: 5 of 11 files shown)


zoza:
i think i did last commits to wrong place.
sorry.

ninja-core/src/test/java/ninja/ContextImplTest.java
```diff
@@ -450,6 +450,28 @@ public void testParseBodyJsonWorks() {
...
+        //init the context from a (mocked) servlet
+        context.init(httpServletRequest, httpServletResponse);
+        
+        when(bodyParserEngineManager.getBodyParserEngineForContentType(ContentTypes.APPLICATION_POST_FORM)).thenReturn(bodyParserEngine);
+        Dummy dummy = new Dummy();
+        dummy.name = "post";
+        dummy.count = 245L;
+        when(bodyParserEngine.invoke(context, Dummy.class)).thenReturn(dummy);
+        
+        Dummy o = context.parseBody(Dummy.class);
```
raphaelbauer, line 468: Should read Object o = ... I guess...

ninja-core/src/test/java/ninja/ContextImplTest.java
```diff
@@ -450,6 +450,28 @@ public void testParseBodyJsonWorks() {
...
+        
+        verify(bodyParserEngineManager).getBodyParserEngineForContentType(ContentTypes.APPLICATION_POST_FORM);        
+        assertTrue(o instanceof Dummy);
```
raphaelbauer, line 471: This will be always true

ninja-core/src/test/java/ninja/ContextImplTest.java
```diff
@@ -450,6 +450,28 @@ public void testParseBodyJsonWorks() {
...
+        dummy.name = "post";
+        dummy.count = 245L;
+        when(bodyParserEngine.invoke(context, Dummy.class)).thenReturn(dummy);
+        
+        Dummy o = context.parseBody(Dummy.class);
+        
+        verify(bodyParserEngineManager).getBodyParserEngineForContentType(ContentTypes.APPLICATION_POST_FORM);        
+        assertTrue(o instanceof Dummy);     
+        assertTrue(o.name.equals(dummy.name));     
+        assertTrue(o.count.equals(dummy.count));
```
raphaelbauer, line 473: The asserts will be always true. This test is mocked => It is only tested that the body parser is called and registered....

ninja-core/src/test/java/ninja/ContextImplTest.java
zoza, line 468: Why?? you not expect to get something else except Dummy.class

ninja-core/src/test/java/ninja/ContextImplTest.java
zoza, line 471: i took this from xml\json test
but you right - this is stupid check

ninja-core/src/test/java/ninja/ContextImplTest.java
zoza, line 473: not correct (as i understand)
if something wrong inside the parser -> o.name != "post"

raphaelbauer:
Hi Anatoly,

I really like what I see so far - very well done! I am currently checking out your code and I am also trying to comment as much as I can (I hope that makes sense...) :)

Please give me some time (2 days or so) to collect feedback :)

Best,

Raphael

zoza:
take your time..   :)

thanks.

zoza:
hi.
after rethinking usage of OGNL in this case from the security perspective,
i think its maybe can be like web shell. (someone can send as key some
"System.gc()")
I'm not sure about that, but i don't want to take chances.

we need to change this to some beanUtils from apache, that work on class
property only.

PS. after some web search i found:
http://blog.o0o.nu/2010/07/cve-2010-1870-struts2xwork-remote.html
http://blog.o0o.nu/2012/01/cve-2011-3923-yet-another-struts2.html

so, definitely we need to change ognl.

raphaelbauer:
Hi Zoza,

here are my thoughts:
1. It is a really cool change :) Something I'd love to see people using!
2. As you said: We should remove ognl by a simple field based parsing (simple reflection would do imho)
3. We need a demo and test to make sure stuff does not break (demo + test inside ninja-core-demo)
4. Some documentation would be cool... otherwise people do not know it :)
5. Nice to have: Did you have a look if the validation works? Eg if we can use JSR303 Annotations on the objects parsed by your parserEngine? Should work imho....

I also thought about extending the html templating engine - right now we can only render Maps, but it is easy to add rendering of Objects. With that we'd have an object coming into a method - that can then be examined and written out by simply calling the render method... Pretty and simple workflow...

Let me know what you think about it...

Best,

Raphael

zoza:
Hi Raphael,
Sound cool.

you asking if I'm ready to do these things?  :)
04.02.2013 10:15 пользователь "reyez" notifications@github.com написал:

raphaelbauer:
On Mon, Feb 4, 2013 at 9:29 PM, Anatoly notifications@github.com wrote:

> Hi Raphael,
> Sound cool.
> 
> you asking if I'm ready to do these things? :)

No :) it is totally up to you! If you want to implement it just go ahead :)
The 5 points were just our idea to make sure the "quality" is okay and the
feature does not break occasionally when other people are adding other
features :) Adding post form parsing the way you proposed it is really
awesome!

Anyhow - we wanted to organize a non-virtual Hackathlon in Feb / March in
Berlin to add some stuff to Ninja. And this sounds like a perfect project
we could implement there within some hours. So just let me know if you want
to implement it or which parts you want to implement and what not. Hmm. You
could also join us virtually - lot's of ideas :)

Just let me know what your plans are!

Best,

Raphael

> 04.02.2013 10:15 пользователь "reyez" notifications@github.com написал:
> 
> > Hi Zoza,
> > 
> > here are my thoughts:
> > 1. It is a really cool change :) Something I'd love to see people using!
> > 2. As you said: We should remove ognl by a simple field based parsing
> > (simple reflection would do imho)
> > 3. We need a demo and test to make sure stuff does not break (demo +
> > test
> > inside ninja-core-demo)
> > 4. Some documentation would be cool... otherwise people do not know it
> > :)
> > 5. Nice to have: Did you have a look if the validation works? Eg if we
> > can
> > use JSR303 Annotations on the objects parsed by your parserEngine?
> > Should
> > work imho....
> > 
> > I also thought about extending the html templating engine - right now we
> > can only render Maps, but it is easy to add rendering of Objects. With
> > that
> > we'd have an object coming into a method - that can then be examined and
> > written out by simply calling the render method... Pretty and simple
> > workflow...
> > 
> > Let me know what you think about it...
> > 
> > Best,
> > 
> > Raphael
> > 
> > —
> > Reply to this email directly or view it on GitHub<
> > https://github.com/reyez/ninja/pull/71#issuecomment-13065382>.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/reyez/ninja/pull/71#issuecomment-13097488.

## 

inc: http://ars-machina.raphaelbauer.com
tech: http://ars-codia.raphaelbauer.com
web: http://raphaelbauer.com

PR #74, opened 2013-02-07
Feature/session cookie http only

roughy:
HttpOnly flag at the session cookie. 
On a supported browser, an HttpOnly session cookie will be used only when transmitting HTTP (or HTTPS) requests, thus restricting access from other, non-HTTP APIs (such as JavaScript). This restriction mitigates but does not eliminate the threat of session cookie theft via cross-site scripting (XSS).


raphaelbauer:
really nice...!

PR #75, opened 2013-02-10
Feature fluffy post form parsing

raphaelbauer:
@zoza could you please have a look? This is a simple addition that enables the html templating engine to directly render objects. I integrated it into your contactsForm example html... Thanks!


zoza:
you dont need to convert the object

   ObjectMapper.convertObjectToMap(object));

simply add it as value to root map

zoza:
The only one thing: 
the object must have getters\setters, but this is anyway good practice

raphaelbauer:
On Sun, Feb 10, 2013 at 11:36 PM, Anatoly notifications@github.com wrote:

> The only one thing:
> the object must have getters\setters, but this is anyway good practice
> 
> True. On the other hand (at least in my experience) it is simply a waste -
> especially for dumb Pojos... I think I have to think about that a bit
> more...

ra

zoza:
You have an IDE. One click and done
11.02.2013 9:53 пользователь "reyez" notifications@github.com написал:

PR #76, opened 2013-02-16
refactor ninja boot process

zoza:
Hi Raphael,
First of all, probably you don't need to take the "1.1" commit.
Can you explain (or give me some article) how do not send all changes or changes that already in upstream.

So... this boot refactor is necessary for me, as i added support of Apache shiro security for my app and i want to have ability to update the ninja in easy way. (not change the ninja-core only for me)
Shiro has guice module and in constructor he want servletContext
and i think that this refactor making the boot process more straightforward.

i also changed all tests stuff for new boot way.


raphaelbauer:
Hi zoza,

I'd like to know some details why you want to include Shiro. What can Shiro do what Ninja's secure annotations cannot do for you (eg http://www.ninjaframework.org/documentation/filter.html)?

What is your current use-case?

Best,

Raphael

zoza:
Hi Raphael,

I'm adding shiro to my app, not to core. but for doing this in the right
way i need to change the core.
I assume that exists more 3-party libs (as guice modules) that want
servletContent at injector creation time.
Also I assume that exists some 3-party libs that wants native servlet
filters.
And as i said before: start ninja from servlet filter its like enter to
home through the window.

You can't equal Shiro security and ninja sec filter.
ninja sec filter CAN be same like the shiro only after many months of dev
 :)
http://shiro.apache.org/features.html

If you have more questions - ask me..

raphaelbauer:
Yes. What is your exact use case? I just played around with Shiro and think it is easily possible to use Shiro together with Ninja's filters...

zoza:
Please look at shiro-guice-aop and shiro-guice-web
http://shiro.apache.org/guice.html
They wrote all code already

i simple define
addFilterChain("/user/profile/**", AUTHC);
or add annotation on method
@RequiresRoles(value = "admin")

of course i can implement this via ninja filters, but it's done already.

but shiro-guice-web use guice-servlets for defining servlet filter

raphaelbauer:
Just digged in the source of Shiro and the Guice implementation. I guess Shiro is using server side sessions. Right?  So you are loosing the statelessness of a Ninja application I guess...

zoza:
in my personal case it's not an issue.

but in common case you can easily override sessionManager or sessionDAO to
use cookie as now

raphaelbauer:
okay... I'll meet some stakeholders of Ninja on Thursday :) Will keep you posted!

raphaelbauer:
hi zoza,

i asked on the mailing list. at the beginning we talked a lot of being independent of the servlet api. now it feels a bit wrong to integrate servlets into ninja. but maybe i am wrong... if nobody objects we can integrate it...

i also played around with your code. pretty well done :) 
i made some tiny adjustments and wanted your feedback:
https://github.com/reyez/ninja/compare/zoza-servlet-integration-sketch
- i added an optional conf.ServletModule (will be loaded when present in a ninja project) => if it is not there only the default filter of ninja will be loaded => a bit more compliant to the way it is done until now (conf.Module + conf.ServletModule)
- i removed the dependency on the servletcontext. do we need that really?

apart from that i really like the way you are booting ninja now. it is a lot better to use guice to load everything as far as i can see...

if you add me as collaborator i can also contribute to your pull request...

best,

raphael

zoza:
1. i saw binding bind(NinjaServletDispatcher.class).asEagerSingleton()
   twice:

in configureServlets() and also in Configuration.java
i think its better to put all servlet filters binding in one place in
configureServlets()

> i removed the dependency on the servletcontext. do we need that really?
> :) this is mandatory for shiro. i started all these changes because of it.

raphaelbauer:
okay... binding NinjaServletDispatcher is clearly an error - thanks for pointing that out... Regarding ServletContext => I would add that to conf.ServletModule if that is okay for you...

zoza:
ok, i will install my shiro modules from ServletModule

raphaelbauer:
I just saw that ServletModule already has a getServletContext() - this is the one you need right? So - no need to change anything...

You can eg inject the servletcontext... in an applicationcontroller...
    @Inject 
    ServletContext servletContext ;

zoza:
yes, its ok,
after your merge i will check if everything work fine with my code

zoza:
so.. you planning to merge my pull request??

raphaelbauer:
On Sun, Mar 3, 2013 at 8:33 PM, Anatoly notifications@github.com wrote:

> so.. you planning to merge my pull request??

I planned to finish my code + make a new test project + write some
documentation this week. is that okay?

Cheers,

Raphael

zoza:
Fine.. :)
04.03.2013 12:43 пользователь "reyez" notifications@github.com написал:

PR #79, opened 2013-03-03
Feature/error objects in validation

philipsommer:
- Fixes in Validation
  - Validation for controller parameter fields as well as validation for injected beans


raphaelbauer:
Hi Phil,

just played around with it and it works like a charm... 

It would be really cool to have the ninja-core-demo extended with some validation examples (for json and for html forms). This is really straight forward => maybe I can do that next week...

In any case we'll meet next week to discuss that :)

Really good work!

Raphael

zoza:
Hi Raphael, what about to add feature "validate on demand" ?? as we talked?

PR #82, opened 2013-03-13
fixed egg-chicken problem

zoza:
before this fix, we boot ninja (and as result create injector for
servletModule) before servletContext populated in
super.contextInitialized(servletContextEvent);


raphaelbauer:
Thanks Anatoly,

I think I saw several times minor problems in the dev mode due to this...
Cheers,

Raphael

PR #84, opened 2013-03-29
Set content type when it is not explicitly set

qubicllj:
I use rrd4j to generate picture and convert it to bytes array, "application/octet-stream" will make browser to download the picture which is not what i want.

### DIFF
--- a/ninja-core/src/main/java/ninja/utils/ResultHandler.java
+++ b/ninja-core/src/main/java/ninja/utils/ResultHandler.java
@@ -121,7 +121,11 @@ public class ResultHandler {
             } else if (result.getRenderable() instanceof byte[]) {
                 // Simply write it out
                 try {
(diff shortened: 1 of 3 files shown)


raphaelbauer:
Hey qubic,

sorry for my last post - my bad - I looked at the wrong file revision. Your change makes 100% sense. Thanks for that!

Two more things:
Could you please 
- add a tiny testcase to ResultHandlerTest.java to make sure byte[] does not get a forced octet stream
- add the bug you fixed to changelog.md (ninja-core/site/markdown/developer/)?

This will really help us to make sure the bug does not reappear :)

If you have problems extending the testcase let me know. I think I can do that too asap...

Thanks again for your contribution!

Best,

Raphael

raphaelbauer:
absolutely perfect! thanks :+1:

PR #85, opened 2013-04-07
Remove dependency of the core on javax.servlet

tbroyer:
This is a breaking change but rather with limited impact:
- `Context.getHttpServletRequest` / `Response` methods were `@Deprecated` already
- `FlashCookie` / `SessionCookie`'s `save` method now take `Result` as a second parameter, but users are not supposed to call those methods (or are they?)
- several classes relocated (`ninja.async` → `ninja.servlet.async`, `ninja.utils.CookieHelper` → `cookie.servlet.CookieHelper`, `ninja.utils.ResponseStreamsServlet` → `ninja.servlet.ResponseStreamsServlet`)
- **biggest breaking change:** introduce a new `ninja-servlet` module where all servlet-related code lives (so that `ninja-core` doesn't depend on `guice-servlet` or `javax.servlet`)

This paves the way for supporting other containers, such as Netty or Grizzly.

TODO: move the test lib to an in-process model, possibly splitting the jetty-based implementation into its own module (or removing it entirely); similarly to how Jersey 2.0 provides several backends for tests.


raphaelbauer:
Hi Thomas,

looks really good.

Just one thing: Why did you add Result to the Session and Flash Scopes? When you finalize the headers the request stuff gets copied over automatically... https://github.com/reyez/ninja/blob/develop/ninja-core/src/main/java/ninja/servlet/ContextImpl.java and finalizeHeaders . Maybe I am missing something here...

Thanks,

raphael

tbroyer:
Have a look at the changes in `FlashCookieImpl` and `SessionCookieImpl`: they used `getHttpRequestResponse` to `addCookie`. If you remove that method, then they have no way to add a cookie to the response other than `Result#addCookie`, which is the proper of doing it: read from `Context`, write to `Result`.

raphaelbauer:
Ah yes. I missed that one. Very cool Thomas...

PR #91, opened 2013-05-07
ninja-core-demo-archetype should use fixed ninja version in pom.xml

seratch:
ninja-core-demo-archetype is an awesome tutorial! But it doesn't work because pom.xml depends on ninja 1.4-SNAPSHOT.

Please merge this PR and re-deploy the archetype.

Thanks,
-seratch

### DIFF
--- a/ninja-core-demo-archetype/src/main/resources/archetype-resources/pom.xml
+++ b/ninja-core-demo-archetype/src/main/resources/archetype-resources/pom.xml
@@ -26,7 +26,7 @@
 	<parent>
 		<groupId>org.ninjaframework</groupId>
 		<artifactId>ninja</artifactId>
-		<version>1.4-SNAPSHOT</version>
+		<version>1.4</version>
 	</parent>
 
 	<build>
@@ -64,12 +64,12 @@
 		<dependency>
 			<groupId>org.ninjaframework</groupId>
 			<artifactId>ninja-core</artifactId>
-			<version>1.4-SNAPSHOT</version>
+			<version>1.4</version>
 		</dependency>
 		<dependency>
 			<groupId>org.ninjaframework</groupId>
 			<artifactId>ninja-servlet</artifactId>
-			<version>1.4-SNAPSHOT</version>
+			<version>1.4</version>
 		</dependency>
 
 		<dependency>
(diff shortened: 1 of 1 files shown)


raphaelbauer:
Thanks for the fix! I'll push 1.4.1 asap

seratch:
Though 1.4.1 is now available on Maven central, 1.4 is still used by default. It works if users specify archetypeVersion,, but it's not a good situation.

```
mvn archetype:generate -DarchetypeGroupId=org.ninjaframework -DarchetypeArtifactId=ninja-core-demo-archetype -DarchetypeVersion=1.4.1
```

I'm afraid that archetype-catalog.xml seems not to be updated. Do you know how to fix it?

https://oss.sonatype.org/content/repositories/releases/archetype-catalog.xml

raphaelbauer:
On Tue, May 7, 2013 at 10:23 PM, Kazuhiro Sera notifications@github.comwrote:

> Though 1.4.1 is now available on Maven central, 1.4 is still used by
> default. With specifying archetypeVersion, it works, but it's not a good
> situation.
> 
> mvn archetype:generate -DarchetypeGroupId=org.ninjaframework -DarchetypeArtifactId=ninja-core-demo-archetype -DarchetypeVersion=1.4.1
> 
> I'm afraid that archetype-catalog.xml seems not to be updated. Do you know
> how to fix it?
> 
> https://oss.sonatype.org/content/repositories/releases/archetype-catalog.xml

Okay the sonaytype archetype-catalog I get contains 1.4.1. But
http://repo1.maven.org/maven2/archetype-catalog.xml does not contain it.

I guess the sonatype archetypes will be synced once in a while over to the
repo1 catalog... I'd bet 1.4.1 will be there soon (guess within 24hrs)...
Then you no longer have to specify 1.4.1 :)

Does that make sense?

Best,

Raphael

seratch:
I see. Thank you for your kind reply!

Thanks,
-seratch

raphaelbauer:
On Tue, May 7, 2013 at 11:58 PM, Kazuhiro Sera notifications@github.comwrote:

> I see. Thank you for your kind reply!

Thanks for catching the bug and submitting a fix :)

Best,

Raphael

raphaelbauer:
Quick update: https://issues.sonatype.org/browse/MVNCENTRAL-192
says:
The archetype-catalog.xml file on Central is actually updated once a week by a scripted job, so you shouldn't need to do anything to get it picked up eventually.

PR #93, opened 2013-05-24
access the session-cookie easily in templates

pthum:
Until now, you had to pass e.g. the username manually in every controller-method to render it in a template. Now, the values of the session-cookie will be added to every template and can be accessed by calling the key as variable in the template (this can be useful for an individualized header for logged-in users).

I don't know if this is a good solution. Maybe its better to add a variable to the application.conf and check whether this set to true? Another possibility may be to prefix all keys (like shown here: http://www.java2s.com/Tutorial/Java/0140__Collections/Clonesamapandprefixesthekeysintheclone.htm ). But this could lower the performance...

### DIFF
--- a/ninja-core-demo/src/main/java/controllers/ApplicationController.java
+++ b/ninja-core-demo/src/main/java/controllers/ApplicationController.java
@@ -103,8 +103,9 @@ public class ApplicationController {
 
     public Result session(Context context) {
         context.getSessionCookie().put("username", "kevin");
+        //context.getSessionCookie().put("", "");
 
-        return Results.html().render(context.getSessionCookie().getData());
+        return Results.html();//.render(context.getSessionCookie().getData());
 
     }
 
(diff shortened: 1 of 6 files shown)


raphaelbauer:
Hi pthum,

this is really cool :)
My personal opinion is to use "session" as implicit value for the template namespace (afaic ruby and play do it the same way). If you got a username in your session you can use then ${session.username} in your template...

This is similar to doing
map.put("session", context.getSessionCookie().getData());

(Almost the same what you did - you just store it inside another accessor).

Would be cool if you could change that (and update the docs, and clean the code a bit). And don't forget to add yourself to team.md (the contributors list :) )...

Once that's done we'll add your feature asap.

Thanks again for your contribution! really cool :)

Cheers,

Raphael

pthum:
Oh, of course! You're right, it's better to use "map.put("session", context.getSessionCookie().getData());" .
I did what you said and would implement the contextpath-support, as mentioned in the mailinglist ( https://groups.google.com/forum/#!topic/ninja-framework/Q812aKhfiEk ), too. but i don't know the current state of seratch's work...?
I also changed the documentation for internationalization (the use of "Messages" instead of "Lang") and fixed two typos in the i18nController of ninja-core-demo, but i don't know whether to commit this here or in another pull request...

PR #94, opened 2013-06-05
Feature contextpath support

pthum:
i added the contextpath-support in templates. i just extended the context-interface, added it in the contextimpl, wrappedcontextimpl and fakecontext. at last i put it in the map of the template-engine-freemarker and added it as implicit variable in the documentation.


raphaelbauer:
Thanks! Looks good :+1:

PR #96, opened 2013-06-06
Bigfix messages parsing

raphaelbauer:
- fixes an issue with flash cookies and their translation
- adds the much better option to translate messages inside the template via ${i18n(messageKey, parameters...}


pthum:
yeah, great! this looks very good (even when its very disturbing to change all the i18n-keys in an app :) ).

raphaelbauer:
Thanks! You can still use the old way. We should turn that off at some point in time... :)

PR #100, opened 2013-06-14
Dev mode only ftl auto reload

makotan:


raphaelbauer:
Hey thanks :)

what does this actually fix? Or what is the problem you are encountering?

Raphael

makotan:
only when the dev mode, reflected in the modification of reload ftl.html

raphaelbauer:
okay I see...
The thing is => I usually use Ninja together with jetty that reloads when something new is in the target directory. Eclipse on the other hand copies new stuff from src over to target when I change something. That means I modify a ftl.html file, eclipse copies it over to target, jetty restarts and I get my new ft.html.

Doesn't that work for you? 

Thanks :)

makotan:
restart of the jetty will be unnecessary.Is it still required?

raphaelbauer:
Hmm.
I agree that a restart of jetty is totally unnecessary.
On the other hand.
I start jetty usually on the command line mvn jetty:run
=> it then picks up changes in its classes directory.

Eclipse by default copies the classes anyhow into the classes directory and triggers a Jetty reload.

So in my case - even if we point the ftl.html to the sources directory => it will trigger a reload in my case anyways...

My question:
How do you start Ninja? Via maven?

makotan:
is via maven
I makes frequent modification of ftl. And I would like to see the results quickly.

raphaelbauer:
How did you configure jetty not to restart when a ftl.html changes (I assume your IDE copies over the file to the target dir when changed)..?

makotan:
Only when dev mode,I was changed from setClassForTemplateLoading to setDirectoryForTemplateLoading.
To get the path from System.getProperty("user.dir"), and added "/src/main/java"

raphaelbauer:
thanks! very good feature :)

PR #101, opened 2013-06-27
skip the ftl render when redirect(3xx)

makotan:
skip the ftl render when redirect(3xx)

### DIFF
--- a/ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
+++ b/ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
@@ -238,10 +238,14 @@ public class TemplateEngineFreemarker implements TemplateEngine {
         // Here I set a file directory for it:
         try {
 
-            Template freemarkerTemplate = cfg.getTemplate(templateName);
+            if ((result.getStatusCode() % 100) != 3) {
+                // not use redirects
 
-            // convert tuples:
-            freemarkerTemplate.process(map, responseStreams.getWriter());
+                Template freemarkerTemplate = cfg.getTemplate(templateName);
+
+                // convert tuples:
+                freemarkerTemplate.process(map, responseStreams.getWriter());
+            }
 
             responseStreams.getWriter().flush();
             responseStreams.getWriter().close();


raphaelbauer:
hi makotan :)

thanks for the pull request...
I think it is in general a very good idea to provide the default possibilty of redirecting without the need for a body (and hence a ftl.html file).

But http://www.ietf.org/rfc/rfc2616.txt clearly says that redirects should allow for html to be included (in case clients don't understand it).

My suggestions would be the following:

1) A new class NoHttpBody.
2) An extension to https://github.com/ninjaframework/ninja/blob/develop/ninja-core/src/main/java/ninja/utils/ResultHandler.java

that checks if object.getRenderable() instanceof NoHttpBody and if so skips the render step (only a call to context.finalizeHeaders(result)).

3) Change in Results java:
    public static Result redirect(String url) {

```
    Result result = status(Result.SC_303_SEE_OTHER);
    result.addHeader(Result.LOCATION, url);
    result.render(NoHttpBody());

    return result;
}
```

This is I think the same  as your pull request, but allows 
1) Users to customize the behavior
2) Works for all content types and not only html.

Downsides:
It may break rendering for "old" users of Results.redirect(...)..

Does that make sense or is this too complicated? Don't hesitate to critizise the approach :)

Thanks!

Raphael

makotan:
I think your suggestions is good.
This feature is even want an implementation of either :)
Would you like to fix compatibility is lost? I think Well, the fix we should be cautious.

pthum:
i think that this is a cool suggestion (with the NoHttpBody and so on). i got the problem too and thought about to open an issue  or find a solution by myself to create a pull request.
at the moment i create a result at the beginning, set with a template that does not expect objects (in my case the index-page, but maybe it should be a better way to create an own empty template for this case..) and then i either do the redirect or set another template and render stuff (dependend on the logic).

As i see it, the RFC just says that there should be a page with a link to the redirection-target. So this could be an automatically generated body of the "NoHttpBody" :)

makotan:
The only good if you set "views/system/redirect.ftl.html" to template when Results.redirect(...) is called?

raphaelbauer:
Hi @pthum and @makotan,

I think your ideas of auto-generating the http body with a link are even better. Using a "views/system/redirect.ftl.html" would even allow users to overwrite the default styling of the pages - just in case.

Do you have already any code to review? Otherwise I might have time next week to implement the feature.

Cheers,

Raphael

makotan:
Hi Raphael,

> Do you have already any code to review?

No.

If you make, do I close this?

raphaelbauer:
Hi,

I found some time to add the NoHttpBody stuff. 
Please check out https://github.com/ninjaframework/ninja/pull/105 
Comments are really welcome :)

This mimicks the way Rails and Play do handle redirects. They do not show a body by default (what is not 100% according to the http spec... but ok).

Cheers,

Raphael

PR #103, opened 2013-06-29
UserAgent ftl switch support

makotan:
UserAgent ftl switch support

### DIFF
--- a/ninja-core/src/main/java/ninja/Result.java
+++ b/ninja-core/src/main/java/ninja/Result.java
@@ -97,6 +97,8 @@ public class Result implements Serializable {
 
     private String template;
 
+    private String uaType = "";
+
     /**
      * A result. Sets utf-8 as charset and status code by default. 
      * Refer to {@link Result#SC_200_OK}, {@link Result#SC_204_NO_CONTENT} and so on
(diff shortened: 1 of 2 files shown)


pthum:
i suppose, that it would be nice if you would add a more detailed description to your pull requests. especially you should write what it does and why you're doing this/what you've intended to do. 
i can't imagine why it should be useful to have a structure like "Mozilla/5.0 (Windows NT 6.1; WOW64; rv:21.0) Gecko/20100101 Firefox/21.0/views/Application/index.ftl.html" ..?

makotan:
It is the image that sets the UserAgent in Filter.
For example, you can use it to set the Result#uaType by converting "Mozilla/5.0 (Windows NT 6.1; WOW64; rv:21.0) Gecko/20100101 Firefox/21.0" to "firefox".
To be used on the final is "firefox/views/Application/index.ftl.html".

raphaelbauer:
Hi Makotan,

can you describe a bit more what you are trying to achieve? What is your use case? 

A workaround for you certainly is to use Results.html().template(pathToYourUaBasedTemplate);
You can even extend Results.java to make that happen automatically for you I guess. Ninja is pretty flexible in that regard.

Best,

Raphael

makotan:
In the beginning, I thought it was going to use a Results.html().template(...) to switch the template.
But, I noticed that you can use the same file UserAgent differ only in.
I thought there we should be able to add the UserAgent.
use case is a Web application of a special terminal for the browser installed.

raphaelbauer:
wont-fix

PR #107, opened 2013-08-24
Exclude type:String from being converted to JSON string again

sojin:
objectMapper.writeValue(OutputStream, String) will adorn the String with extra quotes. Which is in fact an invalid JSON according to the RFC 4627. I believe we should bypass if the type is String. 
Moreover this will give users flexibility to use any other JSON serializer in application if they need.

### DIFF
--- a/ninja-core-demo/src/test/java/controllers/ApplicationControllerTest.java
+++ b/ninja-core-demo/src/test/java/controllers/ApplicationControllerTest.java
@@ -119,7 +119,7 @@ public class ApplicationControllerTest extends NinjaTest {
                         + "validation?email=john@example.com");
 
         // And assert that stuff is visible on page:
-        assertEquals(response, "\"john@example.com\"");
+        assertEquals("john@example.com", response);
 
         response = ninjaTestBrowser.makeRequest(getServerAddress() + "validation");
 
--- a/ninja-core/src/main/java/ninja/template/TemplateEngineJson.java
+++ b/ninja-core/src/main/java/ninja/template/TemplateEngineJson.java
@@ -33,7 +33,7 @@ import com.google.inject.Singleton;
 public class TemplateEngineJson implements TemplateEngine {
 
     private final Logger logger;
-    
+
     private final ObjectMapper objectMapper;
 
     @Inject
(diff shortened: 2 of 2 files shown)


raphaelbauer:
I think Jackson produces perfectly fine Json. And something like "string" is valid Json as well.

Also tried that in Javascript, and it gets evaluated properly:
            var json = '"result"',
            obj = JSON.parse(json);

```
        alert(obj);
```

As far as I can tell http://www.json.org/ and validators like http://www.freeformatter.com/json-validator.html also tell me that "me" is okay..

What do you think?

sojin:
Yes I agree, "me" is a valid JSON. But if the String is already a JSON string, it must have quoted already. For eg:

```
public static void main(String[] args) throws JsonProcessingException {
    ObjectMapper m = new ObjectMapper();
    String value = "me";
    String jsonStringBySomeOtherJsonLib = m.writeValueAsString(value); // Could be from GSON, setting this JSONstring in render map in Controller.
    String ninjaRenderedJsonString  = m.writeValueAsString(jsonStringBySomeOtherJsonLib); // Ninja trying to convert to JSON String again.
    System.out.println("jsonStringBySomeOtherJsonLib - " + jsonStringBySomeOtherJsonLib + ", " + "ninjaRenderedJsonString - " + ninjaRenderedJsonString);
}
```

output:
jsonStringBySomeOtherJsonLib - "me", ninjaRenderedJsonString - "\"me\""

For eg:, Play framework provide RenderJSON(String jsonSting) 
https://github.com/playframework/play1/blob/master/framework/src/play/mvc/results/RenderJson.java
to give user flexibility to render jsonString without adding extra quotes.

raphaelbauer:
Hi Sojin,

I see the "clean" way how to do this as depicted in: https://gist.github.com/reyez/6329621
A "renderable" always bypasses any default rendering engine. We could potentially just add that code to Results.java.

You can then use this simply by returning ResultJsonCustom.render(yourString) in your controller code.

Btw. The code in Play1 has some performance issues - it is not really a good practise generating a "new" Gson() all the time... Maybe we should improve that :)

Let me know if that helps :)

Best,

Raphael

sojin:
Yup, that looks good. Thanks!

raphaelbauer:
Hi Sojin,

just had some minutes... Can you check if https://github.com/ninjaframework/ninja/pull/108 fits your needs? It is now implemented in a backwards compatible way. This also allows us to use e.g. Results.xml().renderRaw(customXml) or Results.html().renderRaw(customHtml) and so on... It is not limited to json...

Best,

Raphael

sojin:
Hi Raphael, 

The change looks perfect. 

Just a quick question on a separate note,

How are we going to mention the ninja.mode (to prod) or any other env var for that matter, when deploying to Tomcat? As you are probably aware, adding env variable in Tomcat is not straight forward and bit tedious (editing files in Tomcat conf dir etc.). I believe the rationale behind to recognize the mode externally is probably to make the package env agnostic?

But have you given any thought of making use of application.conf to add something like "ninja.mode=prod/dev/test"? while env variable (if given) taking priority?

Rgds
Sojin

raphaelbauer:
Hi Sojin,

actually our environment usually is plain Jetty, Heroku or AppEngine or self hosted. We did not really have the need to use anything different than the stupid system variable "ninja.mode".
But you are right - in Tomcat that is not really cool.

Feel free to implement something that fits your needs. JNDI is certainly something that should be used. Or a context variable in web.xml...

Have a look at:
https://github.com/ninjaframework/ninja/blob/develop/ninja-core/src/main/java/ninja/utils/NinjaPropertiesImpl.java
and
https://github.com/ninjaframework/ninja/blob/develop/ninja-servlet/src/main/java/ninja/servlet/NinjaBootstap.java

This should get you started :)

Best,

Raphael

raphaelbauer:
jointly declined.

PR #116, opened 2013-10-29
Adding consistent method for beans validation.

nobullet:
Added consistent hasBeanViolation(String) method.
I believe this API can be revised also.

### DIFF
--- a/ninja-core/src/main/java/ninja/validation/Validation.java
+++ b/ninja-core/src/main/java/ninja/validation/Validation.java
@@ -114,6 +114,14 @@ public interface Validation {
     void addBeanViolation(FieldViolation fieldViolation);
 
     /**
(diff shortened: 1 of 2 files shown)


raphaelbauer:
Hi nobullet,

first of all thanks for he pr. You are absolutely right. The validation Api can and should be revised. If you have ideas let me know. Or let's discuss this via the mailing list or skype / hangout.

I merged your pr, but I'll add a testcase (there was just a test in the integration-test, but not for the class you modified) before releasing it.

Thanks :)

Raphael

raphaelbauer:
Hey @nobullet 

turns out there is an array of tests at ControllerMethodInvokerTest.
There are a lot of gems, but the validation should be documented properly and improved...

Best,

Raphael

PR #119, opened 2013-10-31
Added bean validations tests to ControllerMethodInvokerTest (from my previous pull request).

nobullet:
> > turns out there is an array of tests at ControllerMethodInvokerTest.
> > There are a lot of gems, but the validation should be documented properly and improved...

Added bean validations tests to ControllerMethodInvokerTest.

Regarding improvements.
I think ninja.validation.Validation is documented very well. But site documentation doesn't mention how to use fields violation (it mentions only bean violations). May be this should be added to site documentation?


raphaelbauer:
Perfect. Very well done!
Maybe you are right. The documentation is really thin. There is currently also a question on the mailing list. (https://groups.google.com/forum/#!topic/ninja-framework/7RskQR9NVRs) Afics the thread is exactly about the feature you just implemented!

PR #121, opened 2013-11-05
Ninja DAO abstract class helper

emiguelt:
Dao Test Base class to help testing real DAOs using JPA

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ bin
 .metadata
 *.iml
 .idea
+*~
 
 # when developing it is cool linking
 # the ninja framework in the main demo application
--- a/ninja-core/src/site/markdown/developer/changelog.md
+++ b/ninja-core/src/site/markdown/developer/changelog.md
@@ -1,3 +1,6 @@
+Version 2.2.0
+* 2013-11-05 New NinjaDaoTestBase class to facilitate test your DAOs with a real database (emiguelt)
+
 Version 2.1.0
 =============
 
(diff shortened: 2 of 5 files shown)


raphaelbauer:
Pretty cool @emiguelt!

Some minor things:
- I guess it would make sense to add a@After method that calls  persistService.stop(); or did I miss something?
- It would be great if you could add some documentation to ninja-core/src/site/markdown/testing.html. Your commit is pretty cool and the others should know about it :)

And last but not least please add a tiny snippet to changelog.md and also team.md (as described here: http://www.ninjaframework.org/developer/getting_started.html)

I'll merge and release that asap...

Cheers,

Raphael

emiguelt:
Hi @raphaelbauer!
Thanks a lot for checking my PR.

About the _persistenceService.stop()_, before I had a _jpaInitializer.stop()_ in a @after method, but it generates a _java.lang.Exception: DEBUG -- CLOSE BY CLIENT STACK TRACE_. I read about this error and it is not a real exception (http://stackoverflow.com/questions/8403227/weird-error-close-by-client-stack-trace) . In anyway I decided to remove the _stop()_ call and avoid this error message. I checked it and all is working ok.

About other suggestions, I added some documentations about the NinjaDaoTestBase class.

Best,

Miguel

raphaelbauer:
Hi @emiguelt,

I think shutting down the JPA connection is important. If we don't do this this might result in blocked connections - especially when connecting to real db systems. But the solution to the stupid error message is simple. And actually this is something we should have done way before. All logback.xml files of Ninja are currently set to <root level="debug">. We should change that to <root level="info"> in all logback files...

This helps a lot to separate noise from real error messages...

What do you think?

Raphael

raphaelbauer:
I just changed the loglevel of the archetypes and the integration tests to "info". Please let me know if that works for you...

emiguelt:
Hi Raphael,
Done, I annotated two methods to initialize and stop the persistence service with the JUnit lifecycle.
Best,
Miguel :)

raphaelbauer:
nice!

PR #137, opened 2014-02-03
Primitive type parsing for Post form params

metacity:
Simply put: Post form body parser enhanced to parse and set primitive type (or corresponding object wrapper) fields of objects. 

(First commit to an open source project ever! And thank you for Ninja :+1: )


raphaelbauer:
Hi @metacity ,

thanks for the nice feature and congrats to your first contribution :+1: 

we need some more stuff to make this happen:
1) A testcase that makes sure your feature works and will work in future versions of Ninja. Either here: ApplicationControllerTest - testPostFormParsingWorks method. (in ninja-servlet-integration-test). Or possibly better as mocked test in ninja-core (there is no mocked test for the BodyParserEnginePost yet).
2) Documentation (maybe a tiny paragraph or listing what can be parsed: form_parsing.md)
3) Then add yourself to team.md as contributor
4) Sign your contributor agreement from: https://docs.google.com/forms/d/1Yasrxa17kYfaNDgbRvn77rM3WTnU_Um0rwz3GfzPp9g/viewform

Looks like a lot of stuff that's still to do - but actually you are already there. We just have to make sure that Ninja stays stable in the future (testcases) and people know about your feature  (docs) :)

If you got problems please don't hesitate to ask. For instance I can help writing the mocked test or anything else :) Jut let me know!

Thanks again!

Cheers,

Raphael

metacity:
Hello,

Thanks for you input! I addressed your points (maybe I'll add myself to team.md when I have done a little more ;) in the new commits. I didn't want to pollute the existing contact form classes and files, so I used the existing FormObject model (in ninja-servlet-integration-test) as it wasn't being used anywhere. 

One question/note: I chose to use i.e. `Integer.valueOf(someString)` even when populating a primitive type field. This of course creates unnecessarily an object, but imo is more readable, because we don't have to use i.e. `field.setInt(t, someInt)` specifically for primitive int type. And processing post forms most likely isn't the bottle neck anyway. What's your take?

ninja-servlet-integration-test/src/test/java/controllers/ApplicationControllerTest.java
```diff
@@ -135,22 +135,68 @@ public void testPostFormParsingWorks() {
...
+        
+        formParameters.put("primShort", "32456");
+        formParameters.put("objShort", "32455");
+        
+        formParameters.put("primChar", "Z");
+        formParameters.put("objChar", "X");
 
         String response =
                 ninjaTestBrowser.makePostRequestWithFormParameters(
-                        getServerAddress() + "/contactForm",
```
raphaelbauer, line 144: Can you remove the old /form and /contactForm stuff? (in routes.java, ApplicationController and the views?).

The reason is that it does not make sense having /contact and /contactForm when there are no tests. Your /form covers all cases.

ninja-servlet-integration-test/src/test/java/controllers/ApplicationControllerTest.java
```diff
@@ -135,22 +135,68 @@ public void testPostFormParsingWorks() {
...
+                        getServerAddress() + "/form",
                         headers,
                         formParameters);
```
raphaelbauer, line 179: You can also make it more clear that you are dealing with Json by using that one:
        ObjectMapper objectMapper = new ObjectMapper();
        FormObject formObject = objectMapper.readValue(response, FormObject.class);

Then you can make all asserts on the formObject...

(ninjaTestBrowser is not really nice to use in that scenario though...).

raphaelbauer:
Your pull request looks really nice. I Added some tiny comments. Would be great if you could fix the stuff...

Don't forget to add your change to changelog.md / your name to team.md and sign the contributor's agreement (really simple and online).

Once that's done I'll merge your stuff asap :)

Thanks!

metacity:
Will do (contributor's agreement already signed earlier)! 

The reason I preserved contactForm stuff was because the documentation about post form parsing was clearly based on them. But true; no tests, not really useful to keep them. 

About ninjaTestBrowser: what's the reason for it not being really nice in ObjectMapper's case? The original test (and many other tests in ApplicationControllerTest) are based on testing whether strings are contained in the response HTML (Freemarker template), but using a JSON string seemed clearer and less polluted by template HTML. Automatic object rendering with Freemarker (`Results.html().render(parsedObject)`) will then need its own test, but it's not directly associated with post form parsing anyway.

raphaelbauer:
Perfect!

Regarding json => There are historical reasons why the other tests use the silly "check if wanted string is in response" approach. I mean - it works - and it is halfway okay if you use it with html output (beacause html is hard to parse). But it is a bit clumsy...

Your approach using a json is a lot better and much more precise :)

metacity:
Should be in shape now! I also included a simple test for direct object rendering with Freemarker (which is what the original Post form parsing test did as a side-product). I also noticed that a renderable object is referenced with the object's **class** name (returned by `getRealClassNameLowerCamelCase(Object)` in SwissKnife class) instead of object's name . Fixed the section in form_parsing.md about that.

raphaelbauer:
Nice work! thanks for contributing :)

PR #143, opened 2014-02-13
Added convert string parameter to other class type

amsz:


raphaelbauer:
Hi amsz,

I added some tiny tings. Eg the converter method now returns null when something goes wrong and no longer throws a Runtime exception (this will stop the server)...

Very good job!

Cheers,

ra

amsz:
It is a beautiful change, but it may be introduce a bug in BodyParserEnginePost.invoke(), I just catch a UnsupportedOperationException if convert fail, now should judge if it is null.

raphaelbauer:
Hey amsz,

very good catch! you are absolutely right. On the other hand the code was a bit buggy anyway - because setting a not supported Field type (like let's say StringBuffer) to the incoming String does not make sense either.

I added a testcase and rewrote the stuff. If something is not supported and cannot be converted it will become null.

Let me know what you think (it's already in develop) :)

Cheers,

Raphael

amsz:
I have no good idea, may be expose a method that user can register a custom type convertor, like BodyParserEngine::register(String contextType, Class<T> classOfT, BodyConvert convert)

PR #144, opened 2014-02-15
Feature to specify custom location for static assets

sojin:
Hi Raphael, 

Sometimes its useful to let application serve static assets from outside of project directory. Could be as a first to step to let front end server like nginx/cherokee to serve them.

Added property in application.static.asset.basedir in app conf file, which will trigger ninja to serve them from that location.

Added a documentation around this as well in static_assets.

Please let me know your thoughts.
Rgds
Sojin

### DIFF
--- a/ninja-core/src/main/java/ninja/AssetsController.java
+++ b/ninja-core/src/main/java/ninja/AssetsController.java
@@ -16,32 +16,20 @@
 
 package ninja;
 
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.net.URLConnection;
-
 import com.google.common.base.Optional;
-import ninja.utils.HttpCacheToolkit;
-import ninja.utils.MimeTypes;
-import ninja.utils.NinjaProperties;
-import ninja.utils.ResponseStreams;
-
-import org.apache.commons.io.IOUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import com.google.common.io.ByteStreams;
 import com.google.common.io.Files;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
+import ninja.utils.*;
+import org.apache.commons.io.FilenameUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import javax.swing.text.html.Option;
-
+import java.io.*;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLConnection;
 /**
  * This controller serves public resources under /public
  * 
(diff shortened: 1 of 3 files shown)


ninja-core/src/main/java/ninja/AssetsController.java
```diff
@@ -452,4 +461,20 @@ public static String getFileNameFromPathOrReturnRequestPath(Context context) {
...
     }
 
+    private Optional<String> getNormalizedAssetPath(NinjaProperties ninjaProperties){
+
+        Optional<String> assetBaseDir = Optional.absent();
+        String baseDir = ninjaProperties.get( "application.static.asset.basedir");
+
+        if(baseDir != null){
+
+            String pathSeparator = baseDir.substring( baseDir.length() - 1 );
```
raphaelbauer: Maybe you can use Guava's Files.simplifyPath. Saves some lines...

ninja-core/src/main/java/ninja/AssetsController.java
```diff
@@ -287,68 +294,71 @@ private URL getFileFromAssetsDir(Context context) {
...
+
+    private Optional<URL> getUrlForFile( File possibleFileInSrc) {
+        try {
+            return  Optional.fromNullable( possibleFileInSrc.toURI().toURL() );
+
+        } catch(MalformedURLException malformedURLException) {
+
+            logger.error("Error in dev mode while streaming files from src dir. ", malformedURLException);
+        }
+        return null;
```
raphaelbauer: return null is likely not what we want... Optional.absent() possibly better?

ninja-core/src/main/java/ninja/AssetsController.java
```diff
@@ -452,4 +461,20 @@ public static String getFileNameFromPathOrReturnRequestPath(Context context) {
 
     }
 
+    private Optional<String> getNormalizedAssetPath(NinjaProperties ninjaProperties){
+
+        Optional<String> assetBaseDir = Optional.absent();
+        String baseDir = ninjaProperties.get( "application.static.asset.basedir");
```
raphaelbauer: Can we make "application.static.asset.basedir" a static final String somewhere?

raphaelbauer:
Hey @sojin,

this is cool! I added some tiny remarks. Maybe they make sense to you... Would be nice if you could also add your change to changelog.md. I'll merge your stuff then.

Thanks!

Raphael

sojin:
Hi @raphaelbauer ,

Thanks for the comments. Updated the pull request with necessary changes.

For the getNormalizedAssetPath(), I've used FilenameUtils.normalizeNoEndSeparator(). 
My intention was to remove the trailing slashes and normalize the path so as to have ninja behaves as expected irrespective of what user inputs.

Files.simplifyPath() doesn't work with windows style separator and it doesn't remove the trailing slashes from Linux style path (although it normalize them). 

Thanks
Sojin

raphaelbauer:
Perfect :+1: 
Thanks! It's already in develop.

PR #156, opened 2014-03-11
Modify NinjaRunner to make use of injector's built-in injectMembers(obj)

smallufo:
injector.injectMembers(testObj) takes cares of all my previous work.
Sorry I was not familiar with Guice's injector.


raphaelbauer:
ah nice! I did not think about that... ;) Obviously much better and less code :)

PR #159, opened 2014-03-17
Fix changelog dates

cwildfoerster:
I've noticed that the dates within the changelog are wrong, so i fixed them.


raphaelbauer:
this was not the changelog, but the roadmap... ;) just kidding... no idea how we managed to mess that up :) thanks for fixing that!

PR #160, opened 2014-03-25
Add Generic Type Information into ArgumentExtractor and Make core configuration customizable

gakuzzzz:
Hi,

Thank you for awesome framework!
### I want to do
1. use generic types (ex. `Optional<ID<User>>`) in `Controller` method arguments.
2. Using javac option `-parameters` in JavaSE8, associate `Controller` method arguments with request parameter names or path parameter names.
   - this function needs JavaSE8. so It is enough, if the function is provided with plug-in.
3. (optional) make framework user enable to custom the behavior when `@PathParam` or `@Param` is ommited
### Problems
1. `ArgumentExtractor` can not use generic type information since `ArgumentClassHolder` has only `java.lang.Class<?>`
2. `ArgumentExtractor` can not access `Method` instance and the index of parameters.
3. Since [ControllerMethodInvoker.build](https://github.com/ninjaframework/ninja/blob/ninja-3.1.3/ninja-core/src/main/java/ninja/params/ControllerMethodInvoker.java#L69) is static method, it can not change in configuration.
4. It can not override `ninja.Configuration` bindings.
### Requests

This PR contains below. but, there are more better solutions, teach me them please.
1. Add `java.lang.reflect.Type` into `ArgumentClassHolder`.
2. [In `ArgumentExtractor` construction](https://github.com/ninjaframework/ninja/blob/ninja-3.1.3/ninja-core/src/main/java/ninja/params/ControllerMethodInvoker.java#L238-L244), the child injector binds `Method` instance and the index of parameters.
3. Create `ControllerMethodInvokerBuilder` and Remove the static `ControllerMethodInvoker.build`.
4. Make `ninja.Configuration` enable to oberride by defining `conf.OverrideCoreModule`

best regards,


raphaelbauer:
Wow! that looks awesome!

Unfortunately I am a bit busy right now - therefore I cannot review everything right away. Please expect some feedback latest end of next week...

Best,

Raphael

gakuzzzz:
Alright, You don't need to feel obligated :smile:

ninja-servlet/src/main/java/ninja/servlet/NinjaBootstrap.java
```diff
@@ -47,6 +48,7 @@
     private Logger logger = LoggerFactory.getLogger(NinjaBootstrap.class);
 
     private static final String APPLICATION_GUICE_MODULE_CONVENTION_LOCATION = "conf.Module";
+    private static final String APPLICATION_GUICE_OVERRIDE_CORE_MODULE_CONVENTION_LOCATION = "conf.OverrideCoreModule";
```
raphaelbauer, line 51: Do we really need this?
Wouldn't it be better to just use conf.Module, and inialize it via Modules.override?

That way we would not need to add a new configuration module + documentation + maintenance...

raphaelbauer:
Hey...
I think that is pretty cool...

One thing is missing: could you add your change into changelod.md, and add your name to team.md?

And maybe we can go without the override module (As I commented).

danielsawan:
i think this can help solving the problem of this issue https://github.com/ninjaframework/ninja/issues/367 :)

PR #163, opened 2014-04-05
updated documentation for sending mails

svenkubiak:
The sendMail example did expect the return of Result, which did not feel quit correct as their is no obvious connection to Freemarker. Also I added a try-catch for the postoffice.send function, which makes the example complete.


raphaelbauer:
Thanks!

PR #169, opened 2014-04-23
add possibility to define custom package for application module and routes

avarabyeu:
As a part of https://github.com/ninjaframework/ninja/issues/61. Pretty light change, shouldn't has impact on core module and plugins.

### DIFF
--- a/ninja-servlet/src/main/java/ninja/servlet/NinjaBootstrap.java
+++ b/ninja-servlet/src/main/java/ninja/servlet/NinjaBootstrap.java
@@ -25,9 +25,10 @@ import ninja.Ninja;
 import ninja.Router;
 import ninja.application.ApplicationRoutes;
 import ninja.lifecycle.LifecycleSupport;
+import ninja.logging.LogbackConfigurator;
 import ninja.scheduler.SchedulerSupport;
+import ninja.utils.NinjaConstant;
 import ninja.utils.NinjaPropertiesImpl;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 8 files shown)


avarabyeu:
Yes, my fault. Missed one file in commit

ninja-servlet/src/main/java/ninja/servlet/NinjaBootstrap.java
```diff
@@ -63,33 +64,33 @@ public Injector getInjector() {
...
     public synchronized void boot() {
-        
+
         initLogbackIfLogbackIsOnTheClassPath();
-        
+
         if (injector != null) {
             throw new RuntimeException("NinjaBootstap already booted");
         }
-
```
raphaelbauer, line 72: thanks for cleaning up the whitespaces :)

raphaelbauer:
Hey avarabyeu,

thanks for implementing this :)
There is one thing I do not understand properly: To make your feature work you still need conf/application.conf to define the path for the module. Right?

If that's the case - what is the benefit of moving Module.java into any directory, when application.conf is still nailed in /conf/.?

Should not sound rude :) I am just trying to understand your motivation! Why did you for instance not consider using a system property to make that happen?

Thanks!

Raphael

avarabyeu:
Raphael, the main idea is to move compiled sources to application packages (to follow java conventions). As for configuration files, i see no problems to keep them in "conf" folder :) What do you think? Just because it's little bit unusual to keep java sources with no packages...

raphaelbauer:
Hey avarabyeu,

I think we can include that change :) Your code looks nice, and the test is really important :) Thanks for your contribution. There are only some final things missing. 
1. Add your feature to the documentation. Without that nobody will know about it. Docs is at ninja-core/src/site/markdown.
2. Add your changes to changelog.md and your name to team.md.
3. Sign the contributor agreement (electronically) at https://docs.google.com/forms/d/1Yasrxa17kYfaNDgbRvn77rM3WTnU_Um0rwz3GfzPp9g/viewform

That's it already :)

Thanks for improving Ninja!

Cheers,

Raphael

avarabyeu:
Thanks, Raphael!

raphaelbauer:
I am a bit busy right now. Give me some days to check it out...

avarabyeu:
That's ok, keep your time, Raphael)

PR #174, opened 2014-05-08
Added a getHostname() method to the context.  

mattjonesorg:
This is helpful for supporting multi-tenant applications where each customer has their own hostname.

Today, it is as simple as context.getHeader("Host"), but if the context already has many other wrappers around headers, I thought this would be helpful.

### DIFF
--- a/ninja-servlet/src/main/java/ninja/servlet/ContextImpl.java
+++ b/ninja-servlet/src/main/java/ninja/servlet/ContextImpl.java
@@ -327,7 +327,9 @@ public class ContextImpl implements Context.Impl {
     }
 
     @Override
-    public String getHostname() { return httpServletRequest.getHeader("host"); }
+    public String getHostname() {
+        return httpServletRequest.getHeader("host");
+    }
 
     public void handleAsync() {
         synchronized (asyncLock) {
--- a/ninja-servlet/src/main/java/ninja/servlet/ContextImpl.java
+++ b/ninja-servlet/src/main/java/ninja/servlet/ContextImpl.java
@@ -326,6 +326,9 @@ public class ContextImpl implements Context.Impl {
         return httpServletRequest.getRequestURI();
     }
 
+    @Override
+    public String getHostname() { return httpServletRequest.getHeader("host"); }
+
     public void handleAsync() {
         synchronized (asyncLock) {
             if (asyncStrategy == null) {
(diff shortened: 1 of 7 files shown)


ninja-servlet/src/main/java/ninja/servlet/ContextImpl.java
```diff
@@ -326,6 +326,9 @@ public String getRequestUri() {
         return httpServletRequest.getRequestURI();
     }
 
+    @Override
+    public String getHostname() { return httpServletRequest.getHeader("host"); }
```
raphaelbauer: could you please reformat that to something like 

methodName {
...
} (like in getHostname of wrappedContext)

I also like one liners, but that's just the general way how the code in Ninja is written. Makes it simpler to maintain everything in the long run :)

ninja-test-utilities/src/main/java/ninja/utils/FakeContext.java
```diff
@@ -108,6 +109,12 @@ public String getRequestUri() {
         return requestUri;
     }
 
+    @Override
+    public String getHostname() {
+        return hostname;
+    }
+    public void setHostname(String hostname) {this.hostname = hostname;}
```
raphaelbauer: Same here... three lines instead of one :) please :)

raphaelbauer:
Nice!

three simple things are missing:
- add your change to changelog.md
- add yourself to team.md
  and sign the contributor's agreement here: https://docs.google.com/forms/d/1Yasrxa17kYfaNDgbRvn77rM3WTnU_Um0rwz3GfzPp9g/viewform

More general information is here: http://www.ninjaframework.org/developer/getting_started.html

Thanks!

ninja-test-utilities/src/main/java/ninja/utils/FakeContext.java
mattjonesorg: I was trying to be consistent with similar methods, such as getAcceptEncoding(), which is a one-liner.  But I'll commit the change to make it three.  :)

raphaelbauer: Thanks! (Don't take it personally... it's just my experience that over the long run those "small" things really help to keep the codebase in shape. ... ;) ).

mattjonesorg: Nothing taken personally!  :)  I'm just glad I can contribute!

mattjonesorg:
All steps completed!  Thanks for developing such a great framework!

raphaelbauer:
Thanks! The framework is a community effort - contributions like this one make it what it is :)

PR #175, opened 2014-05-11
fixed flash documentation

svenkubiak:
This looks like a copy&paste error, but "flash_" shoud be "flash.". Also the correct syntax for object and parameter check in freemarker is (foo.bar)??


raphaelbauer:
Very very good. Thanks for the improvement!

PR #176, opened 2014-05-16
added http status 401, organized imports

svenkubiak:
I have added HTTP Status 401 "Unauthorized" which was missing and did an automatic "organized imports".


raphaelbauer:
Nice! Thanks!

PR #180, opened 2014-05-21
Added HTTP 201 (Created), convenience methods for codes 201 and 401, and made HTTP status code variables final

metacity:
HTTP spec doesn't actually require (only "should include") a Location header in a 201 Created response, hence optional url..


raphaelbauer:
very useful! thanks!

PR #182, opened 2014-05-31
Allowing more path entries in Javascript function path for JSONP.

nobullet:
I followed the instructions to render JSONP (http://www.ninjaframework.org/documentation/working_with_json_jsonp.html), but I have a problem with my callback function.

When I pass ?callback=MyApp.Path.myCallback123, it does not produce:

MyApp.Path.myCallback123({'response': 'data'})

instead it produces:

onResponse({'response': 'data'})

When I pass ?callback=MyAppPath.myCallback123 (removing first '.', or second), it works fine.

In my case, I need to pass a parameter like "callback=Ext.data.JsonP.callback4". Here, I don't want to override all related classes in ExtJS, so how can I solve this problem with Ninja?


raphaelbauer:
Hey nobullet :)

Really cool. Thanks! Could you add a testcase for the regex. Maybe by extracting the security matcher to an external static helper class and testing that in turn?

It would be also great if you could add a tiny comment to the source code about the rationale of the security test.

I guess you rationale is this: http://tav.espians.com/sanitising-jsonp-callback-identifiers-for-security.html

And therefore good testcases for the matcher would be those (also mentioned in the article):
Here are some examples:

> > > is_valid_jsonp_callback_value('somefunction')
> > >   True
> > > 
> > > is_valid_jsonp_callback_value('someobject.somemethod')
> > >   True
> > > 
> > > is_valid_jsonp_callback_value('somearray[12345]')
> > >   True
> > >   It accepts all valid Javascript identifiers as defined in the ECMAScript   specification:
> > > 
> > > is_valid_jsonp_callback_value(u'Stra\u00dfe')
> > >   True
> > > 
> > > is_valid_jsonp_callback_value(r'\u0062oo')
> > >   True
> > > 
> > > is_valid_jsonp_callback_value('$42.ajaxHandler')
> > >   True
> > >   But disallows everything else:
> > > 
> > > is_valid_jsonp_callback_value('alert(document.cookie)')
> > >   False
> > > 
> > > is_valid_jsonp_callback_value(r'\u0020')
> > >   False

This'd really help to make your great jsonp feature maintainable in the long run :)

Does that make sense to you?

If there is anything where I can help please let me know!

Thanks!

Raphael

nobullet:
Please review the CL again: 
1. changed regexp alittle
2. testing the static helper method instead of regexp
3. added javadoc for the whole class with description and links
4. more tests for the validator

raphaelbauer:
Absolutely perfect! Thanks!

PR #186, opened 2014-06-09
Fixed a wrong variable when logging

metacity:


raphaelbauer:
very good catch. Thanks! 

Can you add your fix to changelog.md and yourself to team.md (If you haven't done so). Helps us to keep track of what is happening. (Even if that "simple" fix may be an edge case).

I'll merge the fix asap...!

raphaelbauer:
Just got some time to do the merge and cut a new release...
Thanks for the fix!
Really great and important - those things look tiny but are a huge help :)

PR #189, opened 2014-06-10
Use SecureRandom instead of Random to generate the app secret

metacity:
.. and use StringBuilder instead of StringBuffer (no synchronization needed).

### DIFF
--- a/ninja-core/src/main/java/ninja/utils/SecretGenerator.java
+++ b/ninja-core/src/main/java/ninja/utils/SecretGenerator.java
@@ -16,6 +16,7 @@
 
 package ninja.utils;
 
+import java.security.SecureRandom;
 import java.util.Random;
 
 
@@ -29,7 +30,7 @@ public class SecretGenerator {
      */
     public static String generateSecret() {
         
-        return generateSecret(new Random());
+        return generateSecret(new SecureRandom());
 
     }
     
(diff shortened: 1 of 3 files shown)


ninja-core/src/main/java/ninja/utils/SecretGenerator.java
```diff
@@ -46,16 +47,16 @@ protected static String generateSecret(Random random) {
         
         String charsetForSecret = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
         
-        StringBuffer stringBuffer = new StringBuffer();
+        StringBuilder stringBuilder = new StringBuilder(64);
```
raphaelbauer, line 50: very good catch using StringBuilder. The whole thing is threadsafe inside the method - so there is no reason in using StringBuffer...

ninja-core/src/main/java/ninja/utils/SecretGenerator.java
```diff
@@ -29,7 +30,7 @@
      */
     public static String generateSecret() {
         
-        return generateSecret(new Random());
+        return generateSecret(new SecureRandom());
```
raphaelbauer, line 33: Very good! Long overdue...

raphaelbauer:
Really important and nice PR!

PR #191, opened 2014-06-30
Catch Exception when not found any user

anhtuank7c:
When submit not existing user, line 27 will get an exception:
javax.persistence.NoResultException: No entity found for query
I just wrap block code into try{}catch block.


raphaelbauer:
Thanks! good catch :)

PR #195, opened 2014-07-21
Optional annotation-based route definition and reverse routing improvements

gitblit:
While I think the Ninja framework has great potential, I am not a fan of the fragility of the routing mechanism.  I agree with the concept of the single-file Routes registration approach, but in practice the dependence of a _named_ method is a major weakness.  Perhaps this can be improved with Java 8, but not likely all aspects of routing will be improved by that change.

This pull request gives developers the choice of how to register their routes:
1. single-file route definitions using discrete HTTP method / named controller method registration (`GET()...with(UsersController.class, "index")`)
2. annotation-based route definitions within the controller (`register(UsersController.class)`)

``` java
public class UsersController {

    @RouteDef(uri="/users")
    public Result index() {
        ...
    }

    @RouteDef(uri="/users/{username}", order=2)
    public Result getUser(@PathParam("username") String username) {
        ...
    }

    @RouteDef(uri="/users/{username}", method=HttpMethod.PUT)
    public Result updateUser(@PathParam("username") String username) {
        ...
    }
}
```

Additionally, this pull request implements _best-fit_ reverse routing methods for the requested HTTP method where **best-fit** is defined as the matching route that has the fewest query parameters.  This allows the most common use cases for reverse route generation to work without specifying the controller method  name.

``` java
// returns /users
String indexRoute = router.getReverseGET(UsersController.class);
// returns /users/james
String userRoute = router.getReverseGET(UsersController.class, "username", "james");
```

This pull request also fixes a potential null pointer exception which could occur during reverse route generation for a route that failed registration.

I can split this PR into multiple PRs, if that is your preference.

### DIFF
--- a/ninja-core/src/main/java/ninja/Route.java
+++ b/ninja-core/src/main/java/ninja/Route.java
@@ -32,6 +32,16 @@ import java.util.regex.Pattern;
  */
 public class Route {
 
+    public static class HttpMethod {
+        public final static String OPTIONS = "OPTIONS";
+        public final static String HEAD = "HEAD";
+        public final static String GET = "GET";
+        public final static String PUT = "PUT";
+        public final static String POST = "POST";
+        public final static String DELETE = "DELETE";
+    }
+
+
     //Matches: {id} AND {id: .*?}
     // group(1) extracts the name of the group (in that case "id").
     // group(3) extracts the regex if defined
--- a/ninja-core/src/main/java/ninja/RouteBuilderImpl.java
+++ b/ninja-core/src/main/java/ninja/RouteBuilderImpl.java
@@ -175,6 +175,10 @@ class RouteBuilderImpl implements RouteBuilder {
         // Calculate filters
         LinkedList<Class<? extends Filter>> filters = new LinkedList<Class<? extends Filter>>();
         if(controller != null) {
+            if (controllerMethod == null) {
+                throw new IllegalStateException(
+                        "Route does not have a controller method");
+            }
             filters.addAll(calculateFiltersForClass(controller));
             FilterWith filterWith = controllerMethod
                     .getAnnotation(FilterWith.class);
--- a/ninja-core/src/main/java/ninja/RouteBuilderImpl.java
+++ b/ninja-core/src/main/java/ninja/RouteBuilderImpl.java
@@ -22,6 +22,7 @@ import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.Set;
 
+import ninja.Route.HttpMethod;
 import ninja.params.ControllerMethodInvoker;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -40,32 +41,32 @@ class RouteBuilderImpl implements RouteBuilder {
     private Result result;
 
     public RouteBuilderImpl GET() {
-        httpMethod = "GET";
+        httpMethod = HttpMethod.GET;
         return this;
     }
 
     public RouteBuilderImpl POST() {
-        httpMethod = "POST";
+        httpMethod = HttpMethod.POST;
         return this;
     }
 
     public RouteBuilderImpl PUT() {
-        httpMethod = "PUT";
+        httpMethod = HttpMethod.PUT;
         return this;
     }
 
     public RouteBuilderImpl DELETE() {
-        httpMethod = "DELETE";
+        httpMethod = HttpMethod.DELETE;
         return this;
     }
 
     public RouteBuilderImpl OPTIONS() {
-        httpMethod = "OPTIONS";
+        httpMethod = HttpMethod.OPTIONS;
         return this;
     }
     
     public RouteBuilderImpl HEAD() {
-        httpMethod = "HEAD";
+        httpMethod = HttpMethod.HEAD;
         return this;
     }
         
--- a/ninja-core/src/main/java/ninja/RouteDef.java
+++ b/ninja-core/src/main/java/ninja/RouteDef.java
@@ -1,0 +1,62 @@
+/**
+ * Copyright (C) 2012-2014 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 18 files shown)


gitblit:
Hmm.  I didn't notice the `ninja-servlet` tests before.... but it doesn't look like they were working before my changes anyway.  Looks like a package name error.

gitblit:
Fixed the bug in the `Log the resolved routes on Ninja startup` commit.

raphaelbauer:
Hi @gitblit,

wow - that's a great PR. You even included docs. Thanks for your contribution!

The team is currently working on Ninja 4. One of the additions is annotation based routing. Our prototypic implementation is here: https://github.com/raphaelbauer/ninja/tree/ninja4/ninja-annotation-router

The main difference to your approach (apart from the fact that there's no documentation at ninja4 and just a rudimentary set of testcases) - is that ninja4 uses classworlds to scan for controllers that are annotated - and that the annotation-router is designed as module.

Not sure how we can handle this. Would you mind having a look at our branch. It would be awesome if we could integrate both changes. What do you think?

Thanks,

Raphael

gitblit:
Interesting.  I wasn't aware new development was happening in a separate repo.

In your module approach... do I have to pick using `AnnotatedRouter` or `RouterImpl` up front or can I use both at the same time?

I had a quick look at the `AnnotatedRouter` and IMO my implementation is much more complete and flexible.  The two main advantages of the Ninja4 prototype is the classpath scanning so you never have to manually register controllers and _(I will assume)_ that `org.reflections` returns annotated methods in their defined order which would eliminate my _order_ attribute.  Assuming I understand `org.reflections` correctly, then we've moved the route registration ordering problem from the method level to the controller class level as we don't have a deterministic class registration order.

If automatic registration is desirable, I can integrate that based on `org.reflections` but my instinct would be to base it off a new annotation that accounts for runtime mode & registration order rather than just a package setting in Ninja properties.

Something like:

``` java
@Target({ ElementType.TYPE })
@Retention(RetentionPolicy.RUNTIME)
@Documented
public @interface Controller {
    NinjaMode mode() default NinjaMode.prod;
    int order() default 1;
}
```

However, I'm not sure automatic controller registration is an improvement since we have to manually specify registration order for every controller.  Alternatively, I could use `org.reflections` _just_ for method ordering to eliminate the _order_ attribute.  That is up to you.

As for the module design, the benefit is unclear to me.  Is it testability?

raphaelbauer:
First of all Ninja development happens in the the Ninja repo. The ninja4 branch I mentioned is just a sketch of ideas to work out stuff that is missing. It has been developed based on ideas by clients and has some chance to become real - so that stuff might or might not end up in an official N4 release.

Regarding module design. This is super important. If you develop a great module you don't want to be dependent on Ninja's release cycle. On the other hand Ninja has to become an even better platform to support a huge amount of usecases - without the need to incorporate everything into the core. That way ninja stays maintainable and is understandable for people that want to contribute.

Apart from that - and before we continue the discussion.

Your PR contains a lot of great stuff that must go into the core. That is everything except the register method and the routes annotation @Def and so on.
I'd like to merge eg:
- getReverse*
- TemplateEngineFreemarker*
- tests
- bugfix against the NPE

Could you make a separate PR so I can merge it cleanly? Please don't forget to sign the contributor's agreement https://docs.google.com/forms/d/1Yasrxa17kYfaNDgbRvn77rM3WTnU_Um0rwz3GfzPp9g/viewform

If that's done we can continue the discussion how we can add the route based stuff into the core... Should give us (or me) some more focus on the matters :) I hope we can integrate both approaches so everyone is happy :)

Thanks for your efforts!

gitblit:
I will split this into multiple PRs.  Do you want one per change/commit (Gerrit style) or just one PR with the non-annotation-related changes?

I agree with modular design.  I'm less thrilled with out-of-band maintenance (ninja-mustache seems to be abandoned without artifacts on Central, ninja-morphia has disappeared).  My preference would be to have a /ninja-metrics/ module in this repo alongside /ninja-core/, /ninja-maven-plugin/, etc.  That keeps the metrics module up-to-date with whatever changes happen to the core and artifacts get pushed up during release.

What features do you expect to keep in the ninja-core module?

I see things like FreeMarker, JPA, and Flyaway as candidates for modules.  The router seems like a core item, though - unless you are splitting the router into two pieces: registration & querying/matching.  If you do that, then a pluggable registrar makes sense.

raphaelbauer:
PR: Just one per all annotation related changes.

Your list is quite right: Freemarker, JPA and Flyway could become separate modules (JPA and Flyway should - and we should have a connection pooling basis). The router is part of the core - you can of course replace the router with your own implementation if you bin your own router during startup.

ninja-metrics actually sounds like a good idea... I really love the metrics lib... 

Regarding routes: What about using only reflections based auto discovery as outlined in ninja4 branch? (+ordering support)?

Would it be possible to have a hangout about some topics? I'd invite Darren who is working on similar features right now. We should team up...

gitblit:
Definitely.  Let me break this PR apart (and a small piece of the metrics PR) and get those merged.  Then we can do a Hangout to discuss annotations and how to proceed with Metrics.

I've been thinking more about maintaining a deterministic route order and I'm not sure if that is really necessary.  Traditionally, web.xml gives you deterministic ordering.  OTOH, annotation-based Servlet3 is non-deterministic AFAIK.  And if I look at Jersey-based solutions, I don't think they register routes deterministically either. My current thinking is route ordering is perhaps irrelevant when coupled with a smart url dispatcher/matcher/router.  Perhaps something like the _best-fit_ url matcher approach included in this PR for reverse routing.

This is a long-winded way to say that I think reflections-based auto discovery will fly as long as the url dispatcher/router gets smarter.  It may make sense to break the existing Router into a Registrar and a Dispatcher.

raphaelbauer:
Great!

Could you please send me an email so I can arrange a hangout session (Email is on my gh page)? There is at least one other team member that will join us...

Looking forward to your PRs!

gitblit:
This PR is obsolete.

PR #196, opened 2014-07-22
Add integrated Metrics support

gitblit:
This is a basic integration of [Metrics](http://metrics.codahale.com) support.  It currently collects and exposes Route and JVM metrics over JMX (install MBean plugin in VisualVM).  Individual controller methods may be annotated with `@Timed` for the integration to collect and report metrics.

This PR will create a merge conflict with #195 since they both implement a `getRoutes()` method for the Router.  Assuming one or the other is merged, I will rebase.

### DIFF
--- a/ninja-core/src/site/markdown/documentation/basic_concepts/metrics.md
+++ b/ninja-core/src/site/markdown/documentation/basic_concepts/metrics.md
@@ -171,6 +171,40 @@ Add the following settings to your `application.conf`.
     metrics.ganglia.address = ganglia.example.com
     metrics.ganglia.port = 8649
 
+### Reporting Metrics to InfluxDB
+
+Ninja Metrics supports reporting to [InfluxDB](http://influxdb.com).
+
+Add the following dependency to your application `pom.xml`.
+
+    <dependency>
+        <groupId>org.ninjaframework</groupId>
+        <artifactId>ninja-metrics-influxdb</artifactId>
+        <version>${ninja.version}</version>
+    </dependency>
+
+Bind the InfluxDB integration in your `Module.java`.
+
+<pre class="prettyprint">
+@Singleton
+public class Module extends AbstractModule {
+
+  @Override
+  protected void configure() {
+    bind(NinjaInfluxDB.class);
+  }
+}
+</pre>
+
+Add the following settings to your `application.conf`.
+
+    metrics.influxdb.enabled = true
+    metrics.influxdb.address = localhost
+    metrics.influxdb.port = 8086
+    metrics.influxdb.database = mydb
+    metrics.influxdb.username = root
+    metrics.influxdb.password = root
+
 ### Reporting Metrics to Librato
 
 [Librato](http://metrics.librato.com) is a cloud-based metrics database and dashboard service.
--- a/ninja-core/src/site/markdown/documentation/basic_concepts/metrics.md
+++ b/ninja-core/src/site/markdown/documentation/basic_concepts/metrics.md
@@ -1,0 +1,94 @@
+Metrics
+==============
+
+Ninja provides optional integration with [Metrics](http://metrics.dropwizard.io/) for measuring the responsiveness of your application.
+
+Setup
+------------------
+
+### Add the Ninja-Metrics dependency
+
+    <dependency>
+        <groupId>org.ninjaframework</groupId>
+        <artifactId>ninja-metrics</artifactId>
+        <version>${ninja.version}</version>
+    </dependency>
+
+
+### Install the `MetricsModule` in your `conf.Module` class.
+
+<pre class="prettyprint">
+@Singleton
+public class Module extends AbstractModule {
+
+    @Override
+    protected void configure() {
+
+        install(new MetricsModule());
+
+    }
+}
+</pre>
+
+### Subclass InstrumentedNinja in your `conf.Ninja` class.
+
+<pre class="prettyprint">
+package conf;
+
+import ninja.metrics.InstrumentedNinja;
+
+public class Ninja extends InstrumentedNinja {
+
+}
+
+</pre>
+
+### Collecting Route Metrics
+
+Now you are ready to start annotating your controllers.
+
+You have three choices in the collection of route metrics:
+
+- *Do nothing*
+- *Metered*
+A meter measures the rate of events over time (e.g., “requests per second”). In addition to the mean rate, meters also track 1-, 5-, and 15-minute moving averages.
+- *Timed*
+A timer measures both the rate that a particular piece of code is called and the distribution of its duration.
+
+1. Start by sprinkling `@Metered` or `@Timed` on some of your controller methods.
+2. Start up VisualVM (and install the MBeans plugin) or JConsole.
+3. Browse your app and refresh the collected metrics.
+
+<pre class="prettyprint">
+package controllers;
+
+@Singleton
+public class AppController {
+
+    @Timed
+    public Result index() {
+
+        return Results.html();
+    }
+
+}
+</pre>
+
+### Collecting Cache Metrics
+
+If you want to instrument your NinjaCache, Ninja-Metrics supports instrumenting both the EhCache and the Memcached implementations.
+
+In your `application.conf` file specify:
+
+    cache.implementation = ninja.metrics.InstrumentedEhCache
+
+or
+
+    cache.implementation = ninja.metrics.InstrumentedMemcached
+
+Reporting
+------------------
+
+By default, Ninja Metrics will expose your metrics over JMX. You may disable this behavior by setting *metrics.jmx=false* in your `application.conf` file.
+
+You can view the collected metrics using VisualVM (with the MBeans plugin installed) or using JConsole.
--- a/ninja-core/src/site/markdown/documentation/basic_concepts/metrics.md
+++ b/ninja-core/src/site/markdown/documentation/basic_concepts/metrics.md
@@ -30,7 +30,7 @@ public class Module extends AbstractModule {
 }
 </pre>
 
-### Subclass InstrumentedNinja in your `conf.Ninja` class.
+### Subclass InstrumentedNinja in your `conf.Ninja` class. (optional)
 
 <pre class="prettyprint">
 package conf;
@@ -86,9 +86,32 @@ or
 
     cache.implementation = ninja.metrics.InstrumentedMemcached
 
-Reporting
-------------------
+### Collecting Additional Metrics
 
-By default, Ninja Metrics will expose your metrics over JMX. You may disable this behavior by setting *metrics.jmx=false* in your `application.conf` file.
+#### JVM Metrics
+
+You may optionally enable JVM-level details reporting by setting *metrics.jvm.enabled=true* in your `application.conf` file.
+
+    metrics.jvm.enabled = true
+
+#### Logback Metrics
+
+You may optionally enable reporting of Logback log-level counts by setting *metrics.logback.enabled=true* in your `application.conf` file.
+
+    metrics.logback.enabled = true
+
+### Reporting Metrics via JMX for VisualVM/JConsole
+
+### Exposing Metrics via JMX for VisualVM/JConsole
+
+By default, Ninja Metrics will expose your metrics over JMX. You may disable this behavior by setting *metrics.jmx.enable=false* in your `application.conf` file.
 
 You can view the collected metrics using VisualVM (with the MBeans plugin installed) or using JConsole.
+
+
+### Reporting Metrics to 3rd party services
+
+It is also possible to collect & report your Metrics data to several cloud-based services such as:
+
+- [Librato](http://www.librato.com)
+- [DataDog](http://www.datadoghq.com)
--- a/ninja-core/src/site/markdown/documentation/basic_concepts/metrics.md
+++ b/ninja-core/src/site/markdown/documentation/basic_concepts/metrics.md
@@ -102,12 +102,28 @@ You may optionally enable reporting of Logback log-level counts by setting *metr
 
 ### Reporting Metrics via JMX for VisualVM/JConsole
 
-### Exposing Metrics via JMX for VisualVM/JConsole
+By default, Ninja Metrics will expose your metrics over JMX. You may disable this behavior by setting *metrics.jmx.enabled=false* in your `application.conf` file.
 
-By default, Ninja Metrics will expose your metrics over JMX. You may disable this behavior by setting *metrics.jmx.enable=false* in your `application.conf` file.
+    metrics.jmx.enabled = true
 
 You can view the collected metrics using VisualVM (with the MBeans plugin installed) or using JConsole.
 
+### Reporting Metrics to Graphite
+
+Ninja Metrics supports reporting to Graphite.
+
+    metrics.graphite.enabled = true
+    metrics.graphite.address = graphite.example.com
+    metrics.graphite.port = 2003
+    metrics.graphite.pickled = false
+
+### Reporting Metrics to Ganglia
+
+Ninja Metrics supports reporting to Ganglia.
+
+    metrics.ganglia.enabled = true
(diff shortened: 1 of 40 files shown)


gitblit:
Note: There are other things to explore with this integration.  For example, there is an ehCache interface to Metrics - but it is out of date.  It is built against 2.6.6 which has API changes compared to 2.8.x.

gitblit:
I backed-off the JVM reporting because of unnecessary unit testing complexity.  It's better, actually, since out-of-the-box it will only report what you register - currently routes.

raphaelbauer:
Hi James,

looks great! You asked in the other PR, why our sketches are added as modules. It's not because of testability. The general consensus is that ninja-core should only contain the core. And be as extensible as possible. That way we get both flexibility for users and don't get too dependent on 3rd party libs. And the release cycle of external modules (your metrics module) are also completely independent of Ninja's release cycle.

In case of your Metrics addition: It would be awesome if you could create a module and push that to maven central. Afaics will work just fine using a NinjaFilter. That way we don't have to add any complexity to ninja-core and still get all benefits of Metrics (+ you are not dependent on Ninja's release cycle for additions/fixes and so on). What do you think?

If there are some missing things (like hooking into router request and so on) let's discuss that and make Ninja as extensible as possible. We are discussing some ideas here and it would be awesome if we could work towards a win-win situation :)

Best,

Raphael

gitblit:
I have rewritten and rebased this PR to be based on two new modules:
1. ninja-annotations
2. ninja-metrics

To be completely effective (and agnostic to the underlying http engine), the metrics module needs special consideration in the ninja-core.  For example, I want to automatically wrap Cache instances with InstrumentedCache instances and I want to use InstrumentedNinja instead of DefaultNinja, if it is present and the application does not specify it's own Ninja class.  I also needed to automatically load the Ninja Metrics module so I introduced Reflections and a @NinjaModule annotation to handle that.

This is all subject to change as needed.

To try this out:
1. `mvn install` in the repository root
2. `mvn ninja:run` in the ninja-servlet-integration-test
3. Startup VisualVM
4. Install the MBeans plugin through the Tools->Plugins  menu
5. Drill into the NinjaJetty instance listed in VisualVM and select the _MBeans_ tab
6. Browse the test app and click on the different available _ninja.cache_ and _ninja.requests_ metrics

Example screenshot attached .

ninja-metrics

gitblit:
That CI test failure is a false positive. Everything tested fine.  Failure is something else.

gitblit:
I've rebased this PR on the current HEAD of develop and I have simplified it.  It is now an optional & standlone module that I'd like to see in upstream Ninja instead of sitting around in my repos.  :)
1. I've discarded the changes to core to automatically instrument everything
2. I've discard the stubs for HealthCheck support as I do not have a plan for that at the moment
3. I've updated to the current Metrics release 3.1.0 which has been renamed for DropWizard.

The servlet integration test is instrumented for metrics so if you run that app & startup a tool like VisualVM you can inspect collected metrics.

raphaelbauer:
hey gitblit... that looks awesome :) I really want that! I nevertheless wanted to go through the changes in detail and think about some things... Will do within the next week... Thanks for your efforts!

raphaelbauer:
Hey gitblit,

some feedback - but nothing to do on your side - I'll propose an implementation the next days:
- tests are missing (not your problem - there were no tests for the cache classes before - that's embarrassing, but I'll change that).
- I'll add graphite support to the core - all our servers are using graphite, so I need that anyway
- I'll try to rewrite the "getRouteMetric" part. afaics we can use guice interceptors to that end. The benefits include a bit cleaner code in the Ninja class - and the ability to measure all methods - not only route methods :)

I'll tackle these 3 things (likely on monday) and get back to you :) Any feedback always welcome! I hope there is nothing I have missed :) Thanks for that great contribution!

Cheers,

Raphael

gitblit:
Sounds good.  Looking forward to seeing your version.

raphaelbauer:
Just wondering - but is there a special reason why you are using three different registries? Would it be ok to use just one with different prefix?

gitblit:
Take a look at the above screenshot of VisualVM.  I wanted to keep the metrics organized.  Cache metrics are discrete and separate from Routing metrics which are discrete and separate from general purpose App metrics - which I would advocate is where your Guice Method Interceptor should log non-controller and non-cache metrics.

Lumping them all together may lead to a mess in sifting through data.  That depends on the JMX client, for sure. I didn't play with JConsole - maybe it's better.

raphaelbauer:
ah okay... I see... in graphite it's a bit different...

raphaelbauer:
Can you have a look at https://github.com/gitblit/ninja/pull/1 (not finished)? Feedback appreciated!

gitblit:
Anything you want me to do to move this forward?  Or have I successfully handed this off to you?

raphaelbauer:
Actually that was one of the things I wanted to discuss with you... My most important question would be if my PR (gitblit#1) does what you expect it to do - especially because I deleted a lot of your code and replaced it with the two interceptors... If that is ok I'd be happy to finalize everything and testcases and so on...

gitblit:
The interceptors are really cool.  They don't measure exactly the same thing.  My version measures the http request from beginning to end - including template/serializer rendering.  Your version measures controller method execution alone.  Both are useful but I guess we can only pick one.

I've been using the end-to-end timing in evaluating the 5 template engines I have.  I have a prototype real-world app and I run the app with each engine over a series of the same pages.  That's pretty useful in evaluating them.

Most people will likely only care about the timings of their specific controller methods - and I probably will too once I choose an engine.

So I think we should move ahead with your interceptor approach.  Slash away.

raphaelbauer:
Great..

Btw. You can always annotate the onRoutesRequest method in you conf.Ninja... should work and measure the "whole" request cycle (it still misses everything that happens before, but that's a completely different story).

raphaelbauer:
Ah... and much more important... What do the microbenchmarks tell? Which templating engines work well? Which ones do suck?

gitblit:
Assuming similar features of the template engines (includes, inheritance, etc) and similar integrations (assetsAt, webJarsAt, i18n, prettyTime), the timing-based benchmarks generally tell me how fast each engine is at rendering the same set of complete pages which use all of the aforementioned features and integrations.  Each of the generated pages also includes file-system retrieved list content so it exercises the template engine's iterator functions as well - 50 items per page.

My findings so far.

 **Dev Mode**
Best performers (all fairly close, but Pebble is the consistent winner):
- Pebble
- FreeMarker
- Trimou

Middle-of-the-road (2x slower than best)
- Thymeleaf

Bottom-of-the-pack (5x slower than best)
- Groovy

**Prod Mode**
Best performers (all fairly close, Pebble & Groovy are top)
- Pebble
- Groovy
- FreeMarker
- Trimou

Middle-of-the-road (2x slower than best)
- Thymeleaf

**The Engines**
[Pebble](http://www.mitchellbosecke.com/pebble/home). String parsing engine Mustache+FreeMarker+AngularJS-like mashup. Nice clean syntax. Young (immature?) engine has at least one important open bug related to if/then/else processing.  I really like this engine- it's the best of the features/capabilities of Mustache & FreeMarker.

[Trimou](http://trimou.org). String parsing engine. Mustache but with server-side extensibility - which you don't get _by design_ from [Mustache.java](https://github.com/spullara/mustache.java).  Developed by a RedHat engineer. Works very well.  Mustache is more readable and understandable than FreeMarker. A solid choice.

[Groovy MarkupTemplateEngine](http://groovy-lang.org/docs/groovy-2.3.7/html/documentation/markup-template-engine.html). Code compiling engine. Surprisingly performant in Prod mode. There are some trade-offs with white-space handling when your templates are code.  It's not a problem, but it may require extra tweaking to get generated pages Just Right.

[Thymeleaf](http://www.thymeleaf.org). DOM parsing engine. All templates must be well-formed XHTML.  All integrations are node attributes.  Disappointing performance, fairly verbose templates, inheritance is a 3rd-party integration.  This would be my last choice and probably not even on my list.

I was also considering trying out [Chunk](http://www.x5software.com/chunk), but I'm not sure I will bother since I think it would be hard to beat Trimou, Pebble, or Groovy.

I'll try to get my modules on Github in the next week or two.

gitblit:
Have you had any spare cycles to move metrics forward?

raphaelbauer:
Yes! I took a bit longer than expected, but I am working on it right now :)

gitblit:
Pushed your changes plus simplifications discussed in the other PR.

gitblit:
I have pending updates for one or two of my other proposals but I don't want to overwhelm your free time so I will hold them back until we get Metrics merged. Naturally, I'd like to put this PR to bed sooner than later. :smiley: 

I've taken a look at Graphite and heard it mentioned in a talk at a conference last week.  Sounds useful & interesting. How are you hooking them together?

gitblit:
[JMXTrans](https://github.com/jmxtrans/embedded-jmxtrans) perhaps?

gitblit:
Ok, that was a dumb question.

I've optionally exposed JVM and Logback metrics.  I also added the Graphite and Ganglia reporters.  They are untested as I don't have a working install of either.

I have been playing with Librato (cloud service) which I can say integrates quickly and easily.  Looks like DataDog will also integrate about as easily - I'm sure there are many others too.

gitblit:
I  have also come across what may be a more current alternative to Graphite, InfluxDB+Grafana.

http://influxdb.com
http://grafana.org

Don't know anything about it - but it looks pretty.  ;)

gitblit:
Alright, I'm done hacking again.  Time to review my changes?

gitblit:
@raphaelbauer Do you know why buildhive is running 2x for every pushed change?  It's strange.  BTW the _unstable_ is a false-negative.  The previous triggered build of the same committed code was a success.

I've been playing with InfluxDB+Grafana.  Amazing.  Not sure if you've had the chance to use that yet.  I pushed an integration to this PR last night for it after I said I was done hacking.

raphaelbauer:
Having a look right now :) 
Graphite is great (The classic talk is: https://www.youtube.com/watch?v=czes-oa0yik). Grafana is a nice frontend for Graphite - it's pretty, but I don't know it very well...

raphaelbauer:
ah - and "no" - I have no idea why buildhive "double checks" the commits...

gitblit:
I've copied the small sources for the influxdb reporter internally (not committed yet) to implement a feature the Librato reporter has - do not report idle stats.  The current reporter dependency will report timers & meters on each pass regardless of whether the metric has changed. Not ideal - although it lets you generate pretty decay graphs.  This change should be upstream - I'll open a PR to the influxdb reporter project after I am satisfied with it.

ninja-core/src/site/markdown/documentation/basic_concepts/metrics.md
```diff
@@ -0,0 +1,239 @@
...
+
+#### Logback Metrics
+
+You may optionally enable reporting of Logback log-level counts by setting *metrics.logback.enabled=true* in your `application.conf` file.
+
+    metrics.logback.enabled = true
+
+### Reporting Metrics via JMX for VisualVM/JConsole
+
+By default, Ninja Metrics will expose your metrics over JMX. You may disable this behavior by setting *metrics.jmx.enabled=false* in your `application.conf` file.
```
raphaelbauer: Is that really the default behavior? Imho should be false by default - I think of security and stability issues - or am I overly cautious? (you know JMX better than me apparently...)

raphaelbauer: forget about the comment - I thought all metrics are included within one module - but you created several modules - so when I activate ninja-metrics-graphite - the jmx stuff is not active - right? Therefore it's absolutely okay... :)

gitblit: I don't actually think it enables JMX TCP serving so this is likely poorly worded - and poorly named by the Metrics project.

What I believe it does is exposes the metric registry through MBeans which _can_ be reported via JMX.  MBeans are also visible from VisualVM/JConsole with a direct process connection - which is how I view them.  I don't believe JmxReporter explicitly starts tcp JMX server - I thought that was a VM level function.

We can certainly disable that flag by default and rename it as needed.  Maybe `metrics.mbeans.enabled` ?

ninja-metrics/src/main/java/ninja/metrics/InstrumentedCache.java
```diff
@@ -0,0 +1,146 @@
...
+ * Instruments the configured Ninja Cache instance for Metrics collection.
+ *
+ * @author James Moger
+ *
+ */
+public class InstrumentedCache implements Cache {
+
+    private final Cache underlyingCache;
+
+    private MetricsService metricsService;
```
raphaelbauer, line 36: we could make all variables final... (of course init() has to be done in the constructor). Tiny thing... Final is always nice...

ninja-metrics/src/main/java/ninja/metrics/MetricsServiceImpl.java
```diff
@@ -0,0 +1,189 @@
...
+            root.addAppender(appender);
+
+            log.debug("Registered Logback-Metrics integration");
+
+        }
+
+        /*
+         * JMX
+         */
+        if (ninjaProps.getBooleanWithDefault("metrics.jmx.enabled", true)) {
```
raphaelbauer: I'd set that to false by default...

ninja-metrics/src/test/java/ninja/metrics/InstrumentedCacheTest.java
```diff
@@ -0,0 +1,49 @@
...
+    @Mock
+    Cache Cache;
+    
+    @Mock
+    
+    
+    @InjectMocks
+    InstrumentedCache instrumentedCache;
+
+    public void teatAddKeyValueExpiration() {
```
raphaelbauer, line 44: that's a great test! WHO HAS WRITTEN THAT? ...
ops... me... ;)

You can remove my stupid stub if you like...

ninja-metrics/src/test/java/ninja/metrics/MetricsServiceImplTest.java
```diff
@@ -0,0 +1,44 @@
...
+    @Mock
+    MetricRegistry metricRegistry;
+
+    @Before
+    public void before() {
+        metricsServiceImpl = new MetricsServiceImpl(metricRegistry, ninjaProperties);
+    }
+    
+    @Test
+    public void testSomeMethod() {
```
raphaelbauer, line 41: also a test that could go away...

raphaelbauer:
Hey @gitblit 

just had a look at the code... It' really great. Especially as you created individual modules for each reporter... I am wondering if the testcoverage is large enough (there are obviously no integration tests for - say - ganglia). But on the other hand why not it's up to you to decide.

I'd like to put the JMX reporting to false by default. Otherwise I'd merge it (and release Ninja 3.4 asap).

Great work!

Raphael

gitblit:
Things I did:
- changed JMX setting name to "metrics.mbeans.enabled", set disabled by default, and updated docs
- added settings per-reporter to control reporting frequency/period
- added a changelog entry
- rebased all changes on develop

Things I did not do:
- remove your test stubs

I believe, other than the lack of tests, this is complete and ready.

gitblit:
I just submitted davidB/metrics-influxdb#10 to skip reporting idle timers, meters, and histograms to InfluxDB.  If that is accepted, merged, and a new artifact is released in a timely fashion, that will definitely be an update I will want to submit to this PR or as a new PR.

As for a next framework release, I'd stay on your existing cadence which puts us at end-of-the-month, right?  We may be able to squeeze in another PR or two.

raphaelbauer:
great :) merged!

raphaelbauer:
needless to say that your metrics stuff really rocks!

gitblit:
_Our_ metrics stuff.  :) Thanks for the help and for merging.

PR #197, opened 2014-07-22
Bumped version in archetype simple

svenkubiak:
Every time I create a new project I do the same thing in the simple archetype pom.xml. So, I thought why not change the archetype itself. 

I bumped most of the versions and removed some blank lines.

### DIFF
--- a/ninja-servlet-archetype-simple/src/main/resources/archetype-resources/pom.xml
+++ b/ninja-servlet-archetype-simple/src/main/resources/archetype-resources/pom.xml
@@ -1,172 +1,200 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Copyright (C) 2013 the original author or authors. Licensed under the 
-	Apache License, Version 2.0 (the "License"); you may not use this file except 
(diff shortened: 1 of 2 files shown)


raphaelbauer:
Hi Sven,

nice :) 

There are two things that come to my mind:
1) Your editor changed from tabs to spaces (or vice versa). So the diff is huge. Could you change that back to the original convention? That way we can clearly see what is going on :)
2) TinyMCE and bootstrap are imho directly referenced in the application ftl.html files (src="/assets/bootstrap/3.0.0...") or so... I bet the application does not properly style now.. Can you make sure that the editor and boostrap gets loaded properly (you can also write a testcase against it...)

Thanks :)

Best,

Raphael

svenkubiak:
Sorry about the formatting. Updated my IDE this week and forgot to change the damn spacing. I have updated the templates now as well. There was one odd thing though. JQuery 1.9.0 is reference in the template, but it isn't in the pom.xml and as far as I can see it didn't come with TinyMCE or Bootstrap. So I added the latest version of JQuery.

raphaelbauer:
perfect... many thanks for updating the archetype :+1:

PR #198, opened 2014-07-22
Throw IllegalStateException when a registered controller method does not exist

gitblit:
It is possible to register a controller method that does not exist.  This is logged but it will trigger a non-informative NPE when building the route.  Instead we should throw an IllegalStateException with a more informative message.


raphaelbauer:
Very useful! +test. Great!

PR #199, opened 2014-07-22
Add HTTP method-based reverse routing options with best-fit routes

gitblit:


raphaelbauer:
Hey @gitblit - really great addition :) You also included clean and readable documentation!

Do you think it's possible to add more tests? For instance RouterImpl.getReverse\* (only GET tested afiacs) and to the reverse routing TemplateEngineFreemarker*Route template helper?

Should not sound rude or anything. You really do a great job and even added documentation to your feature - it's just that we have to take care that Ninja stays a really stable platform - and my feeling is that we should add some more tests to make sure nobody breaks your feature in the future :)

Let me know if you got concenrs of if I can help :)

Best,

Raphael

gitblit:
I haven't forgotten about this.  Will get back to it next week.

gitblit:
I'm not thrilled with stock reverse routing, but I am also not thrilled with this PR.  I'm taking this back to formula.

PR #201, opened 2014-07-23
Log the resolved routes on Ninja startup

gitblit:


gitblit:
Grrrr.  Eclipse doesn't like the redefine-the-package capability used in ninja-servlet.  _sigh_.  Will fix PR shortly.

gitblit:
This _seems_ like a bug to me in the unit test structure of ninja-servlet.  I actually don't understand how they pass at all on CloudBees.  There is an import/package name problem which I have assumed is related to the package redefinition feature.  If I fix it locally, I can run all Maven tests nicely.  If I leave it stock, it fails.  It is confusing me how CloudBees can turn up successful builds of the same code.

Specifically, **controller**.DummyControllerForTesting needs to be **controllers**.DummyControllerForTesting

gitblit:
I haven't forgotten about this.  Will get back to it next week.  Right now routes without controllers and methods are ignored - so that is actually undesirable.  Will fix.

gitblit:
@raphaelbauer Maybe you can review this weekend.

raphaelbauer:
sorry... thanks for reminding me! :)

PR #205, opened 2014-08-01
Add Http Basic Authentication filter and proper 401 unauthorized support

gitblit:
Ninja would benefit from optional http basic authentication, and here it is.  :)

**Step 1:** You need to bind a credentials validator:

``` java
public class Module extends AbstractModule {

    @Override
    protected void configure() {

        // bind a UsernamePasswordValidator
        bind(UsernamePasswordValidator.class).toInstance(new UsernamePasswordValidator() {

            @Override
            public boolean validateCredentials(String username, String password) {
                return "user".equals(username) && "password".equals(password);
            }
        });
    }
}
```

**Step 2:** Annotate protected resources as desired `@FilterWith(BasicAuthFilter.class)`

``` java
@FilterWith(BasicAuthFilter.class)
public class MyProtectedApiController {
}
```

**Note**
I found a discrepancy in _forbidden_ message reporting which I have addressed.  Hopefully that was correct.

### DIFF
--- a/ninja-core/src/main/java/ninja/BasicAuthFilter.java
+++ b/ninja-core/src/main/java/ninja/BasicAuthFilter.java
@@ -1,0 +1,94 @@
+/**
+ * Copyright (C) 2012-2014 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ninja;
+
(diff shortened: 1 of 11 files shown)


raphaelbauer:
Looks very good!

to pick up your point regarding "I found a discrepancy in forbidden message reporting which I have addressed. Hopefully that was correct.".

Do you think it makes more sense having only one method to get a result for an error with the following signature (just a sketch):

Result getErrorResult(Context context, Optional<Integer> httpError, Optional<Exception> exception) {
 ... if(httpErrror == 404) {
       ....
      } else if ...
}

would be a lot more flexible than having thousands of getInternalServerErrror... getNotFoundResult get... at conf.Ninja.

Do you have any thoughts about that?
(That's not a critique of you PR - it's more: "thinking about how we can improve the Ninja API").

Cheers,

Raphael

gitblit:
Yes, that would scale better and be a usable improvement.

I also think the default messages are worded strangely" Oops. That''s an internal server error and all we know."  In fact, pasting it here makes me re-read and understand the intent. "An internal sever error occurred and that is all we know."  Personally, I would reword these to something like "Oops.  You've triggered an internal server error." or "Oops. You've sent a bad (malformed?) request." or "Oops. The requested route is forbidden".

raphaelbauer:
I checked the code. Very well done! Thanks!

PR #206, opened 2014-08-03
RFC: Use @Path and @GET/@PUT/@POST for annotation router

gitblit:
This is an implementation based on our discussion.  It implements JAX-RS-style (don't cry) routing support for both the Controller and Controller Methods.  It removes the former `@Route` annotation.

This supports mode-limited routes.  Multiple modes can be specified for a route via the `@Dev`, `@Prod`, and `@Test` annotations.

This PR does NOT support route ordering - which I have decided may still be important for things like my Swagger module.  I still am thinking that RegEx routes do not belong here; those should be non-annotated only.

I also tweaked the  `AnnotationRouter` to implement the `ApplicationRoutes` interface - this is the same design seen in the `CasinoRouter`.

I think this moves the `AnnotationRouter` forward, but Darren should review.


mattjonesorg:
This looks like a great addition to the framework!

raphaelbauer:
Looks really nice!

gitblit:
I actually want to adjust this PR, if it's ok.  There is a change I'd like to make here:

I want to annotate multiple paths for my routes (I had this feature in my former annotation router PR)

``` java
@Path({"{repo}/commits", "{repo}/commits/{ref}"})
```

This creates a minor issue for annotating controllers with @Path since you could now put multiple entries there.  My instinct is multiple paths on controllers is undesirable so perhaps controllers should be annotated `@Controller("path")` instead.  Alternatively, we could allow multiple controller paths.

What do you think?

raphaelbauer:
Actually I have no idea. I'd ask Darren who got stakes in the feature :) But just two quick shots:
1. You can already handle multiple routes via regex in your route definitions
2. What's so bad about having multiple paths on controller level? I mean - yes you got more permutations for all other controllers - but apart from that?

But again - Darren might have much better ideas than I have. Reason: I am a superhappy with my Routes.java already :D

raphaelbauer:
And another idea: jax-rs 2.x doesn't implement this. Why? This might be completely bogus - but maybe they had very good reasons for not allowing multiple paths for one controller. But maybe not and it's very cool implementing that... Just guessing...

gitblit:
@DazHudson, please review this PR.

ninja-annotation-router-beta/src/main/java/router/AnnotationRouter.java
```diff
@@ -34,77 +34,160 @@
...
-        }
-        
-        return packagesToScanForRoutes;
-    
-    }
-
-    
+public class AnnotationRouter implements ApplicationRoutes {
+
+	final static String ANNOTATION_ROUTER_PACKAGES = "annotation_router.packages";
```
raphaelbauer, line 45: Seems to use tabs... We should stuck to spaces... tiny thing, but important at the end :)

ninja-annotation-router-beta/src/main/java/router/AnnotationRouter.java
```diff
@@ -34,77 +34,160 @@
...
+					Path controllerPath = declaringClass.getAnnotation(Path.class);
+					if (controllerPath != null) {
+						controllers.put(declaringClass, controllerPath.value());
+					} else {
+						controllers.put(declaringClass, new String [] { "" });
+					}
+				}
+
+				final String [] controllerPaths = controllers.get(declaringClass);
+				for (String controllerPath : controllerPaths) {
```
raphaelbauer, line 118: hmm... that does not (yet) work for controllers that extend other controllers... Is that a viable scenario?

@Path("/api)
BaseController() {}

@Path("person)
PersonApi extends BaseController() {

 @Path(person)
  Result getPerson() {...}
}

ninja-annotation-router-beta/src/main/java/router/AnnotationRouter.java
```diff
@@ -34,77 +34,160 @@
...
+					for (String pathSpec : path.value()) {
+						final String httpMethod = getHttpMethod(method);
+						final String fullPath = controllerPath + pathSpec;
+						final String methodName = method.getName();
+
+						if (httpMethod == null) {
+							throw new IllegalArgumentException(String.format(
```
raphaelbauer, line 125: Would be GET be a sensible default? We could print an info message about it...

ninja-annotation-router-beta/src/main/java/router/AnnotationRouter.java
raphaelbauer, line 118: If we make that recursively we could just do away with the set an map... Just walk up the dependency tree recursively and create the route at the end...

ninja-annotation-router-beta/src/main/java/router/AnnotationRouter.java
```diff
@@ -34,77 +34,160 @@
...
+	private boolean allowMethod(Method method) {
+
+		int mode = MODE_ALL;
+		if (method.isAnnotationPresent(Dev.class)) {
+			mode |= MODE_DEV;
+		}
+		if (method.isAnnotationPresent(Test.class)) {
+			mode |= MODE_TEST;
+		}
+		if (method.isAnnotationPresent(Prod.class)) {
```
raphaelbauer, line 175: I somehow don't like binary comparisons on ints. If you like them we can leave them like that. On the other hand I'd really favor having 
1) private boolean allowMethod(Method method, NinjaProperies ninjaProperties) {...}
2) Doing away with the binary ints and using plain old comparisons via bools...

I can change that, too. It's really just a matter of taste...

raphaelbauer:
Hey James, 

I added some remarks. I hope they don't sound rude or so. They just reflect my brain reading code :) The bottom line is that you are writing really really good code, tested and with documentation. So don't regard my comments as sad critique... They are just meant as hopefully helpful remarks that help us to run Ninja's codebase the next decades :) I can also incorporate my changes on your fork if you don't have time :) Keep up the great work... It really looks great! The remarks are really minor....!

ninja-annotation-router-beta/src/main/java/router/AnnotationRouter.java
gitblit, line 45: This module doesn't use the Ninja Eclipse settings AFAICT.  I'll reformat.

ninja-annotation-router-beta/src/main/java/router/AnnotationRouter.java
gitblit, line 118: I hadn't considered parent controllers.  Will give that some thought.

ninja-annotation-router-beta/src/main/java/router/AnnotationRouter.java
DazHudson, line 175: You also have to remember that users can create their own modes, upon which this would fail.

I was going to change my annotation to have an exclude field, because when you think about it, this would normally be used to exclude from production.

gitblit, line 175: Have you tried to create your own mode?  I'm not sure it works like you expect.  The mode is tied to an enum.  If you create an mode named "james" you end up running in "prod" anyway.  At least that was my finding.

raphaelbauer, line 175: I think creating your own mode is not really supported. Prod mode is eg an important signal to some plugins. Freemaker templating eg uses really aggressive caching if ran in prod mode. So you should use prod mode if running in production... On the other hand "dev" is a clear signal that Freemarker should not cache any templates so that changes show up immediately.

A cousin of "modes" is using different application.conf files. You typically want to run different stages (dev, qa, live or so) and configure different databases or services depending on the stage your Ninja app is running on. That's were different application.conf files come in handy. Simply start the servers via -Dninja.external.configuration=application_stage_qa.conf with the configuration of your choice....

I hope that makes the notion of modes a bit clearer. What does that now mean for the routes and its annotations? No idea :D 

If you got any ideas how to improve the doc let me know... It should be there: http://www.ninjaframework.org/documentation/configuration_and_modes.html

gitblit:
Since @DazHudson and I seem to have irreconcilable differences in our approaches I am thinking his proposal for separate annotated route builders has merit. 

The question would therefore be - where is the appropriate place for these modules?  @raphaelbauer has expressed happiness with the current router.  Is there room for one or both within the project or should they be maintained externally?

ninja-annotation-router-beta/src/main/java/router/AnnotationRouter.java
DazHudson, line 175: Hey guys,

I was going off the documentation about modes, but I can see being able to have extra nodes being handy like staging, jenkins, etc ...

I think that we either put effort into it to making modes work, or change the documentation.

D.

On 4 Sep 2014, at 22:07, James Moger notifications@github.com wrote:

> In ninja-annotation-router-beta/src/main/java/router/AnnotationRouter.java:
> 
> > -    \* Determines if this method may be registered as a route for the current
> > -    \* runtime mode. A method may be annotated to run in multiple modes.
> > -    *
> > -    \* @param method
> > -    \* @return true if the method can be registered as a route
> > -    */
> > -   private boolean allowMethod(Method method) {
> >   +
> > -       int mode = MODE_ALL;
> > -       if (method.isAnnotationPresent(Dev.class)) {
> > -           mode |= MODE_DEV;
> > -       }
> > -       if (method.isAnnotationPresent(Test.class)) {
> > -           mode |= MODE_TEST;
> > -       }
> > -       if (method.isAnnotationPresent(Prod.class)) {
> >   Have you tried to create your own mode? I'm not sure it works like you expect. The mode is tied to an enum. If you create an mode named "james" you end up running in "prod" anyway. At least that was my finding.
> 
> —
> Reply to this email directly or view it on GitHub.

Darren Hudson
Mobile: 07866 004000  Office: 01628 623367
www.docfacto.com

DazHudson:
Might be worth thinking about changing routes so they can be changed at runtime.  This way, we could just leave routes.java empty, run own own modules upon startup and then call buildRoutes...

gitblit:
Ok.  That gets back to replacing Router - or having multiple Routers - instead of injecting builders into the current router.  My initial PR #195 took this approach and used something similar to your `@Route` annotation.  Obviously that requires core changes to make happen.

On custom modes.  `jenkins` (or `ci`) is an interesting mode proposal.  @raphaelbauer when mvn test is executed does that use the `test` mode?  I haven't played with modes too much.

DazHudson:
You have to set the mode, I.E -dninja.mode=test ..
But there is a difference between running local test and CI test, so a different mode would be good.

raphaelbauer:
test, prod and dev is too simplistic for even the simplest CI stages. I usually have different databases, services etc pp in my different stages. dev uses my mysql dev and several other internal dev services, qa uses the qa mysql and several internal qa services and so on.

The solution is to use Guice providers that configure themselves via the application.conf. If you deploy Ninja via chef make sure you select the "correct" application.conf and tell Ninja about it via "-Dninja.external.conf=qa.conf". This selectively overwrites keys and makes it supersimple to configure and maintain different more-or-less complex configurations across your stages.

test, dev and prod is cool locally. But for everything else it's just too simplistic.

I hope that makes it roughly clear what the idea behind "modes" is - and what "modes" should not be. eg. it would not make sense having a custom mode - because prod mode is a signal to use aggressive caching for eg the templating engine. 

But anyway - its really not a problem - just use custom application.conf files for your stages. Running tests uses the mode "test" - but you can always override the setting by putting another application.conf on the classpath.

.. and back to the problem: Why don't you activate or deactivate routes based on the presence of a certain application.conf property?

raphaelbauer:
Regarding routers-in-ninja-core: I think it would be cool having annotation based routing inside the core. I'd be willing to maintain it in the long run. But - I already said that - I am personally more than happy with Routes.java and its possibilities. I couldn't be happier :)

That means it's a bit up to you. Ideally we can merge the ideas and maintain one annotation based router. That way we all gain from fixes and improvements :)

As fallback solution you can always maintain your ideas as separate modules and use them inside your own Ninja apps.

gitblit:
I've opened #242 for a new `ninja-jaxrs-routes` module which satisfies all requirements discussed in this PR.

PR #227, opened 2014-09-12
Add ServletContext to ContextImpl to improve 3rd-party integration

gitblit:
Some servlet-based integrations require access to the ServletContext. While
this could probably be faked with a custom wrapper around Ninja's Context
it is easier to just provide the actual context to ContextImpl for those integrations.

One example integration which would benefit from exposing the ServletContext
is the Thymeleaf template engine[1]. It features built-in servlet integration which
can not be used without a ServletContext.
- [1]: http://thymeleaf.org

### DIFF
--- a/ninja-core/src/site/markdown/developer/changelog.md
+++ b/ninja-core/src/site/markdown/developer/changelog.md
@@ -1,6 +1,7 @@
 Version 3.x.x
 =============
 
+* 2014-09-12 Add ServletContext to ContextImpl to improve 3rd-party integration
 * 2014-08-29 Added nicer error screens (ra)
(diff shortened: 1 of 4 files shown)


raphaelbauer:
nice! Regarding templating - imho jamon (http://www.jamon.org/) has a really interesting concept...

gitblit:
I've got three new (& complete) template engine modules in-progress which I will share after I create a test project demonstrating each of them:
- Groovy Template Engine, 2.3+
- Trimou (server-side friendly Mustache, like ninja-mustache but with extensability)
- Thymeleaf

FreeMarker works well and is surprisingly snappy but I am not crazy about it's syntax so I've been looking for alternatives.  I'll take a look at Jamon.  My bias, though, leads me to consider that a project that remains on SourceForge is more or less dead.

PR #228, opened 2014-09-15
Log the registered template and body parser engines

gitblit:
It would be helpful to log the registered template engines with their content types and the bodyparser engines with their accept types.

### DIFF
--- a/ninja-core/src/main/java/ninja/bodyparser/BodyParserEngineManager.java
+++ b/ninja-core/src/main/java/ninja/bodyparser/BodyParserEngineManager.java
@@ -16,10 +16,14 @@
 
 package ninja.bodyparser;
 
+import java.util.Set;
+
 import com.google.inject.ImplementedBy;
 
 @ImplementedBy(BodyParserEngineManagerImpl.class)
 public interface BodyParserEngineManager {
+
+    Set<String> getContentTypes();
 
     BodyParserEngine getBodyParserEngineForContentType(String contentType);
 
--- a/ninja-core/src/main/java/ninja/bodyparser/BodyParserEngineManager.java
+++ b/ninja-core/src/main/java/ninja/bodyparser/BodyParserEngineManager.java
@@ -23,8 +23,20 @@ import com.google.inject.ImplementedBy;
 @ImplementedBy(BodyParserEngineManagerImpl.class)
 public interface BodyParserEngineManager {
 
+    /**
+     * Returns a set of the registered body parser engine content types.
+     *
+     * @return the registered content types
+     */
     Set<String> getContentTypes();
 
+    /**
+     * Find the body parser engine for the given content type
+     *
+     * @param contentType
+     *            The content type
+     * @return The body parser engine, if found
+     */
     BodyParserEngine getBodyParserEngineForContentType(String contentType);
 
 }
--- a/ninja-core/src/main/java/ninja/bodyparser/BodyParserEngineManagerImpl.java
+++ b/ninja-core/src/main/java/ninja/bodyparser/BodyParserEngineManagerImpl.java
@@ -18,9 +18,9 @@ package ninja.bodyparser;
 
 import java.util.Map;
 import java.util.Set;
-import java.util.TreeSet;
 
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
 import com.google.inject.Binding;
 import com.google.inject.Inject;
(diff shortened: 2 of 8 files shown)


raphaelbauer:
Thanks gitblit :) Why don't we just log everything in the constructors of BodyParserEngineManagerImpl and TemplateEngineManagerImpl? That way we don't have to extend the public interface of both classes and we don't have to change NinjaBootstrap... no critique... just a question - logging that information is certainly very useful :)

gitblit:
I like rhythm and pattern.  Moving the logging to the injected constructors could make the pattern of startup logging non-deterministic as you are subject to Guice's internal ordering behavior.  I'll give it a go and report back.

raphaelbauer:
That's certainly a good reason doing logging that way... Let me know what you find...

ninja-core/src/main/java/ninja/template/TemplateEngineManagerImpl.java
```diff
@@ -71,6 +73,11 @@ public TemplateEngineManagerImpl(Provider<TemplateEngineFreemarker> templateEngi
     }
 
     @Override
+    public Set<String> getContentTypes() {
+        return new TreeSet<String>(contentTypeToTemplateEngineMap.keySet());
```
raphaelbauer: Using something like ImmutableSet.copyOf(contentTypeToTemplateEngineMap.keySet()); would make sense imho...

ninja-core/src/main/java/ninja/template/TemplateEngineManager.java
```diff
@@ -27,9 +28,11 @@
 @ImplementedBy(TemplateEngineManagerImpl.class)
 public interface TemplateEngineManager {
 
+    Set<String> getContentTypes();
```
raphaelbauer, line 36: javadoc would be great...

ninja-core/src/main/java/ninja/bodyparser/BodyParserEngineManager.java
```diff
@@ -16,11 +16,15 @@
...
 package ninja.bodyparser;
 
+import java.util.Set;
+
 import com.google.inject.ImplementedBy;
 
 @ImplementedBy(BodyParserEngineManagerImpl.class)
 public interface BodyParserEngineManager {
 
+    Set<String> getContentTypes();
```
raphaelbauer, line 31: javadoc would be great :)

ninja-core/src/main/java/ninja/bodyparser/BodyParserEngineManagerImpl.java
```diff
@@ -71,6 +73,11 @@ public BodyParserEngineManagerImpl(Provider<BodyParserEnginePost> bodyParserEngi
     }
 
     @Override
+    public Set<String> getContentTypes() {
+        return new TreeSet<String>(contentTypeToBodyParserMap.keySet());
```
raphaelbauer: Using something like ImmutableSet.copyOf(contentTypeToTemplateEngineMap.keySet()); would make sense imho...

ninja-core/src/main/java/ninja/template/TemplateEngineManagerImpl.java
gitblit: I'll change it to a sorted immutable list.  I wanted the types deterministically ordered too.  :)

raphaelbauer:
I rather say that one time too often - but whenever I add something to pull requests (comments, questions etc pp) it's NOT meant to be condescending. I only try to keep the whole codebase in good shape and keep certain quality criteria up :) You certainly do a very very good job! And all I say via comments etc may be wrong...

raphaelbauer:
I think it would also be great to have simple unit tests for both getContentTypes implementations.

gitblit:
Where to log.  I'm on the fence here. Which ever you prefer is fine with me.  You have precedent for startup logging like MessagesImpl which complains that I don't have a de translation.  Tell me your preference and I'll make it so.  If it is to move, then perhaps Route logging should be in the RouterImpl to be consistent.

raphaelbauer:
I'd certainly prefer logging inside the constructors. If that turns out to be a really bad idea we can change that to extending the interfaces and logging at NinjaBootstrap...

gitblit:
You want the Logger injected (Ninja) or the Logger instantiated (*Impl) ?

raphaelbauer:
instantiated via LoggerFactory.getLogger... There is not yet a way how tell the injected logger in which class it logs...

gitblit:
I'll move the Routing table to be consistent, but I'll leave *.getContentTypes() - with all review comments satisfied - since there might be other uses for that info.

gitblit:
Changes pushed.

raphaelbauer:
very nice. Will merge locally... Thanks!

PR #229, opened 2014-09-15
RFC: One way to modularize Ninja

gitblit:
This PR is an exercise in determining how `ninja-core` could be broken into a few more pieces.  This lowers the footprint of ninja-core and it's deps by at least 15MB primarily by eliminating hard dependencies on caching and persistence technologies.  It also allows devs to more discretely pick their technologies.

The first two commits (engine logging) have also been submitted separately as #228. Part of their motivation is for this PR, but they have value outside the scope of this PR too.

The third commit sets-up the JPA/persist module since the current design is to pass `NinjaProperties` to the `JpaModule` constructor.

The remainder of the commits are mostly the grunt work required to make ninja-core un-opinionated and to ensure that Ninja passes all unit tests between extractions.  Being un-opinionated means that things like `ninja-freemarker` and `ninja-jackson` **must** be specified as dependencies in the application pom because out-of-the-box ninja-core is incapable of generating content.

I'm not saying this breakdown is the best one, but it does highlight some tricky spots like `JpaModule`, `Migrations`, and `CacheProvider`.

It also begs the question: what is a Ninja Module?  The current answer (Guice Module + Providers) isn't very formal.   In a different PR I advocated an auto-register annotation for modules.  Maybe something like that could be upstreamed or perhaps a Interface definition?  Not sure.  In any event, I'm interested to hear your feedback.


gitblit:
Fixed the archetypes.

raphaelbauer:
Hey gitblit... Thanks! I will have a look at it asap...

raphaelbauer:
Will have a look at it next weekend. I got some time then :) Thanks!

gitblit:
No problem.  This was meant to ignite discussion on _how_ to move things out of `ninja-core`.  I'm not completely satisfied with this PR and I consider it throw-away - although there may be a few ideas worthy of extraction.  I think we need a formal definition of a _Ninja Module_ before we can effectively move bits out of `ninja-core`.

raphaelbauer:
Okay... My observations:
1) Very nice work! 
2) There is a trade-off (surpise!). It was a deliberate decision to include almost everything into ninja-core. Makes developing way easier than maintaining a lot of separate maven modules. It also makes it a lot more complex for users to chose the right packages for the job
3) The separate maven modules looks great - but if we do that we should provide one "opinionated" ninja maven module that resembles the current ninja-core
4) gradle. what about gradle? I really love Maven, but when it comes to complex multi project builds I always have the feeling that Maven begins to sück. Maybe we can use gradle. Darren propsed that. But on the other hand tooling support for maven modules is just awesome... And with gradle - I am not so sure...
5) Regarding a "NinjaModule". One inspiration can be the way Play 1 handled plugins: https://github.com/playframework/play1/blob/1.2.6-security/framework/src/play/PlayPlugin.java . We could create an interface NinjaModule with the relevant methods (onStart, onStop... etcpp). And then have a conf.Plugins where you can add them to your application. With a dsl like addNinjaPlugin(JpaPlugin.class).priority(1000); or so....

Just some random thoughts to get us started :) Maybe we should create a ninjaframework-dev mailing list... github is not perfect for discussion things imho...

raphaelbauer:
ah - any one more thing. You may have noticed that already. But I really dislike when stuff happens and I don't know why. Therefore I'd really like to have a conf.Plugins where plugins are listed explicitly instead of a @NinjaPlugin annotation that magically loads stuff (possibly even in inherited modules). But we can talk about it. Afaik Darren is in love with the @NinjaPlugin annotation approach :)

gitblit:
1. Opinionated POM/Module. Perfect.  `- compile org.ninjaframework:ninja-classic:3.3.2`
2. Gradle. I haven't used it.  I've always run away from Maven and wrote my own Maven-ish toolkit for Ant ([Moxie](http://moxie.gitblit.com)).  I've been using Maven+Eclipse for the last 6 months or so and it's ok as long as your projects are structured the Maven way. I don't hate it as much as I originally had. If Gradle works well too, that would be fine too.  I don't think it has to be part of discussion to extract bits from ninja-core into modules BUT it certainly can be.
3. NinjaPlugin interface & registration.  Excellent idea.  I don't need automatic discovery and I am fine with manual registration.  I do think that the plugin constructor should take NinjaProperties to allow _simple_ runtime configuration (e.g. JpaModule).

raphaelbauer:
Is there anything I can do regarding that PR?

gitblit:
That's a good question.  Are you interested in moving this or something like it forward?

raphaelbauer:
Actually I don't need it personally right now. So I won't be the driving force behind such a PR. It's up to you or anyone interested - what about @dazhudson?

DazHudson:
I Agree that we should have a Ninja Plugin module, and the constructor should be NinjaProperties.
As far as auto discovery goes, well, that could be a plugin right?
We moved to gradle because of the large nature of our Ninja Project, more than happy to share all of the gradle stuff that we have done.

gitblit:
@DazHudson What, if anything, would be the relationship between NinjaPlugin and Guice modules?

DazHudson:
I don't think that there is anything ..  That being said, a module may rely on injection.
When I wrote the JMX module, I did struggle with Guice and loading order.  I think that routing should be a module, right now it created at bootstrap, but I wanted to get my modules loaded before that.

gitblit:
I will push this ahead incrementally by creating a `ninja-classic` POM project and then opening a separate PR for each module extracted from `ninja-core`.

PR #231, opened 2014-09-18
Add explicit text/plain template engine

gitblit:
This change allows you to return something like this:

```
Results.text().render(object);
```

The object's `toString()` will be invoked by the `TemplateEngineText` class.

This is more natural than the currently supported syntax:

```
Results.text().renderRaw(object.toString());
```

and fits in better with the other stock return types:

```
Results.json().render(object);
Results.jsonp().render(object);
Results.html().render(object);
Results.xml().render(object);
```

With the current state of _develop_ the former syntax will throw an exception because of the unbound template engine for `text/plain`.

### DIFF
--- a/ninja-core/src/main/java/ninja/Result.java
+++ b/ninja-core/src/main/java/ninja/Result.java
@@ -328,29 +328,34 @@ public class Result {
     
 
     /**
-     * This method directly renders the String to the output. It
-     * completely bypasses any rendering engine.
-     * 
+     * This method directly renders the String to the output. It completely
+     * bypasses any rendering engine.
+     *
      * Thus you can render anything you want.
-     * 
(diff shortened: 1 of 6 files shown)


raphaelbauer:
thanks :) will have a look the next days!

raphaelbauer:
That's pretty nice and clever :) Can you add the feature to changelog.md? I'll merge asap.

Btw - you can also deprecate the renderRaw(...) method if you want.

gitblit:
- added changelog entry
- updated JavaDoc
- deprecated Result.renderRaw(String)
- add Result.text() method to complement existing Result.contentTypeMethods

raphaelbauer:
Awesome :+1:

PR #235, opened 2014-10-01
Add support for hot reloading of configuration files

gitblit:
There are two scenarios where hot-reloading of Ninja configuration files
would be very useful:
1. reloading localized messages files while in dev mode (i.e. realtime externalization)
2. reloading externally specified application configuration for all runtime modes

### DIFF
--- a/ninja-core/src/main/java/ninja/i18n/MessagesImpl.java
+++ b/ninja-core/src/main/java/ninja/i18n/MessagesImpl.java
@@ -28,6 +28,8 @@ import ninja.utils.SwissKnife;
 import org.apache.commons.configuration.CompositeConfiguration;
 import org.apache.commons.configuration.Configuration;
 import org.apache.commons.configuration.ConfigurationConverter;
+import org.apache.commons.configuration.PropertiesConfiguration;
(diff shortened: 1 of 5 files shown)


gitblit:
@raphaelbauer I'm not sure how to unit test these proposed changes.  I'm willing to rely on commonsconfiguration unit testing here :smile:.  If you have a better idea, I'll make it so.

raphaelbauer:
Hey gitblit. Very nice :)

The only thing that I find a bit suspicious is reloading the external configuration file automatically.

externalConfiguration.setReloadingStrategy(new FileChangedReloadingStrategy());

Is this something you use? Or is it just something you think could be useful...
Because in my production environments I cannot think of a good usecase for that - and knowing that my servers automatically pick up changes and reload stuff makes me a bit nervous...

gitblit:
Actually, I do use that feature in my non-Ninja apps.  It's _very_ handy to be able to adjust runtime settings without causing downtime.  If it makes you nervous, perhaps a cmd-line flag to enable it?

raphaelbauer:
Command flag or something equal sounds great (maybe a simple option inside application.conf?). Default should be not to reload it (for the nervous users ... ;) )...

raphaelbauer:
nice!

PR #238, opened 2014-10-07
Revise the boolean setting controlling external configuration hot-reloading

gitblit:
The addition of the setting within the external application.conf file to control
enabling/disabling runtime hot-reloading for #235 was hastily done.

In retrospect, it would be better to control this behavior by use of a System
property which complements the system property already used to specify the
external configuration location.

```
java -Dninja.external.reload=true -Dninja.external.configuration=app.conf
```

### DIFF
--- a/ninja-core/src/main/java/ninja/utils/NinjaPropertiesImpl.java
+++ b/ninja-core/src/main/java/ninja/utils/NinjaPropertiesImpl.java
@@ -145,7 +145,8 @@ public class NinjaPropertiesImpl implements NinjaProperties {
 
                 // allow the external configuration to be reloaded at
                 // runtime based on detected file changes
-                if (externalConfiguration.getBoolean(NinjaConstant.applicationHotReloadExternalConfig, false)) {
+                final boolean shouldReload = Boolean.getBoolean(NINJA_EXTERNAL_RELOAD);
+                if (shouldReload) {
                     externalConfiguration
                             .setReloadingStrategy(new FileChangedReloadingStrategy());
                 }
--- a/ninja-core/src/main/java/ninja/utils/NinjaConstant.java
+++ b/ninja-core/src/main/java/ninja/utils/NinjaConstant.java
@@ -111,9 +111,6 @@ public interface NinjaConstant {
      */
     final String serverName = "application.server.name";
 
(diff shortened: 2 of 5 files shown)


ninja-core/src/main/java/ninja/utils/NinjaPropertiesImpl.java
```diff
@@ -145,7 +146,9 @@ public NinjaPropertiesImpl(
 
                 // allow the external configuration to be reloaded at
                 // runtime based on detected file changes
-                if (externalConfiguration.getBoolean(NinjaConstant.applicationHotReloadExternalConfig, false)) {
+                final boolean shouldReload = BooleanUtils.toBoolean(System
```
raphaelbauer: I think we can use something like Boolean.getBoolean(NINJA_EXTERNAL_RELOAD) - a bit shorter and no external dep...

raphaelbauer:
Can you tell me why that is better? Just out of curiosity... I'll merge asap...

gitblit:
Great idea.  I amended the commit to use `Boolean.getBoolean()`.

_The Ninja-design reason._ I think it's weird to affect the behavior of loading the file from the file being loaded.

_My personal need reason._  A System property allows external/container-based control.

I am bootstrapping my prototype Ninja app with a more complex harness than ninja-standalone.  My original/old app runs an embedded Jetty, uses properties file configuration, supports hot-reloading, and uses Guice.  You can see the parallels to Ninja which makes your platform a very comfortable fit for me.

The System property approach allows me to enable the reloading behavior without needing to introduce a new setting into my existing app or make my current users add that setting/property.  This allows the Ninja-based app to behave _exactly_ like my users are accustomed to.  The setting approach is less seamless for me as it would require introducing that key-value pair to 20k+ installs.

raphaelbauer:
Very good to know. Thanks! What's the "20k+ installs" thing about? That sounds huge...

gitblit:
That's a very conservative estimate of total install base based on the download count for my [current release](https://bintray.com/gitblit/releases/gitblit/1.6.0/view/statistics).
http://gitblit.com

raphaelbauer:
Ah okay. Really great :) Initially I thought you are managing 20k+ installs yourself...

PR #239, opened 2014-10-08
Allowing custom date formats

pedro-stanaka:
Allow developer to specify date formats using application.conf file.


gitblit:
This is something that should be configurable but two things:
1. Does injecting into a static method seem strange?
2. Shouldn't the format be changeable with each parameter rather than global?

pedro-stanaka:
Now that you're talking that way, It really is strange @gitblit. And I'm making something that should be simple to test a lot harder, because I have to "mock" NinjaProperties. 

Do You have any ideas of what I could do? This parameter that you were talking I should put in the context? 

P.S. I though about some way to do this "automagically" using locale as well.

Thanks in advance!

jlannoy:
Using a custom date formats can now be done by implementing a custom ParamParser for the Date type.

PR #241, opened 2014-10-11
Removed '?' wildchar (if ends with) when reversing the route.

sojin:
Minor fix:
Curretly, getReverseRoute() includes the wild character '?' when reversing a route which ends with '?' (/?, for eg) and this PR fixes that trailing char.

Thanks
Sojin

### DIFF
--- a/ninja-core/src/main/java/ninja/RouterImpl.java
+++ b/ninja-core/src/main/java/ninja/RouterImpl.java
@@ -134,6 +134,10 @@ public class RouterImpl implements Router {
                     urlWithReplacedPlaceholders,
                     ninjaProperties);
 
+            if(finalUrl.endsWith("?")) {
+                finalUrl= finalUrl.substring(0, finalUrl.length() - 1);
+            }
+
             return finalUrl;
 
         }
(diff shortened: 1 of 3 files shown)


raphaelbauer:
Hey Sojin, thanks for the PR :) As far as I understood that's a "hotfix" for a very general problem "Freestyle" regex parts will stay inside the reversed url. 
Just out of curiosity - Why do you have "?" at the end of your url?
I will think about it - maybe I can contribute some more to fix that... :)

sojin:
Hi Raphael,

We would need to have url with '?' to match urls with and without traling slash.

For eg:
we need to have /usr/? to match both /usr and /usr/
So if you use reverse route on the method corresponds to '/usr/?', it will return /usr/?, which is not ideal.
Any better solution would be more than welcome.
Thanks
Sojin

raphaelbauer:
Hi sojin,

some more ideas from me. Just removing a trailing "?" from the route does not seem 100% right for me - because it's a very special case. And implementing that special case inside ninja-core - may be doing the wrong thing for a lot of people.

I think a "clean" solution is to use a named parameter in your route and setting that parameter to empty when reversing the route.

eg.
/user{slash: ?}
and then
router.getReverseRoute(Controller.class, "method", "slash", "");

Or - if you want - set the slash to "/" if you prefer a slash at the end...

That sounds like a clean solution to me...

What do you think?

Cheers,

Raphael

raphaelbauer:
Is it okay to close that issue?

sojin:
Yup.

PR #242, opened 2014-10-18
Add JAX-RS style route builder

gitblit:
This route builder features:
- registration for all standard http methods
- registration of routes for multiple runtime modes
- registration of a route based on the presence of Ninja a Property
- discrete control over registration order

This route builder is used in conjunction with the standard Ninja Router.

This PR is a replacement for #206.


tmarwen:
@gitblit Just brilliant!! This is what I was absolutely missing to get the JAX-RS style I've been always familiar with.

ninja-core/src/site/markdown/documentation/basic_concepts/routing.md
```diff
@@ -353,3 +353,132 @@ a @PathParam("id") would then be rootuser/domain as it is decoded by Ninja.
 
 In principle it is really simple. But it is even simpler to mess encoding / decoding up.
 The article from Lunatech mentioned earlier is awesome and explains everything.
+
+## JAX-RS-style Annotated Routes in Ninja (optional)
+
+*Ninja-jaxrs-routes* allows you to register your routes using annotations similar to JAX-RS.
```
raphaelbauer: Not sure if we can or should use the term "jaxrs" in terms of intelectual property. We are not following the jaxrs specs, and some people may not like it..

Calling it "AX-RS-style" is okay to point to similarities. But for the java packages I'd prefer something else. Same same  - but different. ninja.jaxyrs maybe? Or am I overly cautious?

ninja-jaxrs-routes/src/main/java/ninja/jaxrs/JaxrsRoutes.java
```diff
@@ -0,0 +1,317 @@
...
+        return paths;
+    }
+
+    /**
+     * Returns the set of packages to scan for annotated controller methods.
+     *
+     * @param ninjaProperties
+     * @return the set of packages to scan
+     */
+    private Set<URL> getPackagesToScanForRoutes(NinjaProperties ninjaProperties) {
```
raphaelbauer: why does the method take ninjaProperties?

ninja-jaxrs-routes/src/main/java/ninja/jaxrs/JaxrsRoutes.java
```diff
@@ -0,0 +1,317 @@
...
+
+        Set<URL> packagesToScanForRoutes = Sets.newHashSet();
+
+        packagesToScanForRoutes.addAll(ClasspathHelper
+                .forPackage(NinjaConstant.CONTROLLERS_DIR));
```
raphaelbauer: maybe that should be documented in the markdown docs... We are only scanning for controller methods in the controllers package...

raphaelbauer:
I second @tmarwen - brilliant! Maybe you can have a look at my comments... I'll merge asap... Very good work!

gitblit:
All review feedback is easy enough to adjust and I've done so with the exception of the name.  Will push after we resolve that.

I agree, that perhaps directly employing JAX-RS in the name could be an issue and could be misleading since this is not really a JAX-RS implementation.  The selected name trickles down from package to module to documentation so let's figure it out.
- ninja.jaxyrs
- ninja.jaxy
- ninja.jaxish
- ninja.jax
- ninja.jacks
- ninja.jazz
- ninja.yax (think eastern European 'J')
- ninja.ann (ann-otated)
- ninja.an ([Korean for tranqulity (i.e. REST)](https://en.wikipedia.org/wiki/Ahn_%28Korean_surname%29))

Others?

tmarwen:
In case community feedback is welcomed I would vote up for _ninja.jaxyrs_ or maybe _ninja.jaxish_ which for me seem the most meaningful and avoid confusions :smile:.

gitblit:
@tmarwen I kinda like _jaxish_ too. _jaxyrs_ doesn't roll off the tongue.

raphaelbauer:
I am happy with any choice... Just pick one that sounds good to you... :)

gitblit:
Package, module, and docs renamed to _ninja.jaxy_.

raphaelbauer:
jaxy sounds so cute - it could be a kitten's name!

raphaelbauer:
I'll merge asap. I am also thinking about releasing Ninja 4. The metrics stuff + jaxy is quite impressive James.

gitblit:
Thanks very much.  I'm happy to move Ninja forward.

I wonder if version 4 would be more appropriate for modularization?  I still plan to propose putting `ninja-core` on a diet.  :) The biggest wins would be modularizing `ninja.cache` and `ninja.jpa`.  Both of those have complex Guice providers and both depend on sizable transitives.

For example, `ninja-cache` directly links to ehcache which is a 6MB jar. Perhaps it should/could default to a light-weight Guava-based expiring cache and have ehcache, memcached, & redis (another thing I plan) as plugin/alternate implementations?  Just riffing here.

But my point is, if I can convince you that putting `ninja-core` on a diet is a Good Plan for the future, then that will definitely result in some effort in downstream upgrades from Ninja vA.B.C to Ninja vX.Y.Z.  A good signal of that could be the 4 designator.  But your choice, Mon Capitan.

gitblit:
> jaxy sounds so cute - it could be a kitten's name!

DeviantArt to the rescue.  I give you Jaxy the Ninja Kitten.

Cheezburger has many too.  :)

raphaelbauer:
Merged... Awesome - jaxy FTW!

PR #243, opened 2014-10-23
Update Jetty, Jackson, Guava, Guice, H2, ehcache, spymemcached, commons-...

gitblit:
Update our dependencies to the current releases. There actually is an ehcache 2.9.0 but I thought 2.8.5 might be better for now.


raphaelbauer:
nice!

PR #244, opened 2014-10-23
Sending mails from Template

svenkubiak:
I have added a method to MailImpl for sending a mail with a freemarker template. It simply takes the name of the template and a map for the values to parse and renders it. I am using this in all of my applications for sending emails and I think this would be a nice core feature. Let me know what you think.


gitblit:
This is an interesting proposal.  Personally, I'd like to not introduce more dependencies on FreeMarker into `ninja-core` - I'd actually like to see it as an optional dependency so the end-user can pick their engine without carrying around a FreeMarker dependency.

Could this be implemented in a more generic way - i.e. injecting the registered `text/html` TemplateEngine and rendering the mail templates using the existing methods?

raphaelbauer:
I think that's pretty cool. Also agree what @gitblit says... @svenkubiak Would that be an option for you?

svenkubiak:
I'll agree with @gitblit ... putting freemarker dependencies in the core is the wrong way. I'll think of something :-) I'll close this for now.

raphaelbauer:
If we can help let us know :) I think your idea is pretty cool!

Issue #188, opened 2014-06-10
Path param regex requires excessive backslash escaping


metacity:
Hi,

When using path parameters with regex matching (á la `/assets/{fileName: .*}`), Ninja seems to be requiring an extra "layer" of escaping for backslashes. So for example when setting up a route to match a numeric id:

``` java
// Does NOT work
router.GET().route("/customers/{id: \\d+}").with(Controller.class, "method");

// WORKS
router.GET().route("/customers/{id: \\\\d+}").with(Controller.class, "method");

// Also works
router.GET().route("/customers/{id: [0-9]+}").with(Controller.class, "method");
```

Regex without a named path parameter seems to work like intended. I'll take a closer look later, but after a quick glance through the routing related classes, could this be a result of first extracting the regex part of the route, converting it to a string and then compiling it to `Pattern`? If someone has an obvious solution, go ahead :)

Of course this could just be a documented feature.. Although matching a literal "\" would become quite tedious. Thoughts?

raphaelbauer:
Good catch. Actually I also have to look into the code to see what is happening... Any improvements all insights of course appreciated :)

bazi:
Resolved in [PR#247](https://github.com/ninjaframework/ninja/pull/247). Should it be closed?

@raphaelbauer  Who closes issues? Or the author @metacity ?

metacity:
Ah yes, thank you! :+1:

PR #247, opened 2014-10-25
Fixed issue #188

bazi:
A route containing path params with regex is converted to regular expression by using `Matcher.appendReplacement` which has a special treatment of backslashes (\) and dollar signs ($) in replacement string.
There is a `Matcher.quoteReplacement` method which is intended to produce a _literal_ replacement string for `Matcher.appendReplacement` method.

I think using the method above would be the best and minimalistic solution for this issue to be solved.

### DIFF
--- a/ninja-core/src/main/java/ninja/Route.java
+++ b/ninja-core/src/main/java/ninja/Route.java
@@ -185,7 +185,7 @@ public class Route {
             if (namedVariablePartOfRoute != null) {
                 // we convert that into a regex matcher group itself
                 namedVariablePartOfORouteReplacedWithRegex 
-                    = "(" + namedVariablePartOfRoute + ")";
+                    = "(" + Matcher.quoteReplacement(namedVariablePartOfRoute) + ")";
             } else {
                 // we convert that into the default namedVariablePartOfRoute regex group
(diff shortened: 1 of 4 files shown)


raphaelbauer:
Hey @bazi - that's one of those changes that really has a huge and positive impact :) Well done!
Just two notes:
1) Could you add a testcase that makes sure Ninja keeps on working as expected? There are a lot of changes happening at the same time in Ninja source - and if there is no testcase for your feature it is quite likely that your change will get destroyed :) There are already a lot of testcases regarding routing - simply add one to an already existing file :)
2) Is there any documentation needed (eg on our webpage)?

Thanks again bazi - really cool :)

Raphael

bazi:
1) done -- added test cases
2) in my opinion [routes doc page](https://github.com/ninjaframework/ninja/blob/develop/ninja-core/src/site/markdown/documentation/basic_concepts/routing.md) on our webpage has enough info regarding regex in routes and especially we have this `You can use any Java compliant regular expressions for matching`. When I first read docs, I found it clear and not overwhelming, it is _concise_. I think there is no need to add more right now.

raphaelbauer:
Looks like a perfect PR :) Thanks for your contribution!

PR #248, opened 2014-10-26
Added error message when no bodyparser for the content type was availabl...

eicnix:
When no Content-Type is sent the Content-Type "_/_" is used and because there is no default BodyParserEngine for this Content-Type a body request parse will always fail. 

The added error message should show the user that he needs to declare a proper Content-Type to automatically parse request bodies.

### DIFF
--- a/ninja-core/src/site/markdown/developer/changelog.md
+++ b/ninja-core/src/site/markdown/developer/changelog.md
@@ -1,7 +1,7 @@
 Version X
 =========
 
- * 2014-10-26 Added error log when no suited BodyParserEngine was found
+ * 2014-10-26 Added error log when no suited BodyParserEngine was found(templarthelast)
  * 2014-10-23 Updated libraries:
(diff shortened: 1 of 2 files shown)


raphaelbauer:
Very helpful!
Two things:
1) Would you mind signing the contributor's agreement? 
2) Could you add one line to changelog.md. I know - the change looks small, but it's nevertheless cool to know who has contributed what to each release :) Small-looking usability enhancements like yours are sometimes more important than huge badass features :)

gitblit:
It's great to log this... But perhaps the error should also be returned to the request origin as 'text/plain' ?

eicnix:
@raphaelbauer The agreement is signed and the changelog.md updated
@gitblit This there already an mechanism to abort the request processing and response and error?

raphaelbauer:
Merged the PR separately ({} is more perfomant than + in Strings).

Check out:
https://github.com/ninjaframework/ninja/blob/develop/ninja-servlet/src/main/java/ninja/servlet/ContextImpl.java

Thanks for your contribution! :)

PR #249, opened 2014-10-26
organized imports in all projects

svenkubiak:
After the the ninja-core project I did an "organize imports" on all ninja projects.


raphaelbauer:
very cool! thanks! 

Just as a note: It would be cool if you could add your changes to changelog.md. Even if your stuff looks "small" - it really is a big help in maintaining Ninja :) And that way nobody will overlook your contribution!

(I'll merge anyway - but next time consider adding even such small changes to changelog.md) :)

PR #251, opened 2014-10-28
Bumped version in archetypes

svenkubiak:
Updated the versions in both archetypes to the latest version.

### DIFF
--- a/ninja-servlet-jpa-blog-archetype/src/main/resources/archetype-resources/pom.xml
+++ b/ninja-servlet-jpa-blog-archetype/src/main/resources/archetype-resources/pom.xml
@@ -15,31 +15,25 @@
     
     <url>http://www.ninjaframework.org</url>
     
-    
     <properties>
         <ninja.version>@project.version@</ninja.version>
         <jetty.version>@jetty.version@</jetty.version>
     </properties>
-
-
     <build>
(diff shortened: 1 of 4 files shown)


ninja-servlet-jpa-blog-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -143,32 +131,23 @@
...
             <groupId>org.webjars</groupId>
             <artifactId>tinymce-jquery</artifactId>
-            <version>3.4.9</version>
+            <version>4.0.16</version>
         </dependency>
-
         <dependency>
             <groupId>org.webjars</groupId>
             <artifactId>bootstrap</artifactId>
-            <version>3.0.0</version>
```
raphaelbauer, line 162: not sure - but don't  we have to (unfortunately) also replace 3.0.0 and tinyMce's 4.0.16 at defaultLayout.ftl.html ? Did you run the integration tests or the archetype locally?

svenkubiak:
@raphaelbauer 
You are right. I updated the template and also added the jquery webjar which was missing in the jpa archetpye. Integration-Test was successful.

raphaelbauer:
perfect :) Thanks!

PR #264, opened 2014-11-28
Fix for: 263

raphaelbauer:
Added System.lineSeparator() to handle test file robots.txt in a platform independent way.


jlannoy:
The same modifications have apparently be made within an other PR.

Issue #270, opened 2014-12-14
Enhancement to allow sub routers


fzakaria:
I think if we changed the implementation of RouteBuilder `route` to append every invocation to the previous call you can build complex routes

`
RouteBuilder userRouter = router.route('/user');
userRouter.GET().route('/login')
userRouter.GET().route('/logout')

raphaelbauer:
Interesting. Not sure what the goal of this would be. Feel free to provide some code to make that happen (that is compatible with the way it works now). That will help to discuss the idea in more depth :) Thanks!

PR #272, opened 2014-12-14
Preview or router composition

fzakaria:
I have made some changes that I think might be interesting for allowing Router composition.
The distinction between Router & RouterImpl is a slight shame because it kind of ruins the niceness of thinking that every Router is a self-type (almost like a link list of Routers).

Anyways I was able to get something in without too much jitter of the code and I added some test examples.

Please take a look and let me know.

### DIFF
--- a/ninja-core/src/main/java/ninja/RouteBuilderImpl.java
+++ b/ninja-core/src/main/java/ninja/RouteBuilderImpl.java
@@ -35,7 +35,7 @@ class RouteBuilderImpl implements RouteBuilder {
             .getLogger(RouteBuilder.class);
 
     private String httpMethod;
-    private String uri;
+    private String uri = "";
     private Class controller;
     private Method controllerMethod;
     private Result result;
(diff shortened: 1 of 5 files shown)


fzakaria:
#270 
I think I added a link to wrong issue in my commit so I'm linking it here

raphaelbauer:
That's interesting... The PR is also well made with testcases. Great!

Can you tell me more about your use case? Why doesn't it work for you to simply repeat the routes as Strings inside the router file? (Yes devil's advocate here.. I wanted to know it anyway to get the usecase :) Thanks! ).

fzakaria:
My thoughts were it can let you compose routes in separate files (perhaps eventually in some module/plugin system). Such that you can feed the main Router and attach sub ones to it.

This was my minimal take on it with not much change to the code

raphaelbauer:
Okay. The way we are doing it right now is to add subrouters at the main Routes.java file. So your module would have its own MyModuleRoutes.java which implements ApplicationRoutes.

You can then include the MyModuleRoutes.java like:

```
public class Routes implements ApplicationRoutes {
    @Inject
    MyModuleRoutes myModuleRoutes; // Just a subrouter implementing ApplicationRoutes

    @Override
    public void init(Router router) {
        // module routes
        myModuleRoutes.init(router);

        // ... more routes...
        router.GET().route("/").with(ApplicationController.class, "index");
    }
}
```

Is this something that helps to fix the issue you have - or would you still think that the enhanced router proposed in your PR would be helpful in your scenario?

PR #274, opened 2014-12-15
support to watch multiple class folders in maven multi module project

christiangroth:
Hey guys, I added support to enable ninjs superdev mode for maven multi module projects with quite minimal changes. If maven project contains modules they are picked up via maven session projects / reactor projects. If not everything works as it was designed before. 

We plan to use this feature with the following simplified maven multi module project:

parent-project
- module-a
- module-b
- module-c

You may call mvn ninja:run on parent-project to enable superdev mode for all modules contained.

### DIFF
--- a/ninja-maven-plugin/src/main/java/ninja/NinjaRunMojo.java
+++ b/ninja-maven-plugin/src/main/java/ninja/NinjaRunMojo.java
@@ -27,6 +27,7 @@ import java.util.List;
 import ninja.standalone.NinjaJetty;
 
 import org.apache.maven.artifact.Artifact;
+import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
@@ -44,22 +45,21 @@ import com.google.common.collect.Lists;
 public class NinjaRunMojo extends AbstractMojo {
     
     /**
+     * The Maven Session.
+     *
+     * @parameter expression="${session}"
+     * @required
+     * @readonly
+     */
+    protected MavenSession session;
+    
+    /**
      * @parameter property="project"
      * @required
      * @readonly
      */
     protected MavenProject mavenProject;
-    
-    /**
-     * Directory containing the build files.
-     * 
-     * For webapps this is usually
-     * something like /User/username/workspace/project/target/classes
-     * 
-     * @parameter property="project.build.outputDirectory"
-     */
-    private String buildOutputDirectory;
-    
+
     /** 
      * @parameter default-value="${plugin.artifacts}" 
      */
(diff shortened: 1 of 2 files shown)


ninja-maven-plugin/src/main/java/ninja/NinjaRunMojo.java
```diff
@@ -191,6 +195,38 @@ public void execute() throws MojoExecutionException {
...
+	// check for modules
+	List<String> modules = project.getModules();
+	if(modules != null && !modules.isEmpty()) {
+	    
+	    // find module project in reactor
+	    moduleLoop: for(String moduleName : modules) {
+    	    	for(MavenProject moduleProject : session.getProjects()) {
+    		
+    	    	    // TODO this may be some kind of hack and does not work if two projects with same artifact id but different group id are contained in reactor
+    	    	    if(moduleName.equals(moduleProject.getArtifactId())){
```
raphaelbauer, line 202: Can you explain what's going on here?

raphaelbauer:
Pretty interesting :) Let's discuss it!

Disclaimer: I am not a big fan of Maven+MavenModules as web projects. That never turned out to be good choice...

ninja-maven-plugin/src/main/java/ninja/NinjaRunMojo.java
christiangroth, line 202: These code lines do 
- check if currently handled maven project has modules declared
- search for module projects in maven session / reactor (all projects build by maven). This is needed due to maven API providing module names only an not MavenProject instances

christiangroth:
I noticed that I missed to add dependencies from module projects to classpath, so I will fix/add this.

Nevertheless this is kind of a "political decision", I think. Currently ninja maven plugin just does not support maven multimodule projects, this patch will enable it. If you're not a big fan of maven and multi modules in web projects, what drawbacks do you expect? I agree that most of the projects may be just a single maven standalone project, but if you plan to create something bigger and more complex, then multi modules help to create and keep a more structured software architecture / desing / you name it.

raphaelbauer:
Looks really good - I'll have a deeper look asap, but it can take some days because of the holiday season... :)

jjlauer:
I'd really like multi-module ninja projects supported out-of-the-box.  I understand your opinion on avoiding them for web projects, but I have just as strong an opinion the opposite way.  If you're building any large web app -- breaking up the project into multiple modules is better design for numerous reasons.  You can somewhat accomplish the same using jetty maven plugin, but it's pretty ugly.  Seems like adding support would be a great help to lots of folks using maven with multi-modules.

raphaelbauer:
Hey Joe, sorry that it takes so long... will be reviewed asap... Cheers, Raphael

raphaelbauer:
One remark - because I am not the Maven Pro... Do you know any good way to unit test / integration test Maven Mojos? Right now the Maven Ninja plugin is not very well tested - and if we keep on adding new features like multi module support it would be necessary having some tests to make sure the features keep on working... Any ideas?

jjlauer:
The latest PR merged into develop should address multi-modules as well as custom watch dirs in a backwards-compatible way.

https://github.com/ninjaframework/ninja/pull/335

raphaelbauer:
That's fixed - right? Can we close the PR?

jjlauer:
Yes, https://github.com/ninjaframework/ninja/pull/335 addresses this.

PR #275, opened 2014-12-18
Fix for: 266

t3hc13h:
The message now prints the type of the first (Extracted) and the extra parameter which is causing the exception. I considered including the parameter name but since that may be replaced at compile time I figured knowing the types would be good enough for the time being.


raphaelbauer:
That's pretty cool :)

Would you mind adding the change to changelog.md and yourself to team.md? In order to merge the code you should also sign the contributor's agreement here: https://docs.google.com/forms/d/1Yasrxa17kYfaNDgbRvn77rM3WTnU_Um0rwz3GfzPp9g/viewform

More about contributing: http://www.ninjaframework.org/developer/getting_started.html

t3hc13h:
Done and done.

raphaelbauer:
Perfect!

Issue #263, opened 2014-11-27
Maven build fail on Windows


thibaultmeyer:
Hi,

I encountered a problem when generating .jar on Windows platform. Unit tests fail on the second subproject (ninja-core) and all other are skipped. Compilation works fine on Linux platform (tested on Debian 7 x64)
###### software version

```
C:\Users\Thibault\Documents\Developpement\ninja>java -version
java version "1.8.0_25"
Java(TM) SE Runtime Environment (build 1.8.0_25-b18)
Java HotSpot(TM) 64-Bit Server VM (build 25.25-b02, mixed mode)

C:\Users\Thibault\Documents\Developpement\ninja>mvn --version
Apache Maven 3.2.3 (33f8c3e1027c3ddde99d3cdebad2656a31e8fdf4; 2014-08-11T22:58:10+02:00)
Maven home: C:\apache-maven-3.2.3\bin\..
Java version: 1.8.0_25, vendor: Oracle Corporation
Java home: C:\Program Files\Java\jdk1.8.0_25\jre
Default locale: en_GB, platform encoding: Cp1252
OS name: "windows 8.1", version: "6.3", arch: "amd64", family: "dos"
```
###### Output log

mvn compile : http://pastebin.com/naAAcV6a (Success)
mvn package : http://pastebin.com/8PF5Q4QA (Fail)

raphaelbauer:
Thanks for the report. I think it's pretty clear where this comes from. Git makes sure that text files have a proper line ending under windows. That is different to Linux and Mac. And the test is not take that into account. 

Could you please check  if https://github.com/ninjaframework/ninja/pull/264 fixes the issue? Thanks!

thibaultmeyer:
Subproject "ninja-core" pass unit tests but not the subproject "ninja-servlet-integration-test" 

mvn package: http://pastebin.com/fd1LPTKm

raphaelbauer:
Can you paste the full log? Thanks!

thibaultmeyer:
Of course: https://dl.dropboxusercontent.com/u/37895949/ninja4_package-log-fail.txt

raphaelbauer:
I actually don't have a Windows machine running. Could anyone else chime in? I think there are some people using and building Ninja on Windows, and I wonder if they have a clue... Thanks!

t3hc13h:
@raphaelbauer Your change looks good on my Win8 machine.

raphaelbauer:
[...]

PR #277, opened 2014-12-20
Fix for: #263 (part 2) utf8 chars breaking integration tests on US Windows

t3hc13h:
#263 Has two parts, the new line break in robots.txt and the integration test failures.

On my US Windows box the default encoding is clobbering the unicode in
the responses when read by NinjectTestBrowser. 

Explicitly setting the encoding on the response streams fixes the PrettyTimeController and PersonController integration tests.


raphaelbauer:
Very good finding! Thanks for pointing that out :+1:

PR #280, opened 2015-01-03
Open freemarker template engine configuration.

jlannoy:
Just two little modifications that will allow :
- to act on the freemarker configuration, by injecting the TemplateEngineFreemarker and calling getConfguration() (Useful to add, for example, user-defined shared variables or methods)
- to override the default suffix for freemarker templates
  Please note it's my very first pull request using github, I hope I did it right.


raphaelbauer:
Hi jlannoy,

I think your changes are pretty cool. Some remarks though:
- We need documentation for that. (You find it under ninja-core/src/site...). If you don't document the two features nobody will use them :)
- We need testcases for both features. Otherwise they are prone to get removed or damaged by another PR.

Otherwise the PR looks really nice. Ah - don't forget to sign the contributor's agreement here: https://docs.google.com/forms/d/1Yasrxa17kYfaNDgbRvn77rM3WTnU_Um0rwz3GfzPp9g/viewform

That's it already...

Cheers,

Raphael

ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
```diff
@@ -333,7 +335,23 @@ public String getContentType() {
...
+     * @param fileSuffix file extension
+     */
+    public void setSuffixOfTemplatingEngine(String fileSuffix) {
+        this.fileSuffix = fileSuffix;
+    }
+    
+    /**
+     * @return the freemarker configuration object
+     */
+    public Configuration getConfiguration() {
```
raphaelbauer, line 351: When can we fiddle with the Configuration object? Anytime? Or only during startup. Is it threadsafe? Imho this should be documented (in the docs and in the javadoc...).

ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
```diff
@@ -333,7 +335,23 @@ public String getContentType() {
...
-        return FILE_SUFFIX;
+        return fileSuffix;
+    }
+    
+    /**
+     * Will override the default file suffix for freemarker templates.
+     * 
+     * @param fileSuffix file extension
+     */
+    public void setSuffixOfTemplatingEngine(String fileSuffix) {
```
raphaelbauer: Would it be possible to init the fileSuffix via a variable in application.conf? That way we can set the fileSuffix in the constructor of this class. More importantly we can then set the FILE_SUFFIX to final - otherwise we might run into threading issues...

raphaelbauer:
Yea... I added two comments... Would be great if you could look into them - imho the remarks are important :) Thanks!

jlannoy:
Hi.

For the documentation, I didn't know I could modify the "site one".

For the test cases, I did one locally (in a "hello world" project), but I do not see the best way to include a test in the ninja trunk. Do I try to enrich the "TemplateEngineFreemarkerTest"? Do I operate inside the blog integration test? Or can I create all the fake object I need (controllers/views...) in ninja-core unit tests?

Jonathan

raphaelbauer:
Regarding testcases - as minimal as possible. But the templating engine is now tested quite heavily via integration tests. I'll prepare a stub to make that simpler... Let me see...

raphaelbauer:
Hey... I just pushed an upgrade to TemplateEngineFreemarkerTest. It can now be used to "test" the rendering without a full fledged integration test. I'd think that something like:

```
verify(templateEngineHelper).getTemplateForResult(any(Route.class), any(Result.class),    eq(".mycustom.ending.ftl.html"));
```

will be enough to test whether setting of the suffix works :)

Let me know if I can help further!
Raphael

jlannoy:
Sorry I cannot find your upgrade to TemplateEngineFreemarkerTest. Where did you pushed it?
(Note that with your example I see what I can do.)

raphaelbauer:
My bad - forgot to merge the PR... now it's in ninja: https://github.com/ninjaframework/ninja/blob/develop/ninja-core/src/test/java/ninja/template/TemplateEngineFreemarkerTest.java

Feel free to improve it :)

raphaelbauer:
actually my PR somehow does not compile on the CI system (works locally with my jdk) - a fix is under way...

raphaelbauer:
hey... I hope that works, but I created a PR for your PR which fixes the tests (and compiles fine).

It's here: https://github.com/jlannoy/ninja/pull/7

One thing that is still missing is the documentation for the website. We should document your feature - otherwise nobody knows that you can change the suffix of the templates :)

jlannoy:
Hi. I already did it locally, but the fact is that I don't understand the good way to fix conflicts (mainly using Eclipse, I was about to check documentation for the console way). It's for the changelog, so not really important ; but I'd rather know how to do it on such a simple file. I'll try to commit everything for sunday.

raphaelbauer:
No worries :) Take your time... I just wanted to help ;)

jlannoy:
Raphael, I don't know if you 've seen the commit. Is it okay for you? I didn't not put any code example, I think that it's only important for people who are working with the framework (I mean, not discovering it) and are looking for something specific with Freemarker.

raphaelbauer:
hey jlannoy, I indeed missed the commit :) Looks really good! I'll merge it asap :) Thanks!

h4ck4life:
Hi everyone, is this changes included in 4.0.5 version?

jlannoy:
No, it's not ; but as it has been merged in the trunk, you'll have it in the next released version.

h4ck4life:
@jlannoy okay, thanks for this change anyway. Really useful :)

PR #287, opened 2015-01-21
Allow routes from abstract controllers with Jaxy annotations

jlannoy:
Hello.

While working on a CRUD application, I needed the feature of having routes defined in an asbtract controller. So that each concrete implementation could share the same route. Of course in my case it has sense to make such an abstract controller that will work the same way with different sub implementation. This is why I documented this feature as an "advanced usage" one. I also added the corresponding unit tests, it seems it works well. Tell me what you think about it.

Jonathan


gitblit:
My initial reaction was `huh`?  But now I understand what you are trying to accomplish and my opinion is `hhmmmm`.

So if the parent class is abstract, it's routes are parented at the implementation level.  And if the parent class is non-abstract, the routes remain at the parent path level?  What happens if there are multiple layers of abstract parents?

This might be a little too complicated for the stock Jaxy routes implementation.  What say you @raphaelbauer ?

jlannoy:
I didn't think about non-abstract parent. But yes, actually only routes from abstract controllers would be registered for corresponding implementation levels.

If there are multiple abstraction levels, it works as you could assume it for Java inheritance : the more precise methods will be used.

raphaelbauer:
@gitblit you are jaxy's father - and if you think it is too complicated for the stock jaxy implementation then it may well be too complicated... This does not mean that I think it's crappy code - en contraire - the code is well written, tested and even with documentation. So that's pretty cool :) 
But first and foremost @gitblit is the ambassador of the module - and ideally it should be a discussion between you two :)

PR #290, opened 2015-01-26
method should not throw generic exception

svenkubiak:
The send method in PostOffice should not through a generic exception, but rather the specific ones. Either that, or we put a try-catch inside the send method. Let me know, what you think.


raphaelbauer:
nice :)

PR #292, opened 2015-01-29
Closed potential resource leaks

svenkubiak:
I closed some streams which where potential resource leaks. I also updated the documentation for the postoffice for my PR #290.

### DIFF
--- a/ninja-core/src/main/java/ninja/NinjaDefault.java
+++ b/ninja-core/src/main/java/ninja/NinjaDefault.java
@@ -16,6 +16,7 @@
 
 package ninja;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.util.Properties;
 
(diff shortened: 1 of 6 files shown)


raphaelbauer:
this is very very very valuable. Just one remark: Could we use the catch with resources statement? Then we don't need the chatty finally statement (http://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html). Apart from that it is pretty cool to make the code more stable and fix those stupid bugs :+1:

svenkubiak:
Okay. I changed that for NinjaDefaults and MimeTypes. The NinjaTestBrowser creates the inputstream based on some inner variables. Any ideas for that?

raphaelbauer:
nope... I think that's perfect. try with resources does not work in all circumstances... thanks!

PR #296, opened 2015-02-03
Added method to add/unset a cookie from context

svenkubiak:
I have added a method to add and unset a cookie from the context.

### DIFF
--- a/ninja-core/src/main/java/ninja/Context.java
+++ b/ninja-core/src/main/java/ninja/Context.java
@@ -668,4 +668,18 @@ public interface Context {
      * @return true if the content type is to set {@code application/xml} 
      */
     boolean isRequestXml();
+
+    /**
+     * Adds a cookie to the servlet response
+     * 
+     * @param cookie Ninja cookie
+     */
+    void addCookie(Cookie cookie);
+
+    /**
+     * Removes a cookie from the servlet response
+     * 
+     * @param cookie Ninja Cookie
+     */
+    void unsetCookie(Cookie cookie);
 }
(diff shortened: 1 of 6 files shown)


ninja-core/src/main/java/ninja/Context.java
```diff
@@ -668,4 +668,18 @@ private HTTP_STATUS(int code) {
      * @return true if the content type is to set {@code application/xml} 
      */
     boolean isRequestXml();
+
+    /**
+     * Adds a cookie to the servlet response
```
raphaelbauer: without "servlet" please - Just "Adds a cookie to the response"

ninja-core/src/main/java/ninja/Context.java
```diff
@@ -668,4 +668,18 @@ private HTTP_STATUS(int code) {
...
+     * 
+     * @param cookie Ninja cookie
+     */
+    void addCookie(Cookie cookie);
+
+    /**
+     * Removes a cookie from the servlet response
```
raphaelbauer: without "servlet" please - Just "Removes a cookie to the response"

ninja-servlet/src/main/java/ninja/servlet/ContextImpl.java
```diff
@@ -332,6 +331,17 @@ public boolean hasCookie(String cookieName) {
 
         return ninjaCookies;
     }
+    
+    @Override
+    public void addCookie(Cookie cookie) {
```
raphaelbauer, line 336: We need a test for both addCookie and unsetCookie. Can be mocked. But we need it to make sure the feature does not get removed or crippled by accident :)

raphaelbauer:
Very useful! I added some tiny remarks :)

svenkubiak:
I have added the unit tests and updated the javadoc. Refactored a deprecated call in testGetRequestUri a long the way ;-)

raphaelbauer:
Delicious!

PR #299, opened 2015-02-05
Added Support for Jax Routes methods without own path

eicnix:
- I added Support for Jaxy Routes methods without own path to allow something like this: 

``` java
@Path("/app")
class Controller {

    @Get
    public Result get() {
        Results.text().render("get.");
    }
}
```

This makes the @Path annotation on a method optional. If there isn't one the path of the class will be used. 
- The method `getPackagesToScanForRoutes` was made public to be able to write custom annotation based Routers that search in more than just one package
- Furthermore the huge init-method was split into smaller pieces

### DIFF
--- a/ninja-core/src/site/markdown/developer/changelog.md
+++ b/ninja-core/src/site/markdown/developer/changelog.md
@@ -1,7 +1,8 @@
 Version X.X.X
 =============
 
- * 2012-02-03 minor refactoring in JaxyRoutes init (lukaseichler)
+ * 2015-02-05 Added Support for JaxRoutes methods without own path (lukaseichler)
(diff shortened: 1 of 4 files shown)


gitblit:
Instead of hard-coding the stock method annotations, how about discovering the annotations annotated with `@HttpMethod`. That allows for use of your contribution with custom http methods.

eicnix:
I added the support of HttpMethod discovery through reflections. 

Methods that have @Path in combination without a HttpMethod will still be discovered. This is intended because it leads to the default fallback to use  GET.

gitblit:
:+1:

raphaelbauer:
Thanks Lukas!

PR #300, opened 2015-02-05
Refactoring

svenkubiak:
I organized the imports in all projects, removed some unused variables, changed deprecated methods calls and re-styled the default system pages. To be honest, I never liked the sad-face :-D

### DIFF
--- a/ninja-core/src/main/java/ninja/RouteBuilderImpl.java
+++ b/ninja-core/src/main/java/ninja/RouteBuilderImpl.java
@@ -21,10 +21,13 @@ import java.util.Arrays;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.Set;
+
+import ninja.params.ControllerMethodInvoker;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 import com.google.inject.Injector;
-import ninja.params.ControllerMethodInvoker;
 
 class RouteBuilderImpl implements RouteBuilder {
 
--- a/ninja-core/src/main/java/ninja/WrappedContext.java
+++ b/ninja-core/src/main/java/ninja/WrappedContext.java
@@ -44,7 +44,7 @@ public class WrappedContext implements Context {
 
     @Override
     public String getRequestUri() {
-        return wrapped.getRequestUri();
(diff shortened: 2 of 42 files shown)


gitblit:
@svenkubiak  I'm sure the restyled error pages are really pretty and an improvement, but this introduces an internet dependency on a web framework that might be used within a firewalled intranet.

As an alternative to depending on a CDN for Bootstrap, what about integrating [Skeleton](http://getskeleton.com) into Ninja's classpath?  That depends on Google Fonts for typography, but that is probably acceptable since it should gracefully degrade( to Arial or Helvetica or whatever) without losing your css styling.

svenkubiak:
@gitblit I get you point. Putting a full-blown Bootstrap with CDN in the template was a little overhasty.

What about...
1.) We put the CSS from Skeleton in the header of the template, so we don't have the css "somewhere" in the classpath. The minified version is less than 4 Kb, so that wouldn't be a problem at all.

2.) We only use one template for the error pages. The templates already only differ on the status code and we already know which error to display as we call the different templates.

gitblit:
Bueno.​

svenkubiak:
@gitblit Correct me if I am wrong, but ninja does look up the specific error page, say 403forbidden.ftl.html, in the application and if none found displays the default default page from the core?

If that is the case, it would not be a good idea to reduce to one template for every error as it removes the option for the user to style each error page separately. I don't no how often this requirement occurs, but it would be an option one would want.

I would still stick with putting the CSS in the templates though.

gitblit:
Right.  There is an error-page template per status code in the core.  They
can be individually overridden in your app.​

Getting down to a single error page template has pros and cons.  I think
there are more pros, though, so I would be in favor of that change.  I
agree with stuffing the CSS into the template.

eicnix:
IMHO removing error specific templates would remove complexity in maintaining these templates. On the other hand an application built with ninja would most likely use these error specific templates to provide the best user experience. 

I would suggest that a basic error template is created and every specific error template uses this template. This approach would decrease the needed maintaining time while keeping the error specific templates.

svenkubiak:
Closing this one, as I saw that we already have better error pages PR #311

Issue #23, opened 2012-08-16
CSRF protection for put post delete


raphaelbauer:
Like in play1 or in django:
csrf https://docs.djangoproject.com/en/dev/ref/contrib/csrf/#how-csrf-works

=> a form must have a token that can be validated

raphaelbauer:
Some more information copied from google groups:
https://groups.google.com/forum/#!topic/ninja-framework/KU4Hr6_3nZ8

 The basics are already implemented: 
- Ninja session cookie can generate a unique id.
- Don't allow javascript to read the session
  => That's the basic line of defense...

What is missing is:
- A simple method to include the unique id into a form (if it's a form request for instance) => A tag or something to include into Freemarker rendering.
- A filter to check whether the post request is valid. If the request parameter of the form is identical to the id in the cookie the request is not forged.

In one of our projects I wrote some custom code to do that, but we should include that into the framework. As I said above I'll attach some code to the issue.

raphaelbauer:
Some code:

AuthenticityFilter:
public class AuthenticityFilter implements Filter {

```
public static final String AUTHENTICITY_TOKEN = "authenticityToken";

@Override
public Result filter(FilterChain filterChain, Context context) {

    Session session = context.getSession();
    String authenticityToken = context.getParameter(AUTHENTICITY_TOKEN);

    if (session.getAuthenticityToken().equals(authenticityToken)) {

        return filterChain.next(context);

    } else {

        return Results
                .forbidden()
                .template(NinjaConstant.LOCATION_VIEW_FTL_HTML_FORBIDDEN);

    }


}
```

}

raphaelbauer:
The filter assumes your request contains a parameter called "authenticityToken". Authenticity token is generated and stored in the session via session.getAuthenticityToken() (already implemented).

But you have to include the parameter either via a query or post parameter right now..

noguespi:
How to display the token in the view ? now I'm using **${session.___AT}** it it the right way ?

example : 

```
<a href="${reverseRoute("controllers.DashboardController","logout","xsrf",session.___AT)}">logout</a>
```

raphaelbauer:
hmm... in theory yes. But I think that's imperfect. getAuthenticityToken creates a new token when it's absent.  session.___AT (in your view) does not do that. 

So if you just include ${session.___AT} in your views you have to make sure that somewhere before in your controller session.getAuthenticityToken() has been called. Otherwise your view.ft.html will complain :)

As I write above: 
[...]

PR #302, opened 2015-02-06
added proposal for authenticity handling #23

svenkubiak:
This is a proposal for authenticity handling from #23 

You call in the template either ${authenticity("form")} or ${authenticity("token")} and it is checked via the filter, when you add the filter to one of your controller methods.

This proposal is open for discussion. I will add tests and update the documentation once we agree on a final way to go.


gitblit:
That's a very good idea to create a new template engine method to cut down on boilerplate. :+1:

svenkubiak:
Like I said. It's open for discussion.

ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
```diff
@@ -263,6 +263,8 @@ public void invoke(Context context, Result result) {
         map.put("reverseRoute", templateEngineFreemarkerReverseRouteMethod);
         map.put("assetsAt", templateEngineFreemarkerAssetsAtMethod);
         map.put("webJarsAt", templateEngineFreemarkerWebJarsAtMethod);
+        
+        map.put("authenticity", new TemplateEngineFreemarkerAuthenticityMethod(context.getSession().getAuthenticityToken()));
```
raphaelbauer: Just pass the context. Right now you force the creation of the session cookie and authenticity token for requests that potentially don't use them. The call to getSession().getAuthenticityToken() should be executed only when the template method is called...

raphaelbauer:
I think your solution is pretty nice :)

svenkubiak:
@raphaelbauer 
I changed the code according to your comment and added the documentation. Tests are coming soon.

I am note sold on the template calls. What would be pretty cool, would be a global macro and just call

```
@<authenticityToken/>
@<authenticityForm/>
```

But to make this work flawless, this would has to be available in the freemarker global namespace and to be made available programmatically. Does anyone know if this is possible at all?

raphaelbauer:
Simplifying the syntax via a macro would be a cool idea. You can ask the Freemarker community at http://freemarker.org/mailing-lists.html . They are really helpful and friendly :)

svenkubiak:
Ok. Lets wait for a reply on stackoverflow http://stackoverflow.com/questions/28472751/is-it-possible-to-create-a-freemarker-macro-programmatically

svenkubiak:
No solid answer on stackoverflow yet. If the PR is okay, I suggest we'll go with it for and maybe do a refactoring afterwards.

t3hc13h:
Couldn't you get the desired effect by registering a TemplateDirectiveModel as a shared variable in the template configuration?

svenkubiak:
@t3hc13h I did try the TemplateDirectiveModel, but the passed TemplateDirectiveBody is null as I just call 

   <@myDirective/>

svenkubiak:
Ok, I figured it finally out. You can now just call

```
<@authenticityToken/>
<@authenticityForm/>
```

@raphaelbauer 
I updated the Pull-Request.

ninja-core/src/test/java/ninja/AuthenticityFilterTest.java
```diff
@@ -0,0 +1,73 @@
...
+    public void init() {
+        authenticityFilter = new AuthenticityFilter(ninjaDefault);        
+    }
+    
+    @Test
+    public void testAuthenticityFail() {
+        when(context.getParameter(NinjaConstant.AUTHENTICITY_TOKEN)).thenReturn("foo");
+        when(context.getSession()).thenReturn(session);
+        when(context.getSession().getAuthenticityToken()).thenReturn("bar");
+        when(ninjaDefault.getForbiddenResult(context)).thenReturn(forbidden);
```
raphaelbauer: you can simplify the code a bit... we don't need ok and forbidden results...

```
Just do a check  Mockito.verify(ninjaDefault).getForbiddenResult(context)
```

or

```
Mockito.verify(filterChain).next(context);
```

assertNotNull and assertEquals is not needed imho...

That makes the code a bit shorter...

ninja-core/src/site/markdown/documentation/html_templating/advanced_topics.md
```diff
@@ -133,6 +133,19 @@ prettyTime allows you to format localized relative dates.
 For instance, if you had a date object that represented yesterday, prettyTime would format that as *1 day ago* in the
 preferred Locale of the request.
 
+### authenticity(...)
```
raphaelbauer, line 136: very cool...

ninja-core/src/main/java/ninja/template/directives/TemplateEngineFreemarkerAuthenticityTokenDirective.java
```diff
@@ -0,0 +1,40 @@
...
+import freemarker.template.TemplateException;
+import freemarker.template.TemplateModel;
+
+/**
+ * 
+ * @author svenkubiak
+ *
+ */
+@SuppressWarnings("rawtypes")
+public class TemplateEngineFreemarkerAuthenticityTokenDirective implements TemplateDirectiveModel {
```
raphaelbauer, line 20: we need a testcase for the directive...

ninja-core/src/main/java/ninja/template/directives/TemplateEngineFreemarkerAuthenticityFormDirective.java
```diff
@@ -0,0 +1,42 @@
...
+import freemarker.template.TemplateModel;
+import freemarker.template.TemplateModelException;
+
+/**
+ * 
+ * @author svenkubiak
+ *
+ */
+@SuppressWarnings("rawtypes")
+public class TemplateEngineFreemarkerAuthenticityFormDirective implements TemplateDirectiveModel {
```
raphaelbauer, line 22: we need a testcase for the directive...

ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
```diff
@@ -263,6 +264,9 @@ public void invoke(Context context, Result result) {
         map.put("reverseRoute", templateEngineFreemarkerReverseRouteMethod);
         map.put("assetsAt", templateEngineFreemarkerAssetsAtMethod);
         map.put("webJarsAt", templateEngineFreemarkerWebJarsAtMethod);
+        
+        map.put("authenticityToken", new TemplateEngineFreemarkerAuthenticityTokenDirective(context));
```
raphaelbauer, line 268: we need a test to make sure authenticityToken and authenticityForm is added to the templating engine... We should really make sure that your great feature does not get destroyed by eg a bogus merge...

You can do that either via a unit test or in the ninja-integration-test project...

raphaelbauer:
Hey Sven - the feature, implementation and documentation is pretty cool. I added some remarks. I think we need more test-cases. Otherwise your great feature may get lost between some bogus merges for instance... This really helps to produce stable software in the long run which is good for all of us :) Hope that makes sense - let me know if I can help!

svenkubiak:
@raphaelbauer 
I had some problems with the test cases, but finally figured it out what the problem was. Should be all good now :-D

Could you help me on the merge conflicts by adding the changelog entry after merging?

ninja-core/src/main/java/ninja/template/directives/TemplateEngineFreemarkerAuthenticityTokenDirective.java
```diff
@@ -0,0 +1,40 @@
...
+    private String authenticityToken;
+
+    public TemplateEngineFreemarkerAuthenticityTokenDirective(Context context) {
+        this.authenticityToken = context.getSession().getAuthenticityToken();
```
raphaelbauer: please do this in the execute method. Otherwise you are always adding the __AT to the session - even if it is not needed...

ninja-core/src/main/java/ninja/template/directives/TemplateEngineFreemarkerAuthenticityFormDirective.java
```diff
@@ -0,0 +1,42 @@
...
+    private String authenticityToken;
+
+    public TemplateEngineFreemarkerAuthenticityFormDirective(Context context) {
+        this.authenticityToken = context.getSession().getAuthenticityToken();
```
raphaelbauer: please do this in the execute method. Otherwise you are always adding the __AT to the session - even if it is not needed...

raphaelbauer: I'll do that locally...

ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
raphaelbauer, line 268: Nevermind - I'll do that locally while merging...

svenkubiak:
Ok, Just let me know, if there is anything left to do ;-)

svenkubiak:
@raphaelbauer 
Was a quick fix ;-)

raphaelbauer:
I just changed the test-cases a bit. Really good work Sven - great feature - great addition to Ninja!

Issue #301, opened 2015-02-06
AssetsControllerTest.java has system specific line separator


raptaml:
https://github.com/ninjaframework/ninja/blob/develop/ninja-core/src/test/java/ninja/AssetsControllerTest.java#L270 
has system specific line separator (\n) for String comparison. 
Test fails on Windows system.

raptaml:
```
diff --git a/ninja-core/src/test/java/ninja/AssetsControllerTest.java b/ninja-core/src/test/java/ninja/AssetsControllerTest.java
index 4212a11..f729206 100644
--- a/ninja-core/src/test/java/ninja/AssetsControllerTest.java
+++ b/ninja-core/src/test/java/ninja/AssetsControllerTest.java
@@ -265,10 +265,10 @@
         assertEquals(Result.SC_200_OK, resultCaptor.getValue().getStatusCode());
         // we mocked this one:
         assertEquals("mimetype", result.getContentType());
-
-        // make sure the content is okay...
-        assertEquals("User-agent: *\nDisallow: /", byteArrayOutputStream.toString());
-
+        // make sure the content is okay but pay attention to system specific line separator
+        String sysLineSeparator = System.lineSeparator();
+        assertEquals("User-agent: *" + sysLineSeparator + "Disallow: /", byteArrayOutputStream.toString());
+        
     }

     ////////////////////////////////////////////////////////////////////////////
```

raphaelbauer:
Great catch :) Would you mind creating a PR? Please follow the guidelines how to contribute code: http://www.ninjaframework.org/developer/getting_started.html

PR #306, opened 2015-02-12
#301 get line separator from JVM

raptaml:
here we go :-)


raphaelbauer:
Perfect! Thanks for fixing the issue!

PR #308, opened 2015-02-13
Link targets in system view templates with ${contextPath} used for "Back to Homepage".

jfendler:
All link targets in Ninja's default system views should point to the homepage of the application, rather than the server homepage.
If ninja is running with a context set, this patch will ensure the "Back to Homepage" links actually point to the root of the application.


raphaelbauer:
Pretty cool!

PR #311, opened 2015-02-15
Re-throw template errors and buffer templates.

t3hc13h:
Added rethrowing of template errors in prod mode so the application can
return it's 500 code or otherwise handle the exception from the main
exception handling code. This requires full buffering of the template
output since we can't guarantee that we haven't flushed some part of the
response to the user. I think in most cases this will not matter, as the
performance tests I ran showed slightly higher GC pressure but a bit
more reqs/s with full buffering.

### DIFF
--- a/ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
+++ b/ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
@@ -318,30 +318,20 @@ public class TemplateEngineFreemarker implements TemplateEngine {
         ResponseStreams responseStreams = context.finalizeHeaders(result);
 
         try {
-            
             // Fully buffer the response so in the case of a template error we can 
             // return the applications 500 error message. Without fully buffering 
             // we can't guarantee we haven't flushed part of the response to the
             // client.
-            
             StringWriter buffer = new StringWriter(64 * 1024);
-            
             freemarkerTemplate.process(map, buffer);
-            
             Writer writer = responseStreams.getWriter();
-            
             writer.write(buffer.toString());
-            
             writer.close();
-            
         } catch (Exception e) {            
             logger.error(
-                    "Error processing Freemarker Template {} ", templateName, e);
-            
-            throw new RuntimeException(e);
-            
+                    "Error processing Freemarker Template {} ", templateName, e);   
+            throw new RuntimeException(e);   
         }
-
     }
 
     @Override
--- a/ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
+++ b/ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
@@ -51,6 +51,7 @@ import freemarker.template.DefaultObjectWrapper;
 import freemarker.template.DefaultObjectWrapperBuilder;
 import freemarker.template.Template;
 import freemarker.template.Version;
+import java.io.StringWriter;
 
 @Singleton
 public class TemplateEngineFreemarker implements TemplateEngine {
@@ -316,12 +317,24 @@ public class TemplateEngineFreemarker implements TemplateEngine {
         
         ResponseStreams responseStreams = context.finalizeHeaders(result);
 
-        try (Writer writer = responseStreams.getWriter()) {
+        try {
             
-            freemarkerTemplate.process(map, writer);
-
-        } catch (Exception e) {
+            // Fully buffer the response so in the case of a template error we can 
+            // return the applications 500 error message. Without fully buffering 
+            // we can't guarantee we haven't flushed part of the response to the
+            // client.
             
+            StringWriter buffer = new StringWriter(64 * 1024);
+            
+            freemarkerTemplate.process(map, buffer);
+            
+            Writer writer = responseStreams.getWriter();
+            
+            writer.write(buffer.toString());
+            
+            writer.close();
+            
+        } catch (Exception e) {            
             logger.error(
                     "Error processing Freemarker Template {} ", templateName, e);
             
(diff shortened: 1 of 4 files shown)


t3hc13h:
Here's a quick summary of my performance test results using ninja-servlet-intergration-test as my benchmark project with warmed up VM (1M requests before test) on a 6 core win 8 box over 100K reqs with the results averaged over 3 runs.

Full buffering:
Collections avg: 48
Req/s avg: 1642

Direct write:
Collections avg: 39
Req/s avg: 1587

raphaelbauer:
That's awesome... We wanted to do some measurements here as well.. So give us some days to review and merge the PR... Needless to say that it looks really good - and it's especially cool that you measured performance before / after the change... Really good work...

raphaelbauer:
Hey @t3hc13h - how did you setup the benchmarking exactly?

ninja-core/src/main/java/ninja/template/TemplateEngineFreemarkerExceptionHandler.java
```diff
@@ -53,38 +52,9 @@ public TemplateEngineFreemarkerExceptionHandler(Logger logger,
...
-            pw.flush();
-            pw.close();
-            
-            logger.log(Level.SEVERE, "Templating error. This should not happen in production", te);
-            
-            
-
-        } else {
-            
+        if (!ninjaProperties.isProd()) {
```
raphaelbauer: I think this pw.println error will be never shown in dev or test ... The te will always be thrown, so the exception will always bubble up leading to a general error page... or did I miss something?

ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
```diff
@@ -316,12 +317,24 @@ public void invoke(Context context, Result result) {
...
             
-            freemarkerTemplate.process(map, writer);
-
-        } catch (Exception e) {
+            // Fully buffer the response so in the case of a template error we can 
+            // return the applications 500 error message. Without fully buffering 
+            // we can't guarantee we haven't flushed part of the response to the
+            // client.
+            
+            StringWriter buffer = new StringWriter(64 * 1024);
```
raphaelbauer, line 325: We need a test to make sure error handling works as we expect it to work ...

ninja-core/src/main/java/ninja/template/TemplateEngineFreemarkerExceptionHandler.java
```diff
@@ -53,38 +52,9 @@ public TemplateEngineFreemarkerExceptionHandler(Logger logger,
 
     public void handleTemplateException(TemplateException te,
```
raphaelbauer, line 53: Not sure - but would it make sense to replace this templateerror handler with something more generic? If we have a mechanism in place that catches exceptions bubbling up - why should it not display a nicely formatted error there - with the added benefit that we can see the all kinds of exeptions in the browser while developing (not just templating expceptions)... just an idea...

ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
```diff
@@ -316,12 +317,24 @@ public void invoke(Context context, Result result) {
...
+            
+            freemarkerTemplate.process(map, buffer);
```
raphaelbauer, line 326: Could you remove the blank lines between the statements?

raphaelbauer:
I adde some remarks. I hope they make sense... Let me know what you think :) I can chime in and do some stuff if you want... It's really cool polishing those parts :)

ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
t3hc13h, line 326: Absolutely, while working with the class it seemed like most of the file had extra spacing, so I used spacing for consistency within the class.

ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
t3hc13h, line 325: Agreed, I'll go ahead and make sure these changes are covered.

ninja-core/src/main/java/ninja/template/TemplateEngineFreemarkerExceptionHandler.java
t3hc13h: Woops! You're correct, the generic error page eats the template error page when we always re-throw.

ninja-core/src/main/java/ninja/template/TemplateEngineFreemarkerExceptionHandler.java
t3hc13h, line 53: So like a dev version of the general error page which does a message+stack print? That would be pretty cool, I'll play around with that idea and open another PR focus on that idea.

t3hc13h:
For the benchmark I used by dev machine (6 core AMD, 8GB RAM, SSD OS drive) as the server, compiled the ninja-servlet-integration-test to a single assembly and ran it directly with stdout piped to null. No particular GC or JVM memory settings, just the defaults. Client lower powered win 8 box on the local 1GB LAN.

My goal for the tests was to create a simple environment in which I could reduce the number of variables impacting the tests and measure change between the two output methods.

Before I started the warm up traffic I attached VisualVM to the process to record GC data and monitor JIT activity. For the warm up traffic I used Apache Bench  to run 1 million requests against the home route. That many requests may have been a bit over kill but that seemed to be the point the JIT activity mostly subsided so I figured that was a safe start point.

After the warm up I executed three test runs with each output method with 100K requests each to the home route three times to get my averages.

I'll admit that my tests are rather ad-hoc and I don't have much experience bench marking JVM applications so if you have a preferred method to load test Ninja or gather stats just let me know and I'll take another pass at this.

raphaelbauer:
Some final remarks: Could you please add one line to changelog.md? Once that's there I'll merge. Really nice improvement!

raphaelbauer:
forget my last comment.. I'll add the line and merge... Thanks!

PR #312, opened 2015-02-15
Using context.getContextPath() as prefix in cookie path.

jfendler:
This PR should restrict the validity of Ninja's cookies to the application context, rather than them being valid on the whole server.

### DIFF
--- a/ninja-core/src/main/java/ninja/session/FlashScopeImpl.java
+++ b/ninja-core/src/main/java/ninja/session/FlashScopeImpl.java
@@ -83,7 +83,7 @@ public class FlashScopeImpl implements FlashScope {
 
                 Cookie.Builder cookie = Cookie.builder(applicationCookiePrefix
                     + NinjaConstant.FLASH_SUFFIX, "");
-                cookie.setPath("/");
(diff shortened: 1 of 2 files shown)


raphaelbauer:
Hi Jens, that's exactly what solves the issue :) We unfortunately need a test-case for that (there are already test-cases, so you just have to extend them. I think it would be cool to add one line to the documentation as well (in ninja-core/src/site/...). Did you already sign the contributor agreement? (http://www.ninjaframework.org/developer/getting_started.html).

Okay - a lot of small things - anyway good work :) It's really helpful fixing those issues!

jfendler:
Hi Raphael,
sorry for the delay in my response, but I don't have too much time these day. The contributor agreement is signed.
I'm afraid I would have to take some more time for the test cases - I won't get to it in the next couple of days. But since the PR is not really critical, I guess the project will survive some more days without it being merged. I'll write some once time permits.

raphaelbauer:
Hey @jfendler I added the testcases briefly. Thanks for reporting and fixing the bug. Very nice!

PR #313, opened 2015-02-17
Session cookie encryption

bazi:
We use Ninja framework and in our deployment we needed session cookies to be encrypted. This was achieved by injecting a modified implementation of `Session` interface.
This PR introduces session cookie encryption into Ninja so that others can make use of it.
Here are some important points:
- session encryption is **optional**, it is disabled by default
- enabling it is pretty simple -- ~~a secret key should be included in `application.conf` with `application.cookie.secret` property~~ add `application.cookie.encryption=true` 
- ~~DES~~ AES algorithm is used ~~by default~~ -- mainly because it is strong and is one of the widely supported algorithms of every implementation of the Java platform
- ~~encryption algorithm **can be customized** by implementing `CookieEncryptionKeyGenerator` interface and binding the interface type to that implementation~~
- encrypted session cookie value without sign is encoded in **base64** - it is more efficient in terms of space (which is important as cookie size is limited)

Testing:
- `SessionCookieTest` is updated and run two times by JUnit's `Parameterized` runner
  1. existing unencrypted session cookie tests
  2. encrypted session cookies tests

Documentation is also updated with brief description. Further updates are appreciated.


raphaelbauer:
That's exciting! Really great! Give us some days to check out the code in detail.

balamaci:
Hi, care to please explain how the concept is secure - don't want to attack this, I just need to understand please. 
I mean let's say we choose to keep the username inside the cookie if he was successfully authenticated and on subsequent requests we just run Cookie.decrypt() and if this gives a valid "usename=george" we know that the user was authenticated and the cookie was created by us. Since the cookie cannot be tampered(and another username injected), this should be ok. 

But what stops an attacker to find the secret key by this method: 
The attacker suspects we keep the username in the cookie. He then creates an user's account("george") and receives an encrypted cookie(say "aabb99"). By knowing the encryption algorithm(most ninja apps will have the default alg probably) he can try a lot of values for the private keys Cookie.decrypt("aabb99", "key_try1..9000000") until the decrypted string contains "george" that means he found the right key. He then is able to create himself cookies which whatever user he wants and we'll consider that user authenticated.

metacity:
@balamaci: I think it comes down to the same principal that applies to the current HMAC-SHA1 cookie signing: the secret key just has to be long enough (like currently: 64-char alphanumeric, mixed case) for bruteforcing to be unfeasible.

bazi:
@balamaci: if one wants to encrypt session cookies then maintainer shall provide a strong secret key against brute-forcing as @metacity pointed above.
Securing against cookie tampering is already being handled by current HMAC signing. This PR is more about readability of a cookie - currently it is and by default will be in plain text; when encrypted - only secret key owner can read them and that secret key should be private and strong enough against brute-forcing.

Related notes about this PR:
- default secret key generator gives a warning if provided secret key material is weak or semi-weak
- as noted above, encryption algorithm can be customized to use more strong algorithms than the default DES

raphaelbauer:
First of all great work. The implementation is done very well, tested and documented!

Some remarks from my side:
1. In general we should follow the example of Rails 4 (reverse engineered here: http://big-elephants.com/2014-01/handling-rails-4-sessions-with-go/). Afaics you do that mostly...
2. One thing that follows from that is to use AES/265, DES does not cut it any more. All major JREs should support AES.
3. One idea: Can we use the same application.conf key for signing and encrypting? Imho (I may be wrong) we should use only one of the configuration parameters (application.secret / application.cookie.secret). Maybe we could anyway use application.secret and a configuration variable "application.cookie.encryption = true" to enable the encryption. We can leave that to false by default (in version 4.1) and set it to true in a later Ninja version.
4. Maybe you could also upgrade the code generating the initial application secret (SecretGenerator.java) - there are better ways to do so (http://stackoverflow.com/questions/18228579/how-to-create-a-secure-random-aes-key-in-java).

These are my initial thoughts. I hope they make sense and I really hope we can merge your feature. Really good work :)

bazi:
really good remarks @raphaelbauer, totally agree.
I am glad you liked this work so let me update this PR to address your remarks

raptaml:
@bazi and @raphaelbauer thank you for your work! I will need this feature in my projects. This was one of Ninjas very few weak spots. Looking forward to seeing the merge.

bazi:
Regarding remarks from @raphaelbauer : 
- AES/256 is used to encrypt session cookies. Just make sure JCE Unlimited Strength Jurisdiction Policy Files are installed if encryption is used (also noted in docs)
- application secret is used to encrypt cookies, no separate key for session encryption. Application secret is expected to be a base64 encoded AES key with 256-bit key size (in case of invalid secret, encryption won't work with log messages proposing deleting that secret so that Ninja can auto-generate new one)
- SecretGenerator.java is upgraded
- custom key generator mechanism for session cookie encryption is removed - we already will be using strong algorithm, and application secret is used anyways

raphaelbauer:
Whohooo!

Looks good.

Unfortunately I am getting an error:

```
20:51:01.685 [qtp335471116-17] ERROR ninja.utils.CookieEncryption - Failed to decrypt data. Check if application secret is a valid base64 encoded AES key.
You can remove 'application.secret' key in configuration file and restart application. Ninja will generate new key for you.
java.security.InvalidKeyException: Illegal key size or default parameters
    at javax.crypto.Cipher.checkCryptoPerm(Cipher.java:1021) ~[na:1.8.0_25]
    at javax.crypto.Cipher.implInit(Cipher.java:796) ~[na:1.8.0_25]
    at javax.crypto.Cipher.chooseProvider(Cipher.java:859) ~[na:1.8.0_25]
    at javax.crypto.Cipher.init(Cipher.java:1229) ~[na:1.8.0_25]
    at javax.crypto.Cipher.init(Cipher.java:1166) ~[na:1.8.0_25]
    at ninja.utils.CookieEncryption.decrypt(CookieEncryption.java:106) ~[ninja-core-4.0.6-SNAPSHOT.jar:na]
    at ninja.session.SessionImpl.init(SessionImpl.java:120) [ninja-core-4.0.6-SNAPSHOT.jar:na]
    at ninja.servlet.ContextImpl.init(ContextImpl.java:125) [ninja-servlet-4.0.6-SNAPSHOT.jar:na]
    at ninja.servlet.NinjaServletDispatcher.service(NinjaServletDispatcher.java:82) [ninja-servlet-4.0.6-SNAPSHOT.jar:na]
    at com.google.inject.servlet.ServletDefinition.doServiceImpl(ServletDefinition.java:279) [guice-servlet-4.0-beta5.jar:na]
    at com.google.inject.servlet.ServletDefinition.doService(ServletDefinition.java:269) [guice-servlet-4.0-beta5.jar:na]
    at com.google.inject.servlet.ServletDefinition.service(ServletDefinition.java:180) [guice-servlet-4.0-beta5.jar:na]
    at com.google.inject.servlet.ManagedServletPipeline.service(ManagedServletPipeline.java:91) [guice-servlet-4.0-beta5.jar:na]
    at com.google.inject.servlet.FilterChainInvocation.doFilter(FilterChainInvocation.java:85) [guice-servlet-4.0-beta5.jar:na]
    at com.google.inject.servlet.ManagedFilterPipeline.dispatch(ManagedFilterPipeline.java:119) [guice-servlet-4.0-beta5.jar:na]
    at com.google.inject.servlet.GuiceFilter$1.call(GuiceFilter.java:133) [guice-servlet-4.0-beta5.jar:na]
    at com.google.inject.servlet.GuiceFilter$1.call(GuiceFilter.java:130) [guice-servlet-4.0-beta5.jar:na]
    at com.google.inject.servlet.GuiceFilter$Context.call(GuiceFilter.java:203) [guice-servlet-4.0-beta5.jar:na]
    at com.google.inject.servlet.GuiceFilter.doFilter(GuiceFilter.java:130) [guice-servlet-4.0-beta5.jar:na]
    at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1650) [jetty-servlet-9.2.3.v20140905.jar:9.2.3.v2014
```

I am getting the error when running ninja-servlet-integration-test via mvn ninja:run
and then calling http://localhost:8080/session

Yes - I re-created the application.conf secret correctly (afaics).

Can you reproduce that?

bazi:
I performed following steps and it all worked as expected. Will be looking for reasons but i want to make sure if **JCE Unlimited Strength Jurisdiction Policy Files** are installed on your JRE? AES/256 requires that.

When i run `ninja-servlet-integration-test` via `mvn ninja:run` cookies are not encrypted and everything works fine.
Then i set `application.cookie.encryption=true` and run again - cookies are not encrypted, i have the above errors in logs.
I remove/unset `application.secret` and run again - Ninja generates new secret and everything works fine, cookies are encrypted.

raphaelbauer:
Hey @bazi bazi, you were absolutely right - it was because of the JCE policy files and the 256 bit key (regular JDK apparently only support 128bit). What lead me to think it may be a bad idea to rely on it. Because in real life deployment environments you just have a regular JRE and are not guaranteed to have strong encryption readily available. 

Solutions to that are twofold: 
- Either we stick with 128 bit (should be ok).
- We use an external library: like https://www.bouncycastle.org/java.html for AES/256 encryption. Then we can just add it as dependency and it works out of the box.

What's your opinion?

bazi:
I updated PR to use 128-bit key. It is quite good and we are just encrypting sessions cookies :)
BouncyCastle is a good library but we have a good enough choice so i think it is better not to introduce extra dependency.

raphaelbauer:
Hey,

I got two more comments after discussing the feature with my colleagues I got two more more requests. 
- We should keep the application secret (and the method to create the session cookie) the way it was before (length 64 mixed case). Sorry - that was my idea and my fault. I think it is stupid to artificially limit it to the characters supported by AES. If we keep it the old way the cookie hash uses the full length secret and is strong as it can be.
- The AES crypt should take the first 16 letters of the application secret (should be 128 bit if I am not mistaken) to encrypt the content. Afaicr there is a method in Java's crypto utils to ask for the allowed length of the algorithm secret. That way 128bit don't have to be hard coded. Just pick the first n letters of the application secret and you are ready to go.

Does that make sense?

Cheers,

Raphael

bazi:
Ok, PR is updated. We are getting better...
- current application secret (generation, etc.) is kept - all previous related changes are reverted
- AES key uses first N letters of the application secret. N is defined by `Cipher.getMaxAllowedKeyLength()` value divided by byte size (which is 8 as we all know). If hosting JRE has unlimited strength policy installed then key bit-size is set to 256 which would use 256/8=32 first letters of the application secret (and that's ok as app secret has length 64)

raphaelbauer:
Works perfectly fine. I think that's near - perfect. Are there any other opinions on the code? Would be great to get some other feedback as well. Well done!

raphaelbauer:
I refactored some tiny bits and added a testcase for CookieEncryption. Thanks for that great contribution. Very nice work!

bazi:
Good, now we have session encryption capability out of the box!
From refactorings of @raphaelbauer, I see that encryption/decryption failures now throw runtime exception instead of just logging failures and continuing without encryption. It makes sense because it is better to have errors if not properly set instead of having encryption enabled but going unencrypted until discovered.
Note for developers - **clear your previous Ninja cookies** in your client if you enable encryption.

PR #316, opened 2015-02-19
Do not include .java files in published jar

jjlauer:
Fixes including .java files in final binary jar


raphaelbauer:
That's pretty cool. Thanks for figuring that out!

PR #317, opened 2015-02-19
Use maven basedir for launched JVM process

jjlauer:
The current ninja maven plugin assumes that the user will run "mvn" from "user.dir" system property (the directory the user was in when they launched maven).  A more robust method is to use the basedir property of the maven project class at runtime.  The main issue is for a multi-module project, the user will likely initiate maven from the parent project.  That causes hot-reloading of templates to fail since the maven ninja plugin assumes src/main/java is relative to the working directory -- but really its relative to the maven project running the plugin.

This pull request fixes this issue.  I've tested it on single and multi-module maven projects and it works great.


raphaelbauer:
nice!

PR #318, opened 2015-02-20
Dependency fix

jjlauer:
This commit optimizes ninja-core's dependencies to reduce pulling in extra libraries -- where they are not necessary.  Before these couple tweaks, my assembled tarball of a Ninja-based app weighed in around 31.5 MB in dependencies (ninja-standalone + ninja-ebean + mysql-connector).  After these two small changes, I shrunk it by 19% down to 25.6 MB.  Folks who aren't using the ninja-ebean or mysql-connector libraries will have even more success.

ninja-core: org.jsoup
- Doing a grep on "org.jsoup" for example found zero results
- Seemed like an old/dead dependency
- Saves 0.3 MB for final runtime lib dependencies

ninja-core: ehcache-core rather than entire full ehcache
- ehcache-core just contains vital parts to make caching work
- ehcache is the full distro with its server component, etc.
- Saves 5.6 MB for final runtime lib dependencies

It looks like ehcache's new v3 series will also continue the idea of distributing just a "core" library. Folks who need the entire distro can still add the dependency to their own project.

### DIFF
--- a/ninja-core/pom.xml
+++ b/ninja-core/pom.xml
@@ -84,12 +84,6 @@
             <artifactId>prettytime</artifactId>
             <version>3.2.1.Final</version>
         </dependency>
-
-        <dependency>
-            <groupId>org.jsoup</groupId>
-            <artifactId>jsoup</artifactId>
-            <version>1.7.2</version>
-        </dependency>
     
         <!-- Thread-safe alternative for Java JDK time stuff -->
         <dependency>
@@ -206,9 +200,11 @@
         </dependency>
         
         <!-- ehcache implementation -->
+        <!-- IMPORTANT: only pull in core (~1MB) vs. entire distro which includes a server (~7MB) -->
         <dependency>
-            <groupId>net.sf.ehcache</groupId>
-            <artifactId>ehcache</artifactId>
+            <groupId>net.sf.ehcache.internal</groupId>
+            <artifactId>ehcache-core</artifactId>
+            <version>2.8.5</version>
         </dependency>
         
         <!-- memcached client -->
--- a/ninja-core/pom.xml
+++ b/ninja-core/pom.xml
@@ -204,7 +204,6 @@
         <dependency>
             <groupId>net.sf.ehcache.internal</groupId>
             <artifactId>ehcache-core</artifactId>
-            <version>2.8.5</version>
         </dependency>
         
         <!-- memcached client -->
--- a/pom.xml
+++ b/pom.xml
@@ -501,9 +501,10 @@
             </dependency>
         
             <!-- ehcache implementation -->
+            <!-- IMPORTANT: only pull in core (~1MB) vs. entire distro which includes a server (~7MB) -->
             <dependency>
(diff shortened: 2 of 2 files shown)


bazi:
Cleaning out dependencies is good, i like that. But seems your recent PRs overlap a bit. I saw these changes in other PRs like in #316. I think it is better to keep each PR scope clean.

jjlauer:
Gotcha. Since I wasn't sure if including .java files in the published jar was intentional or not -- I treated them as different PRs.

raphaelbauer:
Hey @jjlauer - that's a very cool improvement. Thanks for working that out!

svenkubiak:
I am having build problems with a clean local maven repository. Looks like the transitive dependency to org.terracotta.internal:statistics:jar:1.0.1 from the new ehcache-core dependency is not in maven central. Can someone confirm that?

jjlauer:
Looks like the pom for ehcache-core declares a few additional repositories for its own dependencies.  Something not recommended by M2/Sonatype, but not enforced.  I'm guessing you are running a M2 proxy or enforcing security of some sort for yourself -- most users wouldn't notice and maven will just add the additional repo.

I'd guess this repo declared would be where it is: http://www.terracotta.org/download/reflector/releases

raphaelbauer:
Works for me...

svenkubiak:
My settings.xml was fetching from http://repo1.maven.org/maven2/. Without settings.xml fetching from https://repo.maven.apache.org/maven2 works like a charm. Sorry for the inconvenience.

raphaelbauer:
No worries :) It's always great finding the root cause of a problem!

svenkubiak:
Sorry that I have to get back to this, but it looks like that you can not fetch the terracotta dependency when using a mirror. I have a Nexus with default configuration (no security enforcements etc.) and dependency fetching is done through a mirror configuration in my settings.xml. The mirror is configured to use my nexus. When using no settings.xml maven will fetch by default from https://repo.maven.apache.org/maven2, which works. When I am fetching with my settings.xml through the mirror, maven can not fetch the terracotta dependency. 

Can someone confirm this issue?

jjlauer:
It seems like your default configuration via the nexus mirror you're
personally using would break transitive dependencies from adding additional
repositories for their own transitive dependencies.  That's why I
originally asked you if you were using a mirror/proxy for yourself.

It seems like this is a config issue specific for yourself.  You'll
eventually hit other projects that do the same thing -- its not a rule
everything is always in maven central.  I've seen nexus configurations that
simply add additional upstream repos they mirror.  Why not add the
terracotta repo I mentioned in my earlier post on this issue?

If this change is backed out (others are having build issues as well) --
then maybe an alternative is to move ehcache to its own module so folks can
swap in/out other cache implementations w/o having to manually exclude it.

svenkubiak:
Sure, I could add the terracotta repo (and will do). But this is the first time I ran into dependency issue with my nexus and ninja. And as this is a default nexus configuration others might run into this problem as well.

PR #321, opened 2015-02-23
On-The-Fly minification of js and css assets

svenkubiak:
I recently took a look at the impressive java full stuck framework wisdom (worth a look at http://wisdom-framework.org/). Wisdom has some really nice features. On of them is on On-The-Fly minification of assets. 

I think this could be a really good enhancement for ninja, so I created a proof of concept. Here is how it works:
- I changed the WatchAndRestartMachine to check if a change on a asset (js|css) occured. 
- If a changed occured it will grab the content from the source file and create a new file adding min.css or min.js. Existing files will be overwritten. The JVM will not be restarted.
- The idea is, that the user uses the .min.js or .min.css in the frontend, but works in the raw files.
- ATM this only works in SuperDevMode, basically because as a developer I want to know, if the minification is correct.

Additional thoughts:
- <strike>The current implementation assumes that you store your css in "stylesheets" and js in "javascripts", which could be a good convention, but I think it would be nice to make this configurable via the application.conf. But I did not get the injection of NinjaProperties to work (yet).</strike>
-  <strike>Same goes for the outputpath, which currently is hard-coded to /src/main/java/assets </strike>

UPDATE: Added configuration via application.conf

Let me know, what you think.

Cheers,
Sven

### DIFF
--- a/ninja-core/pom.xml
+++ b/ninja-core/pom.xml
@@ -252,6 +252,15 @@
             <groupId>com.icegreen</groupId>
             <artifactId>greenmail</artifactId>
         </dependency>
+        
+        <!-- Minification -->
+         <dependency>
+	     	<groupId>com.yahoo.platform.yui</groupId>
+		     <artifactId>yuicompressor</artifactId>
+		     <version>2.4.7</version>
+		     <!-- 2.4.8 seems to have issue on windows : https://github.com/yui/yuicompressor/issues/78 -->
+		     <!-- 2.4.8 failed to process empty file (demo01) : https://github.com/yui/yuicompressor/issues/130 -->
+	    </dependency>
 
     </dependencies>
 
--- a/ninja-core/src/main/java/ninja/utils/MinificationUtils.java
+++ b/ninja-core/src/main/java/ninja/utils/MinificationUtils.java
@@ -1,0 +1,152 @@
+package ninja.utils;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+
+import org.mozilla.javascript.ErrorReporter;
+import org.mozilla.javascript.EvaluatorException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.yahoo.platform.yui.compressor.CssCompressor;
+import com.yahoo.platform.yui.compressor.JavaScriptCompressor;
+
+/**
+ * Convenient class for minification of css and js assets
+ * Based on
+ * https://github.com/davidB/yuicompressor-maven-plugin/blob/master/src/main/java/net_alchim31_maven_yuicompressor/YuiCompressorMojo.java 
+ * 
+ * @author svenkubiak
+ *
+ */
+public final class MinificationUtils {
+    private static Logger LOG = LoggerFactory.getLogger(MinificationUtils.class);
+    private static final String OUTPUTPATH = "/src/main/java/assets/";
+    private static final String ENCODING = "UTF-8";
+    private static final String JS = "js";
+    private static final String CSS = "css";
+    private static final boolean DISABLEOPTIMIZATION = false;
+    private static final boolean PRESERVESEMICOLONS = false;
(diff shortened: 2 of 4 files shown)


raphaelbauer:
Hmm. Not sure actually. There are many very good ways how to make asset minification get to work (and to integrate them into a proper packaging lifecycle).

Some ideas:
- plain sass or grunt via a separate Terminal
- https://github.com/allegro/grunt-maven-plugin
- Move over to gradle and use their processes to run separate tasks (there is already a gradle ninja:run prototype btw).

and so on... Maybe a separation of concerns makes sense. My client side apps are way to complicated to be managed by something like ninja:run.

Anyone else?

raphaelbauer:
Follow up: Here is the gradle plugin: https://github.com/DazHudson/Ninja-Gradle-Plugin - I strongly think that gradle would be the way to go. Otherwise our ninja:run mode will become a second build and packaging system... (imho - and I may be wrong...).

svenkubiak:
Besides Gradle, one could use a Maven Plugin like https://github.com/davidB/yuicompressor-maven-plugin which handles minification. I am using this in some of my projects and my first thought was to integrade the plugin. But we don't have a maven lifecycle when in superdev mode. So, I am stuck at either change .js to .min.js in the template everytime i change something or having a second shell for running the minification and keep .min.js in the template when using the maven plugin for minification.

But I was looking at this from a developer perspective, specially a frontend developer. As a frontend developer I want to work in the raw js/css, but don't want to change the calls to .js and .min.js in the templates, so I always want .min.js calls in the template. And I want to know if minification works correctly. So having this feature would be at least a benefit for a frontend developer (IMHO).

I agree with you, that having the minification done in a maven lifecycle would be a much cleaner way of doing this. But IMHO it has its downsides. Maybe a combination of both could be a solution?

Cheers,
Sven

UPDATE: Just to avoid misunderstanding. The proposed minification is a continues process. Everytime you change the raw js/css file the min.js/min.css file will be updated. It is not a run once when packaging. At least it is not intend to. And now with configuration from application.conf it is disabled by default and you can configured the minification for js and css seperatly.

raphaelbauer:
Okay - my setup usually is to have two terminals. One terminal runs ninja:run, the other terminal runs (eg) yuicompressor-maven-plugin. Whenever a js file changes yui generates the compresses files. ninja:run does not restart, because the change happens in the assets directory. But ninja picks up the changes because in dev mode the AssetsController reloads the files always.

When packaging the application I make sure that the (eg) yuicompressor-maven-plugin runs first, and then the application is packaged correctly. This can be done in one pom file - or by using two maven projects and one parent pom.

This works pretty well for me (I actually used this in conjunction with Gwt).

svenkubiak:
Well, what I am proposing perfectly fits in your setup, but you don't have to open a second shell anymore :-D

The implementation also does not restart ninja:run and just minifies the js/css assets when the "raw" file is changed. And it is disable by default, so the user has the option to enabling it. And you can sill use yuicompress-mave-plugin when packaging if you want to. But this also becomes optional with my proposal.

jjlauer:
I've been thinking about my own asset pipeline workflow in my projects and I wanted to share something I wrote which may change your thinking a bit.

It's my own "watcher" plugin for maven.  The idea is that rather than rely on your IDE for compiling your classes (which isn't very reliable in a large project) during SuperDevMode -- I simply wanted to open a second shell and have maven continuously trigger a "compile" if a source file changes.  Then I also gain the benefit of using maven's process-classes, etc. phases as it would work even outside of Ninja.

https://github.com/fizzed/java-maven-plugins

I planned on sharing this with the Ninja community in the future, but thought it was worth sharing now during this discussion.  You could actually just add onto this idea to add minification as a maven plugin during the process-classes or generate-sources phase.  This watcher plugin would simply work out-of-the-box for you and no changes needed for the SuperDevMode code. 

The obvious benefit is that you can add any sort of maven plugin you need for your asset pipeline.  Same pipeline then used during SuperDevMode or if you're just packaging your app on the command line.

Thoughts?

magikid:
I'm not sure that I agree that this feature belongs in a web framework but I do like that it is opt-in by default.  One thing I'd like to see is a test to go along with this.

PR #324, opened 2015-02-27
ExceptionResult on server error or bad request with underlying cause inc...

jjlauer:
Template engines currently have no way of accessing the underlying Exception (cause) of an internal server error or bad request.  While applications can override NinjaDefault and provide their own implementation and include the exception in the Renderable -- template engines rely on NinjaDefault to setup the Result.

I first tried to include the exception as a new key/value pair in the existing Renderable, but that broke existing applications that count the Renderable as a Message object in the case of a system error.  Thus, I created a new ExceptionResult class that extends Result (akin to AsyncResult) to optionally include the underlying cause of an exception.  The benefit is its backwards compatible and template engines can choose to "unwrap" the result if they wish to inspect the result for an exception.

I've been working on super detailed error pages for DEV mode.  I am considering a PR for that down the road, but I wanted to make sure including the exception was acceptable before I put more energy into it.

### DIFF
--- a/ninja-core/src/main/java/ninja/ExceptionResult.java
+++ b/ninja-core/src/main/java/ninja/ExceptionResult.java
@@ -1,0 +1,68 @@
+/**
+ * Copyright (C) 2012-2015 Joe Lauer
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ninja;
+
+/**
+ * A result when an error occurs in the application (e.g. internal server error).
+ * Will optionally include an <code>Exception</code> as the underlying cause for
+ * the error. Template engines may choose to use <code>ExceptionResult.getCause()</code>
+ * and/or <Code>Result.getRenderable()</code> during the rendering process to provide
+ * a detailed response.  A template engine may handle this type of result
+ * as follows:
+ * 
+ * <code>
+ * public void invoke(Context context, Result result) {
+ *       
+ *       if (result instanceof ExceptionResult) {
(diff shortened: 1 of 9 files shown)


t3hc13h:
Hi @jjlauer, I've been doing a little hacking on a similar feature for DEV mode errors so I wanted to share my ideas here in case you find them useful. Either way, I'll wait to see what you come up with :)

My thought was that we could introduce a new template for DEV mode to which we would pass as much information as can be gathered. We could build out a page detailing the exception, Request and Response objects, as well as whatever else we could pull in, caching, configuration, maybe DB? I'm drawing inspiration from use of [Glimpse](http://getglimpse.com) and [Elmah](https://code.google.com/p/elmah) both of which gathers a huge amount of information and are very useful tools.

Anyways, I look forward to seeing what you come up with.

jjlauer:
Hey Andrew, I was initially working on stacktrace of exception + snippet
of project source code / template source that caused it.   I'm probably
about a week out from having it ready for another pair of eyes to look at
it.  I really like your idea of adding additional info like the request /
response.  I never thought about modules adding diagnostic info, but I
could see why introspecting the cache would be useful in some cases.  I'd
venture we design a method for modules to plugin their diagnostic features
-- but I'm guessing that'll have to be really well designed so it doesn't
actually caused exceptions as well!

I currently have a working stand-alone template (not relying on freemarker)
in a single page format that displays a snippet of source, some info on the
context of the request, and the stack trace.  For rendering exceptions, I
display a  snippet of the template that caused it.  I'll add a link to my
forked commit next week for you to take a peek at.  Perhaps you'd like to
help add additional info / redesign the page and join in on an eventual PR.

jjlauer:
Hey @t3hc13h, I committed my initial diagnostic support to https://github.com/jjlauer/ninja/tree/diagnostic_errors a few minutes ago.  I'm going to keep working on it -- although if you wanted to check out my progress go ahead.

If you do pull my branch down and "mvn install" it -- then when you use that build you'll currently need to add "application.diagnostic=true" to your application.conf as well as be in DEV mode for diagnostic errors to be enabled.

jjlauer:
Hey @raphaelbauer -- while I'm still working on diagnostic support (if you'd like to check out my branch mentioned above I'd really appreciate your feedback as well) -- I still think it would be nice to pass on the exception to the template engine in this PR as well.

jjlauer:
@t3hc13h Just submitted my PR for diagnostic mode

https://github.com/ninjaframework/ninja/pull/329

jjlauer:
Diagnostic error support should make this PR not as useful.

Issue #303, opened 2015-02-07
Error 500 on JSON form submission


thibaultmeyer:
Hi,

They are an issue on the JSON request parsing. If JSON document is empty, the server parser crash (missing  try/catch ?) and return the error code 500.

If I replay the same scenario with an standard request (application/x-www-form-encoded) and no data : the validation process works and Validation object contain error like "this field can't be null".

It could be nice to have the same behavior regardless of the chosen method (JSON, Form-Encoded, XML) and in all case, don't raise error 500 during the form validation process.

_I have tested with cURL and Chrome extension : Advanced Rest Client_
##### My form object

```
public static class ProfileForm {
    @NotNull
    @NotBlank
    public String first_name;

    @NotNull
    @NotBlank
    public String last_name;
}
```
##### ApplicationController

```
public Result updateProfile(Context ctx, @JSR303Validation ProfileForm f, Validation v) {
    if (!v.hasViolations()) {
        // No errors : return object
        return Results.jsonp().render(f);
    }

    // Errors
    return Results.jsonp().render(v);
}
```
##### Example 1 (JSON) : OK (Validation::hasViolations() return false)

```
{
    first_name : "John",
    last_name : "Smith"
}
```
##### Example 2 (JSON) : OK (Validation::hasViolations() return true)

```
{
}
```
##### Example 3 (JSON) : CRASH (http error 500)

```
// empty JSON (length: 0)
```

Console output :

```
13:06:19.311 [qtp66233253-15] ERROR n.bodyparser.BodyParserEngineJson - Error parsing incoming Json
com.fasterxml.jackson.databind.JsonMappingException: No content to map due to end-of-input
...
at ninja.bodyparser.BodyParserEngineJson.invoke(BodyParserEngineJson.java:51) ~[ninja-core-4.0.5.jar:na]
at ninja.servlet.ContextImpl.parseBody(ContextImpl.java:279) [ninja-servlet-4.0.5.jar:na]
at ninja.params.ArgumentExtractors$BodyAsExtractor.extract(ArgumentExtractors.java:299) [ninja-core-4.0.5.jar:na]
at ninja.params.ValidatingArgumentExtractor.extract(ValidatingArgumentExtractor.java:40) [ninja-core-4.0.5.jar:na]
```
##### Example 4 (Form Encoded) : OK (Validation::hasViolations() return false)

```
first_name=John&last_name=Smith
```
##### Example 5 (Form Encoded) : OK (Validation::hasViolations() return true)

[...]

PR #325, opened 2015-02-28
Bug #303 - JSON / XML body parser

thibaultmeyer:
fix critical issue on JSON / XML parsing : when submitted document is empty (size: 0 bytes) or malformed, the method hasViolations() return false. This fix will correct this issue and the method hasViolations() now returning true when the document was not valid

### DIFF
--- a/ninja-core/src/main/java/ninja/bodyparser/BodyParserEngineJson.java
+++ b/ninja-core/src/main/java/ninja/bodyparser/BodyParserEngineJson.java
@@ -16,49 +16,50 @@
 
 package ninja.bodyparser;
 
-import java.io.IOException;
-import java.io.InputStream;
-
-import ninja.ContentTypes;
-import ninja.Context;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
+import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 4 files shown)


raphaelbauer:
That's pretty nice. What about using something like this:

```
        try (InputStream inputStream = context.getInputStream()) {
            t = objectMapper.readValue(inputStream, classOfT);
        } catch (JsonMappingException ex) {
            throw new BadRequestException("Error parsing incoming Json", ex);
        } catch (IOException e) {
            logger.error("Error parsing incoming Json", e);
        }
```

The BadRequestException is usually what we should use. When the request says it's a json, but can not be parsed, then it's a bad request. Let us know if that fits the bill.

Apart from that we'd need some testcases to make sure the feature is stable :)

thibaultmeyer:
Ok, now if something goes wrong during the parsing of a JSON / XML document, the application will return a 400 error.

I have added the same behavior in the IOException catch statement to handle the case where a JSON / XML document was not "complete".

By instance, this XML document leads to an IOException during the parsing process :

```
<s>
    <first_name>John</first_name>
    <last_name>Do</last_name>
</>
```

raphaelbauer:
That's pretty cool! It would be great if you could extend the test-cases to make sure the behaviour does not get removed by accident :)

raphaelbauer:
Looks good :) Very nice!

PR #328, opened 2015-03-05
Changed the HTTP-Content parsing from Class<T> to TypeReference<T>

eicnix:
The current request processing creates for the request's body object something with the same class but instead of the actual object a hashmap is created. Which leads to castexceptions if this object is cast to the actual expected class. 

This pullrequest changes the current jackson usage of `body.parse(Class<T>)` to `body.parse(TypeReference<T>)` from which jackson can create proper objects.

### DIFF
--- a/ninja-core/src/main/java/ninja/Context.java
+++ b/ninja-core/src/main/java/ninja/Context.java
@@ -21,7 +21,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.List;
 import java.util.Map;
-
 import ninja.bodyparser.BodyParserEngineJson;
 import ninja.bodyparser.BodyParserEngineManager;
 import ninja.bodyparser.BodyParserEngineXml;
@@ -29,9 +28,7 @@ import ninja.session.FlashScope;
 import ninja.session.Session;
 import ninja.utils.ResponseStreams;
 import ninja.validation.Validation;
-
 import org.apache.commons.fileupload.FileItemIterator;
-import com.fasterxml.jackson.core.type.TypeReference;
 
 public interface Context {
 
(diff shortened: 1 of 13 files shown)


gitblit:
This PR makes Ninja even more dependent on Jackson and is likely to adversely affect non-Jackson  BodyParserEngines.  Can you elaborate on what your proposal does that the current design does not do?

eicnix:
As described [here](https://stackoverflow.com/questions/6846244/jackson-and-generic-type-reference) `body.parse(Class<T>)` creates a custom LinkedHashMap object with the same name as the Controller-Method-Paramter. Casting this object to the actual Type of the parameter causes a ClassCastException because they are two different objects. 

Using `body.parse(TypeRefence<T>)` creates a new instance of the identical type as the parameter.

gitblit:
I see.  You are trying to deserialize a generic type instead of a concrete type and the parameterized type is lost on compilation due to erasure.  End result is the best Jackson can do is deserialize a hashmap.

Perhaps instead of using the Jackson `TypeReference` we could use Guava to capture the parameterized type information.  The Jackson parsers could then use the Guava type to create/use a `TypeReference`.  This would be friendlier to non-Jackson body parsers.

eicnix:
I removed the Jackson-TypeReferences from all the Interfaces. They are now only used inside the default bodyparsers that are build with jackson.

gitblit:
So we're back to `Class<T>`... but doesn't that defeat the purpose of your request?   You are still unable to deserialize a parameterized type _unless_ you are bypassing the Ninja interfaces and manually casting & using the Jackson BodyParsers.  Maybe I'm missing something.

eicnix:
My propose would be that the default BodyParsers that are using jackson achieve this functionality without having to use jackson classes in the various interfaces. 

If you are using the default BodyParsers this will work now out of the box. If you are building an application with ninja everything will work the same way except that you get proper objects on generic body types. 
Custom BodyParser would need to achieve the generic deserialisation themselves.

gitblit:
> you get proper objects on generic body types

Could you add a test for that?  The current test is for a concrete type which already worked.

PR #329, opened 2015-03-05
Diagnostic dev mode w/ detailed error pages

jjlauer:
This PR implements a "Diagnostic" extension to DEV mode.  While in DEV mode -- its incredibly useful to see what went wrong rather than the default PROD/TEST mode system/view templates.  I added docs to src/site which covers how it works.  Copying that here is an easy way to describe what this does:

---

As of version 4.0.7, Ninja features a new diagnostic extension to dev mode where detailed diagnostic error pages are returned from <code>ninja.NinjaDefault</code> rather than the typical Result which pulls a template from <code>system/views</code>.

For example, in versions prior to 4.0.7, if your controller class method threw an <code>Exception</code> then the default behavior of <code>ninja.NinjaDefault</code> was to catch it in the <code>onException</code> method and return a Result using the template <code>views/system/500internalServerError.ftl.html</code>. The default system views are basic and more intended for use in production than during development.

As of version 4.0.7, the default behavior of all the methods in <code>ninja.NinjaDefault</code> are to first check if dev mode is on and if <code>NinjaProperties.areDiagnosticsEnabled()</code> is true.  If both are true then a <code>Result</code> is returned with the renderable set to a <code>DiagnosticError</code> instance.  Since <code>DiagnosticError</code> implements <code>Renderable</code> -- it knows how to render itself to the output stream and it will bypass the <code>TemplateEngine</code>.

Diagnostic mode is disabled automatically in PROD/TEST modes, but it can also be disabled in DEV mode. Simply add the following to your <code>application.conf</code>:

```
application.diagnostics=false
```

---

You'll also notice that I added a new <code>RenderingException</code> as a first class citizen to the project. Rather than relying on a <code>TemplateEngine</code> to render its own error -- it rather throws a RenderingException which is now handled by NinjaDefault.  I believe this is the correct behavior where all errors are eventually routed thru NinjaDefault.  The side benefit is that _all_ template engines can now throw this exception and take advantage of Diagnostic mode.  Applications can also now override the error in the case of a template rendering exception.

I have a sample project demonstrating all the common errors if you'd like to try it out:

https://github.com/jjlauer/java-ninja-diagnostic-demo

Here is a sample screenshot of a diagnostic error page:

freemarker parser exception


gitblit:
Pretty.

sojin:
Really useful!

t3hc13h:
Dang, that's pretty cool. Nice work!

raphaelbauer:
that looks awesome... I'll have a look asap :)

svenkubiak:
That looks really cool :+1:

metacity:
That is awesome! :+1:

raphaelbauer:
Hey Joe, we just had a deeper look into your code. The conclusion is that it's really exciting and will end up soon in master.Very nice piece of work - well tested and with documentation.

We nevertheless wanted to discuss some things and try out some edge cases. So it will take some days until we merge it. But it will happen.

jjlauer:
Hey @raphaelbauer -- sounds good.

BTW - I considered a number of other approaches (e.g. Guice interceptor). The only major drawback I could see in the design of what I submitted is that folks who override things via "conf.Ninja" will lose diagnostic mode if they don't call the super method and decide to return that result rather than a complete overridden result.  A pretty simple change IMO if apps are interested in keeping diagnostic errors.

Just let me know if you think names need changed, more unit tests, etc.

PerfectCarl:
Would it be nice to add the template code also where the error occurs and display a nice excerpt of the failing code?

image

jjlauer:
hey @PerfectCarl -- the current impl does show the template and offending line.  The sample pic I included with the original PR is just using a 1-liner template so it's not as obvious, but its there.

jjlauer:
hey @PerfectCarl -- here's a sample that's a little more obvious for templates:

freemarker parser exception

And here's one for application code:

application exception

PerfectCarl:
@jjlauer Then that's perfect as it is! Great work !!

> Support for highlight.js could be added in a future (minor) patch

raphaelbauer:
Congrats - very nice work Joe! Just made it into main repo... (with some very minor changes like headers and so on) :+1:

jjlauer:
Thx for merging.  Gave it a whirl and everything works.

PR #333, opened 2015-03-14
Refactored standalone, new features, lots of unit tests

jjlauer:
This PR maintains backwards compatibility with ninja-standalone (except as noted below) and adds a couple new features and includes the addition of unit tests (which were not present earlier).
- Unit tests now written (quite a few of 'em)
- Updated docs w/ the new features
- New standalone "ninja.host" system property allows you to control which address
  Jetty will bind to at startup.  This is useful if you plan on hosting ninja behind a proxy (e.g. ngnix) and simply want to bind it to localhost for simple security (rather than iptables).
- New standalone "ninja.jetty.configuration" system property allows a comma-delimited
  list of jetty.xml files to allow as much custom/advanced configuration of jetty
  as a power user demands.  Multiple http, ssl, or even spdy, etc.  What's nice is that you don't need to package your app as a WAR, but you still can configure jetty however advanced you'd like.  Please note that this feature also required adding "jetty-xml" as a dependency.  Its tiny and I think useful enough to justify including it.
- Standalone mode incorrectly handled exceptions - kept running even if startup failed.
  main() would actually return, but process would continue to run with Jetty daemon
  threads in background.  If app fails to start, then main() exits with a proper error code returned to the shell.
- If server does start, main() no longer exits and relies on Jetty daemon threads to 
  keep JVM alive.  main() now joins() with Jetty server.
- Best effort at unwrapping startup exceptions down to InjectorException - passed up
  chain from ServletContextListener to NinjaBootstrap to NinjaJetty.  Improves (slightly) the slew of guice injector errors at startup.  Maybe we can even use this in the future to cut down on the very large list of injector errors at startup now that we have the original exception.

The only caveat on compatibility:
- The start() method now throws a checked exception.  The old impl threw a runtime exception.  I decided it was better to really throw the exception for why the server failed to start vs. another wrapped exception (cuts a little bit more down on logging if fails at boot).  This also forces main() to catch this exception w/ a compile-time check -- which is why one of the bugs above was not caught earlier.


raphaelbauer:
Very nice!

raphaelbauer:
changelog.md update is missing..

raphaelbauer:
The PR is really nice! If you want to increase the "time-to-merge-into-develop"(TM) it would be cool if you could create atomic PRs with really distinct features. That would really help to review the PRs quickly in the future :) That's just a note - you contributions are really great - we really want to see them in develop quickly!

jjlauer:
Gotcha.  Appreciate the feedback.  I have another very large PR coming
today for the maven-plugin.

I needed the ninja.host feature and then sorta kept unwrapping the onion
leading to this PR.  I'll try to keep my PRs as tight as possible -- since
I also would love to get them into develop more quickly.

I'll update the changelog soon.

On Mon, Mar 16, 2015 at 12:42 PM, Raphael A. Bauer <notifications@github.com

ninja-standalone/src/test/java/ninja/standalone/NinjaJettyTest.java
```diff
@@ -0,0 +1,283 @@
...
+import org.apache.commons.io.IOUtils;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ *
+ * @author Joe Lauer (http://twitter.com/jjlauer)
+ */
+public class NinjaJettyTest {
```
raphaelbauer, line 36: So much better with a test!

ninja-standalone/src/test/java/ninja/standalone/NinjaJettyTest.java
```diff
@@ -0,0 +1,283 @@
...
+    public void startAndShutdownWithDefaults() throws Exception {
+        // absolute minimal working version of application.conf
+        System.setProperty(NinjaProperties.NINJA_EXTERNAL_CONF, "conf/minimal.conf");
+        
+        NinjaJetty nj = new NinjaJetty();
+        // only way to make this test not rely on a port is to at least set the port
+        nj.setPort(randomPort);
+        
+        try {
+            Assert.assertEquals(new Integer(randomPort), nj.port);
```
raphaelbauer: could you statically import Assert? Would make the test more concise...

ninja-standalone/src/main/java/ninja/standalone/NinjaJetty.java
```diff
@@ -30,58 +28,88 @@
 
 import com.google.inject.Injector;
 import com.google.inject.servlet.GuiceFilter;
+import java.io.FileNotFoundException;
+import java.net.URI;
+import org.eclipse.jetty.server.ServerConnector;
+import org.eclipse.jetty.util.resource.Resource;
+import org.eclipse.jetty.xml.XmlConfiguration;
 
 public class NinjaJetty {
```
raphaelbauer, line 44: Could you add a logger? And convert all system.\* things to logger messages?

ninja-standalone/src/main/java/ninja/standalone/NinjaJetty.java
```diff
@@ -30,58 +28,88 @@
...
         
-        ninjaJetty.start();
+        setJettyConfiguration(System.getProperty(COMMAND_LINE_PARAMETER_NINJA_JETTY_CONFIGURATION));
+        
+        try {
+            
+            this.start();
+        }
+        catch (Exception cause) {
+            System.err.println("Unable to start server. ");
```
raphaelbauer: Should be done via a logger

ninja-standalone/src/main/java/ninja/standalone/NinjaJetty.java
```diff
@@ -90,111 +118,196 @@ public Injector getInjector() {
...
+            jettyConfigurationFile = Resource.newClassPathResource(jettyConfiguration);
+            
+            if (jettyConfigurationFile == null || !jettyConfigurationFile.exists()) {
+            
+                throw new FileNotFoundException("Unable to find jetty configuration file either locally or on classpath [" + jettyConfiguration + "]");
+                
+            }
+        }
+        
+        System.out.println("Using jetty configuration file to configure server: " + jettyConfigurationFile);
```
raphaelbauer: Should be done via a logger

ninja-standalone/src/main/java/ninja/standalone/NinjaJetty.java
```diff
@@ -90,111 +118,196 @@ public Injector getInjector() {
...
+        } catch (Exception cause) {
+            
+            // check if there was an underlying injector exception (most likely cause)
+            if (ninjaServletListener.getNinjaBootstrap() != null &&
+                    ninjaServletListener.getNinjaBootstrap().getInjectorException() != null) {
+                
+                // the injector exception is actually what we want thrown
+                throw ninjaServletListener.getNinjaBootstrap().getInjectorException();
+                
+            }
```
raphaelbauer: Ninja's codebases uses else on the same line as "}" like

```
} else {
```

That may sound really picky... but uniformity is important in the long run as the codebase grows :) Thanks!

ninja-standalone/src/main/java/ninja/standalone/NinjaJetty.java
```diff
@@ -30,58 +28,88 @@
...
             
         });
         
+        try {
+            
+            // do not simply exit main() -- join server
+            server.join();
+        }
+        catch (InterruptedException e) {
+            // server probably shutting down
```
raphaelbauer: should we really not log the interrupted exception?

ninja-standalone/src/main/java/ninja/standalone/NinjaJetty.java
jjlauer: So, I think some lines can be converted to logger, but main()'s are hard w/ logging in the case of a startup failure.  What if logging configuration fails -- you may miss the exception.  Thoughts?

ninja-servlet/src/main/java/ninja/servlet/NinjaBootstrap.java
```diff
@@ -191,7 +206,15 @@ protected void configureServlets() {
             return injector;
 
         } catch (Exception exception) {
-            logger.error("Fatal error booting Ninja", exception);
+            
+            // save exception in case other consumers of this class would like
+            // to handle an injector exception a little differently
+            injectorException = exception;
```
raphaelbauer: Why is that necessary? Would it not be enough to simply catch the exception one level above this class? I am just asking because all the isInjector... getInjectorException... etc makes a lot of noise in different classes.

ninja-standalone/src/main/java/ninja/standalone/NinjaJetty.java
jjlauer: Yeah, no worries, I'm picky about that stuff too.  When working previously in NinjaDefault.java -- looks like some of that code uses that style.  No problem converting it here though.

raphaelbauer:
Hey Joe, a lot of really good stuff here. It's really nice to see the classes getting polished more and more :) I added some remarks (nothing dramatic), it would be nice to get some feedback about them... Thanks!

ninja-servlet/src/main/java/ninja/servlet/NinjaBootstrap.java
jjlauer: So, I dislike how I implemented this as well, so its a good discussion.  The real problem is this line:

Preconditions.checkNotNull(injector, "Ninja injector cannot be generated. Please check log for further errors.");

The initInjector() method catches the real exception, logs it, then returns null.  Therefore the real exception causing a startup failure was hidden and merely logged.  Then the Preconditions.checkNotNull method is used to actually trigger the exception and your told "Please analyze log...".  NinjaJetty only got a RuntimeException of the "Please analyze log" exception.

I would actually have preferred to modify initInjector to throw an exception and not return null.  I just wasn't sure that was safe to do since it seems like initInjector was designed to NOT throw an exception.  So the safer route for me was to at least save the original exception and not log it -- so that it could be logged correctly in NinjaJetty.

BTW - Guice startup errors are one of the most annoying aspects of Guice (IMO).  You get 30+ errors and you really only care about the first few.  With the real exception in NinjaJetty, someone could eventually submit a PR to optimize for that error to help drill down quickly to the real problem.  That's why I cared about getting to the real exception.

ninja-standalone/src/main/java/ninja/standalone/NinjaJetty.java
jjlauer: Unfortunately, NinjaBootstrap is where logging is initialized.  Anything before "server.start()" (pretty much everything in NinjaJetty) cannot count on logging being available yet.  One workaround would be to pull out logging initialization out of NinjaBootstrap so it can be the absolute first thing done in the app and then NinjaBootstrap would skip it. I haven't dug into how logging is configured as of yet, so not sure if that's even a good idea.  Thoughts?

ninja-servlet/src/main/java/ninja/servlet/NinjaBootstrap.java
raphaelbauer: If you can rewrite the initInjector to throw exceptions in a better way go ahead!

Btw reply: Yes - guice can be pretty terrible when it comes to error messages. That's one of the value propositions of dagger - but there are cons for dagger as well... So guice is quite okay right now :)

jjlauer:
Hey @raphaelbauer -- I had a rebase mixup on the fixes I submitted -- I think I got that sorted out now.

I made the changes you requested and replied to your comments.  Just let me know what you think.

Thx.

ninja-standalone/src/main/java/ninja/standalone/NinjaJetty.java
raphaelbauer: Depends. Ninja's logging config is only relevant if you did not set the config via -Dlogback.configurationFile. Otherwise the whole config is skipped and logback configures itself without Ninja. We could init the configuration even further above if we get hold of ninjaproperties (which we can do if we use NinjaPropertiesImpl).

jjlauer:
Hey @raphaelbauer -- I researched logging and it was very safe to convert System.out to a logger. Based on testing various configurations, even with a totally broken logback config, logger will still print output to console.  So, good suggestion.

Also, I removed the injector exception design I previously used.  initInjector now throws a checked exception which is re-thrown as an unchecked exception in boot().  Very good suggestion on changing this.  Since initInjector was originally private -- and boot() threw a runtime exception anyway -- should be no issue with this small change.  NinjaJetty now gets the original exception with no kludge.

I concur on Dagger, btw.  It's good for Android apps, but not very good at server dependency injection. Guice still a great choice.

raphaelbauer:
The force is strong with this PR...

PR #335, opened 2015-03-16
Refactor of maven plugin, new features, new unit tests

jjlauer:
I kept peeling back the onion and I ended up overhauling the maven plugin.  This is a large PR and I suspect it may take awhile to get reviewed/merged.

I think the normal Jetty maven plugin is slow (built when polling FS was required to detect changes).  It is unfortunate may users still use it since the SuperDevMode plugin lacked some features.  I think this overhaul fixes that and puts it in great shape to have users add features down the road.  I've been using this for a couple weeks on Windows, Linux, and Mac and it's been working flawlessly for single and multi-module projects.

Should be 100% backwards compatible. Tons of new useful features. Custom watch directories, automatic multi-module support. Redesigned so that the heart of the plugin can easily be reused as the basis for any other build tool plugins (e.g. Gradle, SBT, etc.).

Bugs fixed:
- Fixed bug w/ fast restarts leaving java processes still running on system
- Fixed bug where new files weren't logged
- Fixed bug where new files failed to have exclusion rules applied

New features:
- Include patterns now supported. Useful since negative exclude patterns  are complicated.  Includes are a short-circuit match, then fallback to excludes.
  
  mvn ninja:run -Dninja.includes="(.*)watcher.txt$"
- Support for watching multiple directories. Normal include/exclude rules apply to files in these too.
  
  mvn ninja:run -Dninja.watchDirs=src/main/java,src/test/java
- Support for multi-module projects (watching all directories on runtime classpath). Normal include/exclude rules apply to files in these too.
  
  mvn ninja:run -Dninja.watchAllClassPathDirs=true
- Support for watching all transitive dependencies on classpath.  Yes, you can  install your jar from an unrelated project to your local repo and trigger a restart.
  
  mvn ninja:run -Dninja.watchAllClassPathJars=true
- New "settle down" timeout so fast file system changes in succession only trigger a single restart.  Very important when a user compiles multiple classes at a single time -- and only results in 1 restart when they finish.  Sample console output:
  
  [INFO] Restarting SuperDevMode (126 file change(s) detected)
- New "debug" mode to get a very detailed view of what's happening with plugin.  Will print out info such as the include/exclude rule that applied to a file, any file changes detected, what happened with restarted process, etc.
  
  mvn -X ninja:run
- All configuration properties now have maven associated system property names to make
  overriding them via the command-line much easier. For example:
  
    mvn ninja:run -Dninja.port=8081
- New "skip" configuration property added.  Potentially useful in multi-module projects where you can limit execution to the project you'd like without having to resort to a profile.

New design
- Complete unit test coverage of the core components.  Maven ninja plugin itself dumbed down to entirely setup configuration only (since unit testing execution is very difficult w/ maven these days).
- All the heavy lifting is done by classes in the "ninja.build" directory.  It's possible these could be moved to their own module so they can be repurposed for other build tool plugins.
- Switched to ZeroTurnaround's excellent ProcessExecutor library -- more reliable output stream redirecting, termination, and shutdown hooks.  Simplified code quite a bit.
- Logging uses Maven 3's support for SLF4J. Plugin output integrates into standard Maven output much better.
- Maven plugin switched to use annotations for properties, etc.  Added aliases to properties so they can be set easily via command-line:
  
  mvn ninja:run -Dninja.excludes="(.*)"
- Added help plugin so goal automatically created from annotations.  Sample usage
  
  mvn ninja:help -Ddetail=true -Dgoal=run


jjlauer:
Hey @raphaelbauer -- wondering what your thoughts were on this PR?  While a large refactor, I'm sure you'll appreciate there is a good start of unit tests to build upon for future changes down the road.  Plus, I'm really getting stuck on the bugs I mentioned in *develop branch -- and so far this PR has been solid.

raphaelbauer:
hi @jjlauer all good :) I just did not yet have time to look into it in detail... will happen soon...

raphaelbauer:
Hey Joe,

finally I found some time to look into your code. One word: Awesome! You did a great job improving the dev mode. And finally the dev mode has some good testcases as well. There is much I like about the stuff you did. That begins with the testcases and the use of tempusfugit. Then the much much better annotation based NinjaMojo. But also the very nice improved DelayedRestartTrigger. 

All in all beautiful code and a perfect contribution!

I just removed some spaces here and there and rearranged some brackets.

Thanks!

Raphael

jjlauer:
You made my day :-)

PR #337, opened 2015-03-17
minor fix in ninja-servlet-integration-test

bazi:
This is a minor fix in `ninja-servlet-integration-test` module. Session test case was failing when session encryption is enabled.


raphaelbauer:
What exactly should this PR fix? Thanks! Raphael

bazi:
if session cookie encryption is enabled for `ninja-servlet-integration-test' unit tests, then there is a test failure. This PR checks if session encryption is enabled and if so skips cookie plain text asserts.

raphaelbauer:
But the testcase assumes that the cookie is not encrypted - so the test is absolutely ok... Or do I miss something?

PR #338, opened 2015-03-18
Added Rocker template module to docs

jjlauer:
Adding Rocker templates to module docs


raphaelbauer:
wohooo!!!! That sounds really exciting...

PR #345, opened 2015-04-07
Postoffice should not throw generic exception

svenkubiak:
This is a follow-up to PR #290, originally fixed in PR #298, but I can't remember why i closed #298. Anyway, here is the fix.


jlannoy:
Outdated - replaced by #632

PR #346, opened 2015-04-16
Updated HTTP-Method Annotation resolving

eicnix:
When using JaxyRoutes as a dependency the HttpMethod-Annotations(GET,POST, etc.) were not discovered and no routes were created for methods using only these annotations. 

This PR widens the search for HttpMethod-Annotation and now the whole classpath is scanned once.

### DIFF
--- a/ninja-jaxy-routes/src/main/java/ninja/jaxy/JaxyRoutes.java
+++ b/ninja-jaxy-routes/src/main/java/ninja/jaxy/JaxyRoutes.java
@@ -16,6 +16,20 @@
 
 package ninja.jaxy;
 
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+
+import org.reflections.Reflections;
+import org.reflections.scanners.MethodAnnotationsScanner;
+import org.reflections.scanners.SubTypesScanner;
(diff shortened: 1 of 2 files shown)


raphaelbauer:
Nice! Thanks for reporting and fixing the issue!
@gitblit Does that look good to you?

ninja-jaxy-routes/src/main/java/ninja/jaxy/JaxyRoutes.java
```diff
@@ -207,7 +207,10 @@ public int compare(Method m1, Method m2) {
         Set<Method> methods = Sets.newLinkedHashSet();
 
         methods.addAll(reflections.getMethodsAnnotatedWith(Path.class));
-        Reflections annotationReflections = new Reflections("", new TypeAnnotationsScanner(), new SubTypesScanner());
+	    Reflections annotationReflections = new Reflections(new ConfigurationBuilder()
```
raphaelbauer, line 210: 2 tiny things
1) There are tabs - please change that to spaces - we should keep the code base uniform...
2) Can you think of a good way to "test" the functionality? In other words - how can we prohibit someone to replace the new code with the old one without us noticing the mistake?

raphaelbauer:
There seem to be test failures:

Failed tests:   testWithoutMethodPath(controllers.ApplicationControllerTest): (..)

PR #348, opened 2015-05-05
Add NinjaException.getHttpStatus()

icoloma:
Without a getter method there is no way to retrieve the status code associated to an exception, which seems like a valid concern if anybody wants to make their own subclasses of `NinjaException` (like `InsufficientPermissionsException` or whatever) and customize the JSON error messages returned.


raphaelbauer:
... yea! that makes a lot of sense! Good contribution! Would you mind adding one line to changelog.md?

Then we'll merge asap!

icoloma:
Done! (to the best of my knowledge, that is)

PR #365, opened 2015-06-09
added missing properties in maven plugin

momiji:
The missing properties are ninja.external.configuration and ninja.external.reload
This allows to provide them on mvn command line with -D arguments, and they are propagated to the java started in the maven  plugin.

### DIFF
--- a/ninja-maven-plugin/src/main/java/ninja/maven/NinjaRunMojo.java
+++ b/ninja-maven-plugin/src/main/java/ninja/maven/NinjaRunMojo.java
@@ -19,6 +19,7 @@ package ninja.maven;
 import ninja.build.WatchAndRestartMachine;
 import ninja.build.DelayedRestartTrigger;
 import ninja.build.RunClassInSeparateJvmMachine;
+
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.FileSystems;
@@ -31,12 +32,12 @@ import java.util.Set;
 
 import ninja.standalone.NinjaJetty;
 import ninja.utils.NinjaConstant;
+import ninja.utils.NinjaProperties;
 
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.project.MavenProject;
-
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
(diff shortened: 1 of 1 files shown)


momiji:
this is my very first pull on github or any public repository, plese be gentle with me, I'll do my best to improve myself in the future...

raphaelbauer:
Hey @momiji , thanks for the contribution! We'll have a look asap :)

momiji:
Ok forget it for the moment, I feel like this was the wrong way to achieve my purpouse.

I believe now that the best should be:
- add a ninja.jvmArgs as String[], to be injected as is in jvmArguments
- this will be more generic and allows to add any argument, whatever is needed

The same result will be obtained by adding to the command line:
- `-Dninja.jvmArgs=-Dninja.external.configuration=xxx.conf`
- `-Dninja.watchDirs=.` to restart jetty when configuration file changes

Note: String[] params can be entered on command line by separating different values by `,`.

My personal usage is now:
- loading external configuration
- allow remote debug
- watch for changes on external configuration file

```
mvn ninja:run \
   -Dninja.jvmArgs="-Dninja.external.configuration=dev.conf -Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=8000" \
   -Dninja.watchDirs=.
```

What do you think ?

momiji:
OK I updated the PR with new code, I think it's much better now, and more generic...

danielsawan:
Do you manage to make it auto-reconnect to jetty while debuging ? Because the ninja hot reloader just kill jetty and start it again. So the debug port is just closed and started again and then we have to reconnect eclipse to the jetty.

https://github.com/ninjaframework/ninja/issues/283#issuecomment-70612427

I was wondering if it is possible to connect eclipse directly to ninja:run process and maintain this connexion even if jetty is reloaded. And the ninja:run forwarding the jetty debug port to the ninja:run instance. So the debug port would be manage by ninja and not jetty. If jetty is down for reloading then no problem. When it is up again the ninja:run would automaticly reconnect to the newly created jetty instance without having to reconnect eclipse to jetty manualy.

Don't know if i am clear enought my english is so poor :/

raptaml:
You could use following parameters at startup to attach to running JVM instance:
-Xdebug -agentlib:jdwp=transport=dt_socket,address=9999,server=y,suspend=n
should work for jetty reloading.
You would have to add it to ninja:run goal in your project config. I use netbeans, so no idea for eclipse.

momiji:
@danielsawan: no there is auto-reconnect feature, but it allows to debug after reattaching from eclipse.
@raptaml: ok for debug, haven't tested this way, but now one can add as many options as wanted, like external configuration, and so on.
I just pushed an update to rename the argument, as it is used in the pom.xml and I believe jettyArguments is less clear and consitent than jvmArgs.

Hope this will be included in future release.

momiji:
Hello, any news on this one ? To recall, it allows to send any custom command line arguments to the jvm running ninja when using the ninja maven plugin.
I use it for 2 things :
- set external confiuration to use a dev.conf, that overrides some application.conf settings I don't want to commit in git
- allow debug by adding the necessary -Xdebug and so on

Do you need more information on this, or should I do something more ?

raphaelbauer:
The PR is in the queue :) Sorry for the delay but a lot is happening right now :)

momiji:
this may replaced with my other PR #382 , as I now have undertand I should use different branches and not "develop" if I want to create multiple PR for different needs

raphaelbauer:
Is this really closed?

Issue #88, opened 2013-04-09
Abstractize Commons-Upload behind a Ninja-specific API


tbroyer:
…this would allow using AppEngine's Blobstore (in ninja-appengine) or Servlet 3.0 native multipart support (or any non-servlet container's native support).

raphaelbauer:
Sounds good :) Afaik we only used interfaces of commons upload in client facing code.

PR #368, opened 2015-06-18
Handling multipart requests

bazi:
First of all let me note that this PR is not complete yet and needs community's feedback.
## What this PR targets

Parameter injection into controller methods does not work yet with `multipart/form-data` requests. This issue was addressed in mail group [here](https://groups.google.com/forum/#!topic/ninja-framework/YgPaszB6LqY) where @raphaelbauer proposed to make `@Param("...") String param` work with multipart requests and additionally introduce something like `@File("...") File file` that directly injects uploaded file into controller's method.
Besides, i found an old issue #88 that roughly says to hide commons-upload behind some Ninja API.
Well, this PR is all about those issues.
## What's already done

Almost everything to handle multipart requests. There are new annotations `@FileParam` and `@FileParams` that can be applied to controller method's parameters. For instance, by adding  `@FileParam("myfile") File file` into your controller method you have a ready `java.io.File` instance to read uploaded file contents from.
Additionally, every form parameter of a multipart request (i.e. those that are not attached files) can be injected by `@Param` annotation.
### ~~Cleanup of uploaded files~~

When multipart request are parsed, uploaded files are locally saved in temp directory. I need your feedback on how to clean up those files.
- First thing that comes to my mind is to use `finalize` method of `Context` implementation(references to files are stored in `Context`).
- Or there may be some aggregator of temporary files which deletes them after some time (thinking of a **scheduler**).

What are your suggestions? Can we leave cleanup responsibility to application developers or app administrators? I guess not :)
### ~~Commons-upload~~

Well there are no issues with the library ;) This is more related to issue #88.
Currently to handle file uploads we use `Context.getFileItemIterator()` which returns us an iterator of multipart items aka _parts_. This is the recommended way of handling parts, it is described in [Streaming API](http://commons.apache.org/proper/commons-fileupload/streaming.html). As its name implies parts are _streamed_ and you have to handle them _once_. In this PR, multipart requests are parsed automatically (i.e. iterator is used) and all related parameters and files are available for injection. So `Context.getFileItemIterator()` becomes useless -- it won't anymore return an iterator that you _can_ iterate through _parts_.
I thought to put `@Deprecated` on that method but deprecated methods shall have their old behavior for migration period but in our case that won't be the case. How hard would migration be if that method is removed?


ninja-servlet/src/main/java/ninja/servlet/ContextImpl.java
```diff
@@ -134,6 +143,56 @@ public void setRoute(Route route) {
...
+                    List<File> ls = files.get(name);
+                    if (ls == null) {
+                        ls = new ArrayList<>();
+                        files.put(name, ls);
+                    }
+                    ls.add(target.toFile());
+                }
+            }
+        } catch (FileUploadException | IOException ex) {
+            logger.error("Failed to parse multipart request data", ex);
```
raphaelbauer: Hmm... should we throw the error? So we can fail fast - or just silently ignore that parsing went wrong and log... not 100% sure...

ninja-servlet/src/main/java/ninja/servlet/ContextImpl.java
```diff
@@ -134,6 +143,56 @@ public void setRoute(Route route) {
...
     @Override
+    public void parseMultipart() {
+
+        FileItemIterator it = getFileItemIterator();
+        if (it == null) {
+            return;
+        }
+        try {
+            while (it.hasNext()) {
+                FileItemStream item = it.next();
```
raphaelbauer: we usually use very loooong variable names. FileItemStream item should become FileItemStream fileStreamItem... FileItemIterator it = getFileItemIterator() should become FileItemIterator fileItemIterator = getFileItemIterator() and so on... tiny thing, but helps to keep the codebase readable and reasonable similar :)

raphaelbauer:
That looks very nice!

Obvious stuff missing:
1) Unit tests for the classes...
2) Documentation how stuff works
3) changelog.md ofc

Stuff that we still need to tackle:
I personally run Ninja on systems that do not allow to store uploaded files on the file system (no temp space). So I'd need an "in memory" mode to handle multipart requests. Can you think of a nice way how to disable the "automatically-store-multipart-files-on-temp-disk" thing?

Stuff you asked:
Regarding cleanup of temp files. finalize() sounds a bit too weak as we have no guarantee that finalize() is called at all. What about deleting the temp files once the result returns to ResultHandler? We could use a finally {} block to do that... We must then also add documentation - if someone wants to use the file after the controller method exists - she has to make sure that file got copied somewhere else...

Hope that helps :)

Your PR looks really nice. The uploading facilities were really not a nice part of Ninja. But you improved it very very nicely :) Good work!

bazi:
Updates:
- added full in-memory file upload handling (added `commons-upload` interface implementations), activated in application config
- clean up of temporary files (handled in `ResultHandler` as @raphaelbauer suggested)
- docs, tests, changelog, ...

P.S.
`Context.getFileItemIterator()` is still usable even if file item iterator is used to parse multipart request payload. Further calls are simulated by returning custom implementation of iterator.

raphaelbauer:
This looks really good - unfortunately we are quite busy right now - so it may take some more days to review and merge :)

raphaelbauer:
Looking now into the PR. Quite a lot of code... So reviewing may take a while :)

ninja-servlet/src/main/java/ninja/servlet/MultipartContextImpl.java
```diff
@@ -0,0 +1,290 @@
...
+                FileItemStream fileItemStream = fileItemIterator.next();
+                String fieldName = fileItemStream.getFieldName();
+
+                if (fileItemStream.isFormField()) {
+
+                    // simple key/value item
+                    StringBuilder sb = new StringBuilder();
+                    try (InputStreamReader isr = new InputStreamReader(fileItemStream.openStream())) {
+                        int n;
+                        char[] buf = new char[128];
```
raphaelbauer: can be replaced by a simple Streams.asString(item.openStream());

raphaelbauer:
before we continue a deeper discussion. Do you think it is possible to incorporate MultipartContextImpl into ContextImpl?

That you'd be the first thing I'd try to do - to remove a lot of code and also remove the bridge that checks whether to create a MultipartContextImpl or a ContextImpl.

Providing the InputStream feels handy to me... :)

bazi:
multipart related code moved to ContextImpl

raphaelbauer:
cool... :)
More comments this afternoon :) Thanks for the great work! Hope it's not too annoying for you - major features like yours are great - and need to be reviewed properly :)

bazi:
No problems! It is pleasure to contribute to awesome framework.

ninja-servlet/src/main/java/ninja/servlet/file/NinjaFileItemStreamConverter.java
```diff
@@ -0,0 +1,91 @@
...
+
+        uploadDirectory = new File(dir);
+
+        // check if specified target exists. create if does not exist, otherwise check if it is a directory
+        if (uploadDirectory.exists()) {
+            if (!uploadDirectory.isDirectory()) {
+                throw new ProvisionException("Specified target for upload files in not a directory");
+            }
+        } else {
+            uploadDirectory.mkdirs();
```
raphaelbauer: directory will be created even if I set inMemory to true...

ninja-servlet/src/main/java/ninja/servlet/file/NinjaDiskFileItemStream.java
```diff
@@ -0,0 +1,89 @@
...
+    private File file;
+
+
+     public NinjaDiskFileItemStream(FileItemStream fileItemStream, File directory) {
+
+        this.fileItemStream = fileItemStream;
+
+        try (InputStream is = fileItemStream.openStream()) {
+
+            this.file = File.createTempFile("ninja-upload-", null, directory);
```
raphaelbauer, line 63: This means we can only create a single uploaded file per Ninja instance... That sounds wrong... Did I miss something?

raphaelbauer, line 63: forget my comment... createTempFile takes care :)

ninja-servlet/src/main/java/ninja/servlet/file/NinjaFileItemStreamConverter.java
```diff
@@ -0,0 +1,91 @@
...
+ * This is a factory class that produces {@link FileItemStream} instances that
+ * can be stored and used repeatedly as opposed to instances returned by file
+ * item iterator whose data can be read only once.
+ *
+ */
+@Singleton
+public class NinjaFileItemStreamConverter {
+
+    @Inject
+    InMemoryFileItemFactory inMemoryFileItemFactory;
```
raphaelbauer: That belongs as @Inject into NinjaInMemoryFileItemStream

ninja-servlet/src/main/java/ninja/servlet/file/NinjaFileItemStreamConverter.java
```diff
@@ -0,0 +1,91 @@
...
+     * streaming API.
+     *
+     * @param fileItemStream {@link FileItemStream} instance returned by file
+     * item iterator
+     * @return {@link NinjaFileItemStream} instances that can be stored for
+     * repeated use
+     */
+    public NinjaFileItemStream convert(FileItemStream fileItemStream) {
+        if (inMemory) {
+            return new NinjaInMemoryFileItemStream(fileItemStream,
```
raphaelbauer: Please replace the "new" calls (both NinjaInMemoryFileItemStream and NinjaDiskFileItemStream) with a Provider...
you can simply inject a Provider<NinjaInMemoryFileItemStream> on top of this file and you are ready to go :) I'd create an init method that then takes "fileItemStream" or "uploadDirectory" in case of NinjaDiskFileItemStream.

ninja-servlet/src/main/java/ninja/servlet/file/NinjaFileItemStream.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ninja.servlet.file;
+
+import org.apache.commons.fileupload.FileItemStream;
+
+public interface NinjaFileItemStream extends FileItemStream {
```
raphaelbauer: Is it really necessary to extend FileItemStream? Wouldn't it be sufficient to just add outputStream (and maybe some other methods) to the interface?

ninja-core/src/main/java/ninja/Context.java
```diff
@@ -523,6 +526,36 @@ private HTTP_STATUS(int code) {
...
     /**
+     * Gets the uploaded file stream for the given key from the multipart
+     * request. When multiple files uploaded for the same key, the first file is
+     * returned.
+     *
+     * @param name the key of the file
+     * @return the input stream to read file contents from; {@code null} if no
+     * file uploaded for the given key
+     */
+    InputStream getUploadedFileStream(String name);
```
raphaelbauer: I am thinking about not returning an InputStream, but NinjaFileItemStream. We could return a proper NinjaFileItemStream (a but stripped down so that it does not contain all methods from FileItemStream). If a user wants to copy a file to a proper location it is easy to do so, as well as getting the content type and then it is also clea that the outputstream has to be closed, too...

What do you think about that?

ninja-core/src/main/java/ninja/Context.java
```diff
@@ -523,6 +526,36 @@ private HTTP_STATUS(int code) {
     FileItemIterator getFileItemIterator();
```
raphaelbauer, line 526: Should we mark getFileItemIterator deprecated?

raphaelbauer:
Hey @bazi , I added more comments and questions. The PR now looks better and better. I think one or two more iterations and we can merge your feature :)

momiji:
Hello,

I know I'm comming quite late on this one, but we're just implementing upload in our application, and I've been taking some time to think about this PR.
I'll really loved to have all parameters available, through @Param or context.getParameter(), and I'dl love to have this for files also.

From a user point of view, is it possible to define files like other parameters ?
Example: void myRroute(@Param("filename") String filename, @Param("file") File file);
Example: context.getParameterAsFile("file");

To distinguish between File and InputStream, just replace the File type with InputStream ?
Example: void myRroute(@Param("filename") String filename, @Param("file") InputStream file);
Example: context.getParameterAsInputStream("file");

To make all this "generic", I thought of a sort of "File" provider.
Something like a NinjaDiskFileProvider, to handle File or InputStream with some properties like :
- upload folder
- max allowed file size (we can use ByteStreams.limit(max+1) for checking this, and when file size reaches max+1, we're over)
- and so on for whatever is necessary

Something like a NinjaMemoryFileProvider, to handle InputStream (backed up in a ByteArrayStream), that can return only InputStream with some properties like :
- max allowed file size (we can use ByteStreams.limit(max+1) for checking this, and when file size reaches max+1, we're over)
  (we can use the same property names of course)

I was thinking of something like a provider to a NinjaFileParam object implementing :
- setInputStream(InputStream inputStream) to set input stream
- <T> getParamAs(Class<T> classType) to return the type File or InputStream or whatever else the user may want if creating his own provider
  Question : should we have generic here, or Object getParamAs(Class<?> classType) ?

Another idea was to be able to choose the provider at any time, with annotation on controller's class or method, allowing to have different configurations at different places in the application, like the max file size, or upload directory, if this has a sense for anyone.
Example: 
@FileProvider(NinjaDiskFileProvider.class)
void myRroute(@Param("filename") String filename, @Param("file") InputStream file);

To Resume:
- annotation to select provider : @FileProvider
- provider for disk files : NinjaDiskFileProvider, returns NinjaDiskFileParam that implements NinjaFileParam
- provider for in-memory files : NinjaMemoryFileProvider, returns NinjaMemoryFileParam that implements NinjaFileParam
- ninja properties should be read by the provider and set into the returned NinjaFileParam implementation
- NinjaDiskFileParam save the input stream into a disk file
- NinjaMemoryFileParam save the input stream into a ByteArrayStream
- extend ContextImpl to use all this :), 

NinjaFileParam fileParam = provider.get();
fileParam.setInputStream( ... );
Object paramValue = fileParam.getAs(param.getClass())
...

Finally, do we want to remove existing file code to handle files, or do we keep both methods ? if so, how to tell the ContextImpl to use one way or the another ?
Is it possible to have a default NinjaNoFileProvider to keep old behavior ?
if not, which one to use as default : disk or memory ?

raphaelbauer:
Very nice ideas!
Hmm - any idea how you'd make @FileProvider(NinjaDiskFileProvider.class) work? I don't have a clue right now...

I personally'd like to merge @bazi 's PR first. The PR is cool and facilitates a lot of stuff already. You could then base your work on bazi's simplifications and extend and incorporate your ideas... What do you think?

If there are urgent things it would be best if you could provide code or raise your hands... :)

momiji:
Well, if I was thinking of changing paramextractor to handle file and inputstream, by simply calling getParametetAsFile or AsInputStream.
Change contextimpl to do all form field reading at first call to getparameter(s ?), and as we have the route on context, i propose to look at thé route méthode and class to see if they have thé annotation telling thé provider to use, otherwise default to NoFileProvider, unless à default one has been injected at startup time (in a module I mean).
But maybe am I missing important things, I haven't taken time to look at this yet.
Do think it is feasable tjis way ?

momiji:
well it is, it's quite finished in fact. I didn't do the doc yet - but my englush is awfull, isn't it !
Minor refactoring is needed, some inspired from @bazi for unit tests. looks good and totaly integrated in existing context. I still need to write the disk file provider, but memory is finished.

momiji:
See #383 for full implementation... And big thanks to bazi who largely helped me discovered unit tests !!!

bazi:
@momiji good job! We targeted the same feature that Ninja lacks for now ;-)
I also liked your _file provider injection_ idea, this PR has much simpler approach -- you set it in app config and it stays unchanged.

I will update this PR soon as per @raphaelbauer remarks. I think I should anyways hit the end of this PR.
It is the beauty of a community driven open-source project where issues are identified and attacked by many different sides!

momiji:
Well I took some of your ideas and merge them with my sort of spec... And finally took the time to do the full implementation.

My only regret is that I should have made my remarks earlier, my purpose being KISS principle, that would have prevented me from doing all the job.

You've done a very good job, and creating tests is really annoying me, thanks a lot for leading the path...

I'll wait for your final PR to see if we are missing something.

Note: I believe you didn't handle multiple files with the same name, do you ?

raphaelbauer:
Ok!

That conversation sounds great :) How should we proceed now? First merge this PR and then momiji's?

I actually don't want to interfere much - it would be great if you could agree on a "final" solution that we can mere (I'll have a look at it nevertheless to see that it conforms to the other Ninja standards...).

momiji:
Well i'll let bazi answer.

By thé way when is next release expected, I hace à few more PR suite finished for exceptions page override, simplify URL, and global filters with assets versionning as a first example of use... Global filters is hard to implement, I did'nt expected that !

bazi:
Hey, PR is updated according to feedback from @raphaelbauer 

@momiji I am not sure how it would be best to proceed... One way is to merge this PR and then yours with all your nice features. If your PR is ready right away after @raphaelbauer's review, we can just merge your PR.
What do you think @momiji, @raphaelbauer ?

P.S.  @momiji about multiple files with the same name - this PR handled files with same names

raphaelbauer:
I don't care. We can merge this PR first and then @momiji can submit the improved version.
@momiji Ninja is released 12 times a year roughly.

@bazi I got some more remarks (tiny stuff I did not say until the big things were ironed out...) I'll add that now... Once that's incorporated we can merge it into master... :)

ninja-core/src/main/java/ninja/Context.java
```diff
@@ -41,6 +41,8 @@
     interface Impl extends Context {
         
         void setRoute(Route route);
+
+        void cleanup();
```
raphaelbauer, line 45: maybe we can rename that to cleanupAfterRequest() + add one line of javadoc...

ninja-core/src/main/java/ninja/params/ArgumentExtractors.java
```diff
@@ -187,6 +188,59 @@ public String getFieldName() {
         }
     }
 
+    public static class FileExtractor implements ArgumentExtractor<NinjaFileItemStream> {
```
raphaelbauer, line 191: can you rename the class to NinjaFileItemStreamExtractor?

ninja-core/src/main/java/ninja/params/ArgumentExtractors.java
```diff
@@ -187,6 +188,59 @@ public String getFieldName() {
...
+            return NinjaFileItemStream.class;
+        }
+
+        @Override
+        public String getFieldName() {
+            return name;
+        }
+    }
+
+    public static class FilesExtractor implements ArgumentExtractor<NinjaFileItemStream[]> {
```
raphaelbauer, line 215: Can you rename the class to NinjaFileItemStreamsExtractor?

ninja-core/src/main/java/ninja/params/FileParam.java
```diff
@@ -0,0 +1,38 @@
...
+ * Inject an uploaded file to a controller method.
+ *
+ * This equals context.getUploadedFile(...)
+ *
+ * @author azilet
+ */
+@WithArgumentExtractor(ArgumentExtractors.FileExtractor.class)
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.PARAMETER})
+public @interface FileParam {
```
raphaelbauer, line 34: Can you rename that to NinjaFileItemStreamParam?

ninja-core/src/main/java/ninja/params/FileParams.java
```diff
@@ -0,0 +1,38 @@
...
+ * Inject multiple uploaded files with same key to a controller method.
+ *
+ * This equals context.getUploadedFiles(...)
+ *
+ * @author azilet
+ */
+@WithArgumentExtractor(ArgumentExtractors.FilesExtractor.class)
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.PARAMETER})
+public @interface FileParams {
```
raphaelbauer, line 34: Rename to NinjaFileItemStreamsParam?

ninja-core/src/main/java/ninja/utils/NinjaConstant.java
```diff
@@ -143,7 +143,35 @@
...
+     * {@code false}.
+     */
+    final String FILE_UPLOADS_IN_MEMORY = "file.uploads.in_memory";
+
+    /**
+     * The maximum allowed size of a single uploaded file.
+     *
+     * @see org.apache.commons.fileupload.FileUploadBase#fileSizeMax
+     */
+    final String FILE_UPLOADS_MAX_FILE_SIZE = "file.uploads.file.size.max";
```
raphaelbauer, line 159: Do you think we could simplify that and only use "FILE_UPLOADS_MAX_REQUEST_SIZE". It somewhat also includes the max files size and we got one property less :)

ninja-core/src/site/markdown/documentation/basic_concepts/controllers.md
```diff
@@ -130,6 +130,9 @@ accepted by Joda Time library.
 Therefore you don't have to worry if
 input is for instance Xml or Json. You simply get a parsed object.
 
+Additioanlly, uploaded file streams of a multipart request can be injected into your controller.
```
raphaelbauer, line 133: Additionally

ninja-core/src/site/markdown/documentation/uploading_files.md
```diff
@@ -62,6 +62,60 @@ public Result uploadFinish(Context context) throws Exception {
 }
 </pre>
 
+New way of handling file uploads
+--------------------------------
+Would it be nice if uploaded files of a multipart request could be injected
```
raphaelbauer, line 67: Wouldn't it be nice...

ninja-core/src/site/markdown/documentation/uploading_files.md
```diff
@@ -62,6 +62,60 @@ public Result uploadFinish(Context context) throws Exception {
...
+directly into controller method just like path or query parameters?
```
raphaelbauer, line 68: method"s"

ninja-core/src/site/markdown/documentation/uploading_files.md
```diff
@@ -62,6 +62,60 @@ public Result uploadFinish(Context context) throws Exception {
...
+Ninja allows you to do so -- to inject uploaded file `@FileParam` annotation
```
raphaelbauer, line 69: Ninja allows you to do so using the @Ninja.... annotation.

ninja-core/src/site/markdown/documentation/uploading_files.md
```diff
@@ -62,6 +62,60 @@ public Result uploadFinish(Context context) throws Exception {
...
+is used. Extracted type of this annotation is `NinjaFileItemStream` from which
+you can get `java.io.InputStream` to read file contents from. There is also a
```
raphaelbauer, line 71: Has to be updated to reflect the new NinjaFileItemStream annotations and usage...

ninja-servlet/src/main/java/ninja/servlet/ContextImpl.java
```diff
@@ -134,6 +151,16 @@ public void setRoute(Route route) {
     }
 
     @Override
+    public void cleanup() {
+        // no op for non-multipart request context
+        for (List<NinjaFileItemStream> items : fileItems.values()) {
```
raphaelbauer, line 156: please remove thespacing between : and items (we want to keep the formatting uniform... :) )

ninja-servlet/src/main/java/ninja/servlet/ContextImpl.java
```diff
@@ -134,6 +151,16 @@ public void setRoute(Route route) {
...
+            for (NinjaFileItemStream item : items) {
```
raphaelbauer, line 157: also :

ninja-servlet/src/main/java/ninja/servlet/ContextImpl.java
```diff
@@ -221,7 +260,22 @@ public Integer getParameterAsInteger(String key, Integer defaultValue) {
 
     @Override
     public Map<String, String[]> getParameters() {
-        return httpServletRequest.getParameterMap();
+        // create new map with query params
+        Map<String, String[]> map = new HashMap<>(httpServletRequest.getParameterMap());
+
+        // add params of multipart request
+        for (Entry<String, List<String>> e : multipartParams.entrySet()) {
```
raphaelbauer, line 267: also please remove the space between e and : => please do that everywhere in the PR... Nothing personal.. it's just how the other code of Ninja is formatted...

ninja-servlet/src/main/java/ninja/servlet/file/InMemoryFileItemFactory.java
```diff
@@ -0,0 +1,63 @@
...
+ * {@link FileItemFactory} implementation that produces in-memory stored file
+ * item instances.
+ */
+@Singleton
+public class InMemoryFileItemFactory implements FileItemFactory {
+
+    /**
+     * Default max file size of 10 MB.
+     */
+    public static final int DEFAULT_MAX_FILE_SIZE = (1 << 20) * 10;
```
raphaelbauer, line 38: can you crete another static final variable called TEN_MB and assign it to (1 << 20) \* 10; Then you can remove the comment :)

ninja-servlet/src/main/java/ninja/servlet/file/NinjaFileItemStreamWrapper.java
```diff
@@ -0,0 +1,96 @@
...
+        this.value = value;
+    }
+
+    public void setItemStream(NinjaFileItemStream itemStream) {
+        this.itemStream = itemStream;
+    }
+
+    @Override
+    public InputStream openStream() throws IOException {
+        if (itemStream != null)
```
raphaelbauer, line 60: Please use { and } - that's just the coding convention we use... :D

raphaelbauer:
Okay finished adding my smaller remarks. A lot of formatting and language things :) Don't take it personally - it's really just to make things better and keep things uniform across the codebase...

raphaelbauer:
@bazi Let us know when you are finished - then we can merge the PR :)

momiji:
I doubt you can merge my PR after, as it use a very different approach than this one. It will require to remove too many things from this PR, and should be considered as a drop-in replacement.

Making my own PR was more a challenge to see if the "file provider" was something that could be done, and I found that it was quite easy to implement. It also make me learn things about Github branching, Mockito tests, Guice injection, and so on.

The important thing is to have a way to handle files, whatever way is used to do it, and to do it in a way that let's the final developers do their job, with enough flexibility.

I would have only one request before merging this PR, please be sure you do not break getFileItemIterator() current behavior. It may already be used in some applications, and shod still work as before, meaning providing original streams to the developer.

I'm just saying that because we are already using it in several 3 different ways:
- save file to disk
- save file to disk with on-the-fly encryption
- save file chunks to disk (for very big files) with on-the-fly encryption

I doubt to be able to do all that with the new way to handle uploads, as we have both in-memory and disk, but this is not a problem as long as we can still use the getFileItemIterator(), as actually done.

If not, then this should be mentioned that this is a breaking change, and we will switch to use the internal httpServletRequest and ServletFileUpload, as it was originally coded.

raphaelbauer:
Ok - how should we proceed? I don't care how it is implemented as the way it is currently is very very low level (but works of course). Any improvement would be great.

The only constraint I have: I want to be able to control whether the file is stored in memory or on disk.

Anything else can be discussed :)

I guess it is about you two: @bazi  @momiji  - you did the great job of actually implementing great solutions. Can you agree on one? Or agree on how we want to proceed from those two PRs? Thanks :)

bazi:
@momiji `getFileItemIterator()` is marked deprecated but it surely keeps its behavior.

Anyways, i made a decision. Let's abandon this PR and focus on the other one from @momiji, guess it is #383 but it seems closed. Please provide PR, I will be on the topic.

@raphaelbauer beg your pardon for your time spent on reviewing ;-)

raphaelbauer:
Absolutely no problem! It's a pleasure to read code and improve things together! :)
Let me know when I can help on the "other" PR... As I said above - it would be great if you two could team up to implement the feature!

Cool effort!

I'll close this PR then...

PR #370, opened 2015-06-22
fix directory traversal vulnerability

noguespi:
Fix AssetsController vulnerability, without removing static files feature.

Now when the controller server a static files, it always check the files is in the assetBaseDir.


noguespi:
forgot one fix

PR #371, opened 2015-06-22
Asset fix

noguespi:
Fix AssetsController vulnerability, without removing static files feature.

Now when the controller serves a static files, it always check the files is in the assetBaseDir


liudianpeng:
what???????????

noguespi:
@liudianpeng related to https://groups.google.com/forum/#!topic/ninja-framework/Go97aKNWoCw

liudianpeng:
@serphacker I would like to consult with the redirectt Ninja

liudianpeng:
I would like to consult with the redirect Ninja

PR #372, opened 2015-06-24
Flyway upgrade from 2.3.1 to 3.2.1 (fix many problems)

danielsawan:
Hi there. This is my very first pull request ever so i am still discovering how all this work.
I had many problem using Flyway with ninja but not any problem using flyway standalone client. 

After some research I saw that the Flyway version used in Ninja was very old 2012/2013 so I decided to update it to the last version (2015).

After this update I had no more problem using flyway in Ninja !

PS : in only tested the basic usage (mine). This pull request don’t cover all the required tests. (at least I guess).

### DIFF
--- a/ninja-core/pom.xml
+++ b/ninja-core/pom.xml
@@ -216,7 +216,7 @@
 
         <!-- Migration framework -->
         <dependency>
-            <groupId>com.googlecode.flyway</groupId>
+            <groupId>org.flywaydb</groupId>
             <artifactId>flyway-core</artifactId>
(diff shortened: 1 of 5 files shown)


raphaelbauer:
Nice! Could you update changelog.md to reflect the change? Thanks!

liudianpeng:
Java linkage how to write

liudianpeng:
@raphaelbauer

raphaelbauer:
Nice work!

PR #373, opened 2015-06-25
Enforce unix separator for loading webjars

BjoernAkAManf:
Fixes loading of webjar assets using windows (introduced with commit 259a4e524d845aa7e1dc3ec9333d5d7c61b25eb1 ). The given filename will turn /foo/bar to foo\bar. Turning the path into "META-INF/resources/webjars/foo\bar", which will not work.

This commit aims to remain compatible with any tests or subclasses, that could possibly exist. Works like a charm for debian and windows 7 alike (as far as i can tell with my tests).

### DIFF
--- a/ninja-core/pom.xml
+++ b/ninja-core/pom.xml
@@ -111,7 +111,7 @@
         </dependency>
         
         <!-- Afterburner lib optimizes Jackson and improves
-             performance -->
+        performance -->
         <dependency>
             <groupId>com.fasterxml.jackson.module</groupId>
             <artifactId>jackson-module-afterburner</artifactId>
@@ -251,6 +251,16 @@
             <groupId>com.icegreen</groupId>
             <artifactId>greenmail</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-module-junit4</artifactId>
+        </dependency>
+        
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-api-mockito</artifactId>
+        </dependency>  
 
     </dependencies>
 
--- a/ninja-core/src/main/java/ninja/AssetsController.java
+++ b/ninja-core/src/main/java/ninja/AssetsController.java
@@ -28,7 +28,6 @@ import java.net.URLConnection;
 
 import ninja.utils.HttpCacheToolkit;
 import ninja.utils.MimeTypes;
-import ninja.utils.NinjaConstant;
 import ninja.utils.NinjaProperties;
 import ninja.utils.ResponseStreams;
 
@@ -36,7 +35,6 @@ import org.apache.commons.io.FilenameUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.base.Optional;
 import com.google.common.io.ByteStreams;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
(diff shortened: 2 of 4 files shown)


raphaelbauer:
That's cool! We clearly lack a proper test here. Can you think of a way to test the functionality? Otherwise we may run into the same problems soon.

BjoernAkAManf:
The test are already runnin (under windows) and failing. I just fixed that, but im still working to make the tests work and fail equally under linux.

BjoernAkAManf:
I think this would resolve the testability for windows and unix systems. It's a bummer i had to add a new dependency, but i dont think its sane to introduce a new layer of complexity (inject FilenamesUtil) for the sake of testing only.

raphaelbauer:
Hmm... Powermock... hmm... No...
normalizePathWithoutLeadingSlash is already marked as @VisibleForTesting ... I'd just run the test against it...

Another stupid question... why don't we always just use FilenameUtils.normalize(fileName, true)? We always need unix separators - right?

we could then simplify everything to that:

```
    @VisibleForTesting
    protected String normalizePathWithoutLeadingSlash(String fileName) {
        String fileNameNormalized = FilenameUtils.normalize(fileName, true);
        return StringUtils.removeStart(fileNameNormalized, "/");
    }
```

would make the tests also a lot simpler... (just test normalizePathWithoutLeadingSlash with windows paths and check whether we get unix paths back...).

BjoernAkAManf:
Well, normalizePathWithoutLeadingSlash is used twice. Serving webassets and static files alike. In my lokal test this did not change anything. Windows may support using / instead of \ for File paths, but im not sure this will work interchangeably. In my local Version i'm using the above solution aswell in a subclass.  

This could break subimplementations tho. [Well it is marked as VisibleForTesting anyway]

raphaelbauer:
ah yea.. you are right... I missed that... maybe it's too late ;)

BjoernAkAManf:
I don't think ignoring this issue is a good alternative. It is annoying, to use Ninja in this environment for any windows based platform. [Atleast while debugging. I don't think production needs to support Windows anyway]

raphaelbauer:
The issue is no being ignored :)  I'll propose an alternative with all your changes but without Powermock soon. There is also another tiny issue that will be fixed by that... stay tuned :)

raphaelbauer:
Another thing - Ninja still is a community project, and while we try to cut one release per month - there is no guarantee. Good news is that if you got problems you can always use (at least as workaround) the version of your pull request in your dev's machines... Just call mvn install and you are ready to go... :)

raphaelbauer:
Would you mind having a look at: https://github.com/ninjaframework/ninja/pull/378 It's based on your PR. It just fixes one other issue and added some missing testcases. I kept Powermock. I distaste it really, but it's the only way we can mock the behaviour of Apache Commons in that scenario. You are added as contributor in the changelog. Feedback appreciated :)

BjoernAkAManf:
I did not like powermock either. I feelt uncomfortable using it, as it does introduce new dependencies and complexity for such "simple task". 

By the way: I was using my own workaround, before proposing my pull request already. It was easy to introduce a fix, but i was not sure i covered fulll interchangeability (thus, proposing this change instead). I really enjoy using this library. I always distasted php - especially in terms of performance and so called "features", that i would not expect. Yet i had to use it. I can't stand the usual way to deploy web applications in java. And the combination of guice and convention of configuration is well designed. 

Im currently just getting warm with the possibilities and read through the documentation. I will hopefully contribute more in the future, giving back something for the great projekt i can enjoy working with. Im now going to comment on #378, closing this issue as a result.

PR #376, opened 2015-06-28
Scheduler pool size can now be configured

danielsawan:
scheduler.pool.size=1 by default


raphaelbauer:
Cool!
To merge this we need at least some documentation for the feature. And also some unit tests that make sure that the configuration is read and used :)

PR #382, opened 2015-07-18
add new jvmArgs property, allows to add arguments to the embedded jetty

momiji:
this may replace my other PR as I now understand that I should not use "develop" branch...

### DIFF
--- a/ninja-core/src/site/markdown/developer/changelog.md
+++ b/ninja-core/src/site/markdown/developer/changelog.md
@@ -1,3 +1,8 @@
+Version x.x.x
+=============
+
+ * 2015-07-18 Added jvmArgs to maven plugin (momiji)
+
 Version 5.1.4
 =============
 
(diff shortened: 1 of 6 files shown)


momiji:

raphaelbauer:
Very well done! Good documentation.

Can you think of a test that makes sure we don't break the feature unintentionally? 
- Somehow it would be cool to make sure the array is populated properly on init
- And that splitUnescapedSpaces works properly (make it protected and write a simple test would be pragmatic eg).

momiji:
I'll have a look, but don't how to do the first one. I'll add some tests for the second one.

noguespi:
great PR, that's what I need

noguespi:
Can it be merged for next release or do we have to wait for #409 ?

raphaelbauer:
I'll merge this as part of #409 . You are mentioned as contributor... Great stuff :) (Declining the PR therefore)

PR #384, opened 2015-07-20
added simplification or request uri

momiji:
To prevent things like route path param like {filename} to contain path traversal like ../another_folder/another_file when / are encoded

### DIFF
--- a/ninja-core/src/main/java/ninja/utils/SwissKnife.java
+++ b/ninja-core/src/main/java/ninja/utils/SwissKnife.java
@@ -17,7 +17,10 @@
 package ninja.utils;
 
 import com.google.common.base.CaseFormat;
+import com.google.common.base.Joiner;
+import com.google.common.base.Splitter;
 import com.google.common.primitives.Primitives;
+
 import org.apache.commons.configuration.ConfigurationException;
 import org.apache.commons.configuration.PropertiesConfiguration;
 import org.joda.time.LocalDateTime;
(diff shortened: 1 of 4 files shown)


gitblit:
Did you consider using the Guava `Files.simplifyPath` method?  You'd still have to perform the decoding.

http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/io/Files.html#simplifyPath%28java.lang.String%29

momiji:
Yes. If you have a closer look, you'll see it's a bare copy of Files.simplifyPath with decoding in needed places, espacially for %2E.

momiji:
now all tests should pass... still learning how to switch from one branch to another

momiji:
OK this PR is finished.

momiji:

raphaelbauer:
I am not entirely sold. Not sure if we should do the simplification of every route request. Usually the browser of the user does it already, so relative paths in html pages are perfectly ok. Everything else sounds a lot like something tied to a narrow usecase (mapping urls to local filenames) and I'd say that this should be done by an external library. Another indication is that Jetty itself does not do it...
There is https://tools.ietf.org/html/rfc1808 which details this. 

@gitblit What is your opinion on this?

momiji:
Even if the way to solve the problem is wrong, I still believe there is a problem.

As said in the forum groups, it is quite unexpected for a route like "/path/{name}" to get a value of "name" containing a path traversal like "../../file". And the same for "/path/{name: .*}" that may contain a path traversal outside /path/.

That means that all routes containing variables are not to be trusted, and that all path parameters are to be checked. If you don't, there are so many ways to exploit this, that our security team considered this to be a big hole that must be corrected, something I won't ague as my tests proved they are correct.

Like you said, the problem may appear when using the "name" to serve files, but also when calling another REST service. It's only the two use case we discovered from our use, their are probably many more we didn't think of.

And by the way I'm not that sure Jetty does not do the same. When resolving servlets url, it seemed to me that before looking for the servlet to use, it was also doing some simplification, as seen in the jetty traces (look at ContextHandler). However, it does not send the simplified uri to the servlet, this I agree with you - it's use is just internal to jetty.

Maybe another solution can be to add a new section in the documentation, for hardening applications. It may contain a list of known flaws and a list of (non perfect) counter-measures or ideas that can be applied to solve them, as a guideline for all people that need to secure their application.

I was thinking of something like:
- change cookie prefix in production
- change secret key in production
- do not run as root in production
- add an external component to secure urls

For this case, maybe there are some reverse proxies like F5 that can check urls and automatically refuse them when they are not well formed. I don't have knowledge in this way. Maybe someone that helped building this framework knows about that. In that case, it could be a prerequisite for production.

momiji:
Well let's forget about this one, I feel I just overreact.

We'll still keep our NinjaDefault to correct this in our projects for the moment.

Sorry for taking your time...

raphaelbauer:
On the contrary - if you think it is important we should consider including it. You security team was right to nag: Trusting user input without sensible context specific escaping is very very dangerous. But that's nothing a general web framework can fix. It's like saying: We got an SQL injection because we trusted the web framework's query string escape mechanism. Doesn't work because a web framework doesn't know anything about SQL. That also applies for file path mapping from urls to file paths. Without context and knowledge about the used file systems there will be a lot of security holes (Setting correct access rights on the file system should be mandatory to mitigate effects).

Ok. Long story short: If you want this to be included I am interested in two things:
1) Why doesn't Jetty do that collapsing of urls => I'd try to ask that on the Jetty mailing list
2) Does Rails, Play etc do that? (I guess not) - if not - why?

If it's important for you it would be great if you could ask and find the answers :)

rfc1808 imho applies here, but not because of sec reasons, but because we want to get http right.

Thanks for bringing up the issue!

PR #385, opened 2015-07-20
add global filters that run before computing route

momiji:
To come soon : test case and first example of use I hope.

Please have a look now, as there is one breaking change : a new method in Ninja interface, could not avoid it. Any other idea ?

### DIFF
--- a/ninja-core/src/main/java/ninja/Context.java
+++ b/ninja-core/src/main/java/ninja/Context.java
@@ -682,4 +682,12 @@ public interface Context {
      * @param cookie Ninja Cookie
      */
     void unsetCookie(Cookie cookie);
+
+    /**
+     * Allows to change the requestPath.
+     * Do not change requestPath after context has been used heavily, there might some side effects.
+
+     * @param requestPath The new requestPath
+     */
+    void changeRequestPath(String requestPath);
 }
--- a/ninja-core/src/main/java/ninja/Ninja.java
+++ b/ninja-core/src/main/java/ninja/Ninja.java
@@ -16,6 +16,8 @@
 
 package ninja;
 
+import ninja.Context.Impl;
+
 public interface Ninja {
 
 	/**
@@ -23,6 +25,11 @@ public interface Ninja {
 	 */
 	void onRouteRequest(Context.Impl context);
     
+    /**
+     * When all global filters have finished and a route must be found to continue chaining.
+     */
+    Result findRouteAndChain(Context.Impl context);
+
     /**
      * This result should be used when an error occurs.
      * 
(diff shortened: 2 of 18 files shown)


momiji:
This PR is considered finished, it can be reviewed now...

momiji:
another stupid mistake, didn't mvn clean before mvn test so results were from previous compilation from other branch...

momiji:

raphaelbauer:
Very cool stuff! This PR is in the queue, but may not make it to this month's release depending on the workload :) Stay tuned...

danielsawan:
Nice PR !
By the way using the global filters or creating a global controller (potentialy annotated) which will be extented by all our controller would procude the same results right ? Your global filters is like a shortcut to this usecase with less granularity ?

momiji:
Well, in fact no, it is not quite the same, as the global filter offer one additional feature: it allows to manipulate the context before looking for routes.

Of course, for a security filter, it is quite the same as annotating all controllers or using the new filteredWith() method in route builder, as it allows to add filter to AssetsController without extending it.

But for the resource versioning filter (VersionCacheFilter), we just remove the version from the requestPath, making it totally transparent for the developers and the routes definition.

As this was asked a while ago in the forums, maybe are there other usage ideas.

raphaelbauer:
just a quick note. neglecting the version of an asset can be also easily achieved by using a route like 
router.GET().route("/assets/{._}/{fileName: ._}").with(AssetsController.class, "serveStatic");
which then maps to all sorts of things like /asset/1.0.0/main.css or /asset/1.0.1/main.css

jjlauer:
I do think global filters is a good idea, but I'm curious about the VersionCacheFilter.  Seems like many other frameworks use MD5 fingerprints per-asset.  The advantage is you could set a long cache time and also force browsers to pickup single assets that were modified.  What's the advantage of the VersionCacheFilter?  Did you consider a fingerprint/MD5 approach instead?

In templates, you could have something like ${fingerprintedAssetsAt("css/main.css")}. 

Thoughts?

momiji:
@raphaelbauer You're right, it was the idea at first. However, the HttpCacheToolkit code update purpose is to ensure that if the developer forgets about the version, the asset (and all items using HttpCacheToolkit) will not be cached. 

By the way, The version we're using has evolved a little, and i'm no more certain it is a good idea to have it integrated like this. it's missing a force no cache feature we have added lately, so the best would be to ignore it, and we'll keep the way we're actually doing : a filter + another HttpCacheTookit replacement binded in a module.

@jjlauer Sorry for my late answer, when I had a look on versioned assets in play, it seemed to me it was using something like a timestamp in the query part of the url. Perhaps a timestamp based on the resource last modification date, but i'm not sure.

Computing md5 may seem better, but in our use case we have a bunch of js coming from an external library, with a lot of relative references, and it seemed easier to just add a tag in front of all urls. This way, we're sure clients will update at each new deployment, and we don't have to update the javascript files, they just work as is because they use relative paths.

However, I'm sure there might be cases where fingerprint could be better, or maybe other solutions, depending on the needs of course.

PR #387, opened 2015-07-22
Add support for Jackson JSON Views (@JsonView)

metacity:
I also considered a controller method annotation based approach (like in Spring), but the implementation would've been more cumbersome and invasive. JSON view being set in the Result object is also imo more explicit that the view will be used only for serialization. 

Thoughts?


raphaelbauer:
That's perfect! Great job!

PR #397, opened 2015-08-17
RollingFile appenders to archetypes' logback.xml & minor doc fixes

metacity:
Most people will of course just Google for the logback snippet, but I think it's always a nice surpise to find an included example :+1:


raphaelbauer:
Nice work!

PR #399, opened 2015-08-23
Add charset to error html

mallowlabs:
I found that my localized exception messages are garbled in diagnostic page.
I guess that charset should be specified in these files.


raphaelbauer:
nice!

mallowlabs:
Thanks!

PR #401, opened 2015-09-15
Improved session docs & fixed minor bug with diagnostic mode

jjlauer:
- Fixed status code for forbidden results in diagnostic mode
- Added documentation of configuration properties for session cookies

### DIFF
--- a/ninja-core/src/main/java/ninja/NinjaDefault.java
+++ b/ninja-core/src/main/java/ninja/NinjaDefault.java
@@ -176,7 +176,6 @@ public class NinjaDefault implements Ninja {
         
     }
     
-    // TO-BE-EVENTUALLY-ADDED-TO-NINJA-DEFAULT-IF-FOLKS-LIKE-THIS
     public Result getRenderingExceptionResult(Context context, RenderingException exception) {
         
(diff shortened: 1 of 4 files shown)


metacity:
Nice! :+1: 

One thing I actually just noticed and tested, and because imo it's more like a "feature-needs-doc" and session cookie documentation is now on the table: 

When setting the `application.session.expire_time_in_seconds` property with `application.session.send_only_if_changed` being `true`, the expiration seconds is no longer actually "time of inactivity", but a hard expiration time after the last Set-Cookie is sent to the client. So with e.g. `expire_time_in_seconds` set to 60 and `send_only_if_changed` set to true, a user's logged-in session cookie will simply expire (unless session values are modified within the expiry time frame) in 60 seconds and the user will be logged out (assuming an auth filter) - no matter how many page loads he does. When `send_only_if_changed` is false, the session cookie and its expiration time is naturally refreshed on every HTTP response. 

As Ninja uses client-side session implementation, I don't see this as a bug, but worth documenting.

jjlauer:
@metacity good point -i added some more info about this behavior

raphaelbauer:
That's pretty and perfect! Very nice! Thanks :+1:

PR #404, opened 2015-09-29
Better diagnostic exceptions; improved exception logging

jjlauer:
Refactored diagnostic mode to include info on the context, request, and response at the time of the exception.  The diagnostic error page now has tabs to toggle between the various views.  This also provides a UI design to potentially plugin additional diagnostics down the road (e.g. custom application tabs).

In order to capture the "underlying" Result at the time the exception was triggered, I had to refactor NinjaDefault slightly to pass the underlying result along.  It seems like that only makes sense in the case of a RenderingException or InternalServerErrror.

I also now log exceptions handled in NinjaDefault.onException at the "debug" level.  There was a mailing list question about why JSON parsing errors were silent and they were not currently logged.

ninjaimproveddiagnostic1

ninjaimproveddiagnostic


raphaelbauer:
wow! looks beautiful... plz give me some days to review it :)

ninja-core/src/main/java/ninja/Ninja.java
```diff
@@ -28,6 +28,9 @@
      * 
      * @param context The context for this request
      * @param exception The exception to handle. Can be used to customize error message.
+     * @param underlyingResult The underlying Result if the handling of the
```
raphaelbauer, line 31: that doesn't seem to be there..

raphaelbauer:
Well done! That's incredibly useful!

PR #406, opened 2015-10-01
Fix garbled snippets in diagnostics

mallowlabs:
Snippets are garbled in diagnostic page.
My source codes are written in UTF-8.

I think that `InputStreamReader` should have UTF-8 charset.


jjlauer:
@raphaelbauer The looks good to me.  Nice catch @mallowlabs

raphaelbauer:
Nice catch - very well done - and very cool you provided a testcase!

mallowlabs:
Thank you!!

PR #409, opened 2015-10-07
Refactored servlet and standalone to prepare for eventual non-servlet-based Ninjas

jjlauer:
Fairly large commit to prepare for non-servlet-based contexts, bootstraps, and standalones.  Should finally remove the overall reliance on servlets to Bootstrap a Ninja application.  Here are the main changes...

_ninja-servlet/src/main/java/ninja/servlet/NinjaBootstrap.java_
- Non-servlet functionality extracted to `ninja-core/src/main/java/ninja/Bootstrap.java` and is now part of ninja-core.
- Servlet functionality extracted to `ninja-servlet/src/main/java/ninja/servlet/NinjaServletBootstrap.java`

_ninja-servlet/src/main/java/ninja/servlet/ContextImpl.java_
- Non-servlet functionality extracted to `ninja-core/src/main/java/ninja/utils/AbstractContext.java` and is now part of ninja-core.
- Servlet functionality extracted to ninja-servlet/src/main/java/ninja/servlet/NinjaServletContext.java

_ninja-standalone/src/main/java/ninja/standalone/NinjaJetty.java_
- Non-jetty functionality extracted to `ninja-core/src/main/java/ninja/standalone/AbstractStandalone.java` and is now part of ninja-core.
- Jetty functionality retained in `ninja-standalone/src/main/java/ninja/standalone/NinjaJetty.java`
- New `ninja-core/src/main/java/ninja/standalone/Standalone.java` interface that any standalone is required to implement.

_Other notable enhancements_
- Any Ninja standalone configuration (e.g. ninja.port) now available in `conf/application.conf`.  The priority of settings is systemProperty then configurationProperty then defaultValue.
- New 'ninja.standalone' system property available to change the standalone implementation to be used for NinjaTestServer or maven plugin.  I have been working on an Undertow implementation for Ninja standalone and this let's me pick what Ninja uses in key spots.  It defaults to NinjaJetty.

_Where are the docs & changelog?_

Let me know what you think of the PR first and I can add the final doc and changelog info.


raphaelbauer:
Looks cool - will have a deeper look asap :)

jjlauer:
@raphaelbauer Updated the changelog with the large amount of changes made.  I took the liberty of bumping the version to 5.2.0.  I also finally figured out how to properly test the maven plugin in a modern way.  I took the liberty to finally add support for `-Dninja.jvmArgs` #382 -- but I used an external parsing library from DrJava to handle whitespace rather than rolling our own.  I think there is finally a unit test for the maven plugin to increase your confidence in accepting PRs for it.

To help explain why I made all these changes, I have Undertow support working in another branch https://github.com/fizzed/ninja/tree/feature_undertow -- which I'd like to also PR down the road.  It doesn't use servlets and I didn't want to copy code to make things like a Context to work.

momiji:
I've made other changes to the maven plugin in our project's repo, as it took to much time ot have it merged.
I'll try to push them to my public repo in the next weeks, so you can merge them into your PR. The changes are for allowing multi module project, but also for using jrebel.

jjlauer:
Hey @momiji - I worked on a number of multi-module enhancements to the maven plugin 3-4 months ago that were merged in around 5.1.0-timeframe.  I'm curious what other multi-module features you added?  Would love to take a look at your public fork when you publish it.

momiji:
Just pushed in my old PR.

The main things are:
- / as path in includes and excludes, allows to define all paths with / whether on linux or windows - by doing this, allows to use the same pom in windows and linux, without having to add all includes/excludes path with / and with \ (no more duplicate paths)
- watchDirs now replace default watchDir when provided, allows to remove default watchDir when using jrebel for example
- jvmArgs is now a List<String>, as it should be, better when defined in pom. However, still need a way to define from command line, and as "," is not escapable in maven command line (as far as i know), we need to do it by ourselves, this is the purpose of jvmArgsStr that override jvmArgs if both defined
- remove all .ftl as we don't use freemarker, just considering the maven plugin should be agnostic to freemarker and handle all .html (ftl, rtm, ...)

For the use case, we have 5 web apps with a common lib, all in same parent folder, and of course a parent pom defining all modules - very classic project configuration.

Default pom configuration for all web apps is:

```
                <configuration>
                    <watchDirs>
                        <watchDir>target/classes</watchDir>
                        <watchDir>local</watchDir>
                        <watchDir>../project-lib-common/target/classes</watchDir>
                    </watchDirs>
                    <excludes>
                        <exclude>(.*)/views/(.*)\.html</exclude>
                        <exclude>(.*)/assets/(.*)</exclude>
                    </excludes>
                    <classpaths>
                       <classpath>../project-lib-common/target/classes</classpath>
                    </classpaths>
                </configuration>
```

The classpath is here to use target/classes of the common lib instead of installed .jar - as eclipse recompiles in target/classe, no more need to install common lib. Combined with the last watchDir to automatically restart if common lib is changed.

When started with ninja:run, we use `-Dninja.jvmArgsStr="-Dninja.external.configuration=local/dev.conf %JAVA_DEBUG%"`, `JAVA_DEBUG` empty unless we want debug, in that case: `JAVA_DEBUG=%JAVA_DEBUG% -Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=8000`

When started with jrebel, we don't want other watchDirs than local (as they are handled automatically by jrebel with adequat configuration) so we just override watchDirs with `-Dninja.watchDirs=local` to the command line, in addition to jrebel specific arguments.

Hope this is clear enough, let me know if you want more details.

I'm not sure everything we've done were the good things to do, but well it works and fits our needs, and we now have a very fast way to develop, and depending on the type of work (when not chaging routes for example), we can use jrebel to go even faster.

momiji:
Just discovered #274, (MavenSession), maybe it can be used to automatically discover watchDirs. If so, then I only miss "local". If it works and can be done this way, maybe change this to a "useDefaultWatchDirs" + "watchDirs without override", it will still fit my need. What do you think ?

jjlauer:
@raphaelbauer Curious if you have any thoughts yet on the PR?

gitblit:
I've been running on Undertow for several projects and I highly recommend adopting it over Jetty - with or without the servlet layer.  @jjlauer do you have a benchmark comparison of Ninja+Servlet+Undertow vs Ninja+Undertow? Or are you making the reasonable assumption that fewer layers = better performance?

jjlauer:
No official performance results yet on Ninja+Undertow, but
'ninja-servlet-integration-test' already does run a bit faster with
NinjaUndertow vs. NinjaJetty.  I initially started working on this,
however, to simply take advantage of all the other things Undertow is
pretty slick at.  For example, all the other handlers you can chain
together to build your own customized microservice and not be required to
put something like Nginx in front of it.

There are also other optimizations it could eventually enable -- e.g.
optimizing the double buffering of templates, direct transfer of assets,
websockets, etc.

jjlauer:
@momiji This discussion is probably getting out of scope pertaining to this PR.  However, wanted to reply to some of your questions.  I  think a lot of what you want can already be accomplished with the existing maven plugin:

_/ as path in includes and excludes_

Both includes and excludes accept a regex, so a pattern would match either.  For example:

``` xml
<excludes>
   <exclude>(.*)[/\\\\]views[/\\\\](.*)\.html</exclude>
   <exclude>(.*)[/\\\\]assets[/\\\\](.*)</exclude>
</excludes>
```

If you're using a lot of them then set `[/\\\\]` to a property and maven would swap in that string in place.

``` xml
<properties>
  <dir.regex>[/\\\\]</dir.regex>
</properties>

<excludes>
   <exclude>(.*)${dir.regex}views${dir.regex}(.*)\.html</exclude>
   <exclude>(.*)${dir.regex}assets${dir.regex}(.*)</exclude>
</excludes>
```

_watchDirs now replace default watchDir when provided, allows to remove default watchDir when using jrebel for example_

By default the plugin adds the `buildOutputDirectory` to watchDirs.  If you don't want "target/classes" added by default then couldn't you set buildOutputDirectory to a different value?  That would then be added by default first.

_jvmArgs is now a List, as it should be, better when defined in pom. However, still need a way to define from command line, and as "," is not escapable in maven command line (as far as i know), we need to do it by ourselves, this is the purpose of jvmArgsStr that override jvmArgs if both defined_

While I agree parsing the value is more difficult, accepting a line is easier to use.  You can easily support it from the command-line and adding an array of values for a ton of arguments is pretty tough.  The tokenizer library I used would handle commas in your list.

_remove all .ftl as we don't use freemarker, just considering the maven plugin should be agnostic to freemarker and handle all .html (ftl, rtm, ...)_

You could set `useDefaultExcludes` to false to not have that regex added by default.  However, even if you didn't and you don't use fremmarker, an extra exclude shouldn't really matter, right?

ninja-core/src/main/java/ninja/Bootstrap.java
```diff
@@ -0,0 +1,271 @@
...
+    
+    protected void configure() throws Exception {
+
+        // Bind lifecycle support
+        addModule(LifecycleSupport.getModule());
+        
+        // Scheduling support
+        addModule(SchedulerSupport.getModule());
+
+        // TODO: this is where problems start to occur...
```
raphaelbauer: What TODO is that?

ninja-maven-plugin/src/main/java/ninja/maven/NinjaRunMojo.java
```diff
@@ -297,20 +358,31 @@ else if (file.getName().endsWith(".jar") && this.watchAllClassPathJars) {
...
-        if (contextPath != null) {
-            String systemPropertyContextPath = "-Dninja.context=" + contextPath;
+        if (getContextPath() != null) {
+            String systemPropertyContextPath = "-Dninja.context=" + getContextPath();
             jvmArguments.add(systemPropertyContextPath);
         }
         
+        if (jvmArgs != null) {
+            // use excellent DrJava library to tokenize arguments (do not keep tokens)
+            List<String> tokenizedArgs = ArgumentTokenizer.tokenize(jvmArgs, false);
```
raphaelbauer, line 374: Hmm.... Why don't we just use Guava's splitter to split the stuff easily? Copying the DrJava file with only use case... hmm...

raphaelbauer:
Hi Joe, that looks great!

could you please check the following:
- Fix the merge conflicts (shouldn't be too hard - coming from (https://github.com/ninjaframework/ninja/pull/383))... If it's not working out - let me know - I can do that, too...
- Make sure there are no copyright joelauer or fizzed stuff in the copyright headers
- Usually we don't use @author tags (not a hard rule tough)...
- Usually the codebase uses the for(String var: bla) { .. } not sure... we can relax the spacing of the ":" 
- I don't like integrating the DrJava source code file... I am almost sure we can replace that with a one liner from Guava... (I may be wrong..)....

Once that's done I'll merge the PR :) And I want to check out undertow!

I am also currently releasing Ninja 5.2.0 (with a breaking change because of the much improved file upload handler)...

This change should go out with Ninja 5.3.0 or so...
Hope that's ok... 

Interesting read the code is... 

ra!

ninja-core/src/main/java/ninja/Bootstrap.java
jjlauer: I was attempting to figure out how to build a "lite" version of Ninja -- just an accidental leftover comment that I'll remove.

ninja-maven-plugin/src/main/java/ninja/maven/NinjaRunMojo.java
jjlauer, line 374: Tokenizing shell-like command lines is actually not a trivial problem.  You need to deal with spaces, commas, arguments inside quotes, etc.  Everything on the Internet pointed to DrJava's implementation as the only one not broken.  Looks like even Apache Commons is notorious for being wrong.  It's unfortunately not distributed on Maven central, so copying and pasting it seemed like the best solution.  It's BSD licensed (which is compatible w/ Apache) and other than the package, it wasn't modified.

jjlauer:
@raphaelbauer Rebased my branch and did quite a bit of work to make it merge w/ 5.2.0. Lots of cleanup of javadocs (@author and copyright fixes).

jjlauer:
@raphaelbauer Fixed the changelog merge conflict.  Do you think you'll be able to review this soon?

momiji:
Just answering for the out-of-scope maven plugin...

For the good:
1. excludes
Good point, I'll have to test, but if it works, great :)
2. watchDirs
Understand also, we definitely need a one-line, so we don't really need the array, it will fit both needs. great :)
3. .ftl.html
You made the point, I should had a better look in the doc, just need to verify it works.

The less good:
4. target/classes
We need to have both "target/classes;local;../x/target/classes" or only "local", depending on whether we use jrebel or not.
The problem is I don't want to reconfigure eclipse or having to run mvn eclipse:eclipse, as it makes it more complex to use, and will confuse developers.
Having watchDir seemed to be a good way for doing this.
5. classpaths
Adding classpaths allows to use other project's classpath. I'm not sure if it is what you called "multi-module project", and maybe is there a better way for doing this.

What do you think?

ninja-core/src/main/java/ninja/standalone/AbstractStandalone.java
```diff
@@ -0,0 +1,402 @@
...
+            this.configure();
+        } catch (Exception e) {
+            logger.error("Unable to configure {}", name, e);
+            System.exit(1);
+        }
+ 
+        try {
+            this.start();
+        } catch (Exception e) {
+            logger.error("Unable to start ", name, e);
```
raphaelbauer: {} missing

ninja-core/src/main/java/ninja/Bootstrap.java
```diff
@@ -0,0 +1,267 @@
...
+                = resolveApplicationClassName(ROUTES_CONVENTION_LOCATION);
+
+        if (doesClassExist(applicationRoutesClassName)) {
+
+            Class<?> clazz = Class.forName(applicationRoutesClassName);
+            
+            ApplicationRoutes applicationRoutes = (ApplicationRoutes) injector
+                    .getInstance(clazz);
+
+            // TODO: this is better if bound
```
raphaelbauer: what does that comment mean?

ninja-core/src/main/java/ninja/standalone/Standalone.java
```diff
@@ -0,0 +1,153 @@
...
+
+package ninja.standalone;
+
+import com.google.inject.Injector;
+import java.util.List;
+import ninja.utils.NinjaMode;
+import ninja.utils.NinjaPropertiesImpl;
+
+/**
+ * Interface for wrapping an underlying server (e.g. Jetty) to bootstrap Ninja
```
raphaelbauer, line 25: Very well documented!

ninja-core/src/test/java/ninja/standalone/AbstractStandaloneTest.java
```diff
@@ -0,0 +1,265 @@
...
+        
+            
+            assertThat(standalone.getNinjaMode(), is(NinjaMode.test));
+            assertThat(standalone.getExternalConfigurationPath(), is("conf/standalone.conf"));
+            assertThat(standalone.getContextPath(), is("/othercontext"));
+            assertThat(standalone.getHost(), is("3.3.3.3"));
+            assertThat(standalone.getPort(), is(9002));
+            assertThat(standalone.getIdleTimeout(), is(70000L));
+            
+        } finally {
```
raphaelbauer, line 125: yea!

ninja-core/src/test/java/ninja/utils/AbstractContextImpl.java
```diff
@@ -0,0 +1,195 @@
...
+
+public class AbstractContextImpl extends AbstractContext {
+
+    public AbstractContextImpl(BodyParserEngineManager bodyParserEngineManager, FlashScope flashScope, NinjaProperties ninjaProperties, Session session, Validation validation, Injector injector) {
+        super(bodyParserEngineManager, flashScope, ninjaProperties, session, validation, injector);
+    }
+
+    @Override
+    protected String getRealRemoteAddr() {
+        throw new UnsupportedOperationException("Not supported yet."); //To change body of generated methods, choose Tools | Templates.
```
raphaelbauer: default comments... (several times in this file)

ninja-core/src/test/java/ninja/utils/AbstractContextTest.java
```diff
@@ -0,0 +1,474 @@
...
+    private FlashScope flashCookie;
+
+    @Mock
+    private BodyParserEngineManager bodyParserEngineManager;
+
+    @Mock
+    private Route route;
+
+    @Mock
+    private ResultHandler resultHandler;
```
raphaelbauer: not used...

ninja-core/src/test/java/ninja/utils/AbstractContextTest.java
```diff
@@ -0,0 +1,474 @@
...
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class AbstractContextTest {
```
raphaelbauer, line 50: very well written testcase!

ninja-test-utilities/src/main/java/ninja/NinjaDocTester.java.save
```diff
@@ -0,0 +1,61 @@
+/**
+ * Copyright (C) 2012-2015 the original author or authors.
+ *
```
raphaelbauer, line 3: hmm... is that class needed? (.save)

ninja-core/src/main/java/ninja/standalone/AbstractStandalone.java
```diff
@@ -0,0 +1,402 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Abstract Standalone that implements most functionality required to write
+ * a concrete Standalone.  Introduces new doStart(), doStop(), and doJoin()
+ * methods which are actually where you'll place most of your logic.  You'll
+ * also want to subclass the configure() method to add any configuration
+ * specific to your Standalone.  See NinjaJetty for example concrete implementation.
+ */
+abstract public class AbstractStandalone<T extends AbstractStandalone> implements Standalone<T>, Runnable {
```
raphaelbauer, line 42: I don't want to start a big discussion - usually composition worked better for me in the long run... But it's okay for me the way it is now...

raphaelbauer:
Very nice feature! Very clear and concise code. Code style is cool - I especially like the hamcrest matchers. Tests are very well written. 

Just some very tiny comments. Feel free to merge any time :)

PR #412, opened 2015-10-16
Context.getParameterFileItems() now returns Map<String, List<FileItem>>

jjlauer:
Aligns the new FileItem feature with the rest of Context by using a List<> rather than an array.  Better to adjust this quickly before too many rely on the new interface.


raphaelbauer:
Nice! Thanks!

PR #416, opened 2015-10-22
Allow remember me functionality via session cookie

watsonmw:
https://github.com/ninjaframework/ninja/issues/415

Added Session.setExpiryTime() for setting the expiry time of sessions.

Added 'remember me' toggle to the JPA blog to demonstrate Session.setExpiryTime().


jjlauer:
Pretty useful feature.  I wouldn't mind seeing a few other test cases to verify the expiration actually happens -- seems like the 3 added just verify it's in the cookie.

watsonmw:
Added a test to make sure the session expires.  Went a bit mock crazy, but it seemed the way to go, if there's a better way let me know!

jjlauer:
Nice, looking better. Few more thoughts.  Rather than calling it SessionTime -- a more generic ninja.utils.Clock would be better (wish we were at Java 8 so we didn't have to invent our own).  Then mocking it works great for unit testing.

I know there is a mix of unit testing standards thru the evolving codebase, but to keep the unit tests clean, please use statically imported assertThat and the matchers (e.g. assertThat(obj, is(not(nullValue())))

Could you change the keys to string values in the unit tests as well?  For example, sessionCookie1.put("1", "2"); is a bit more confusing than sessionCookie1.put("a", "2");

@raphaelbauer will probably have more feedback as well, but hopefully I've covered most of what he'd ask for as well.  Curious if you ran into any issues with your implementation during your additional unit tests?

watsonmw:
Ah I like ninja.utils.Clock better!

The thing I fixed was allowing the expiry time to be programatically removed when save cookies on change only is on. Not sure save-cookie-on-change-only makes sense with expiry time, but hopefully the docs are enough to clear up any confusion (the behaviour with this change, is what we have currently).

If having an expiry time fits ok on the default SessionImpl, it might be worth considering if both the expiry and timestamp should be saved in cookies.  I think having just the expiry time would work...

ninja-core/src/main/java/ninja/session/Session.java
```diff
@@ -30,7 +30,8 @@
     String AUTHENTICITY_KEY = "___AT";
     String ID_KEY = "___ID";
     String TIMESTAMP_KEY = "___TS";
-    
+	String EXPIRY_TIME_KEY = "___EP";
```
raphaelbauer: really tiny... but please make sure to use spaces and not tabs in the source code...  :)

ninja-core/src/site/markdown/documentation/basic_concepts/sessions.md
```diff
@@ -95,6 +95,33 @@ public Result clearSession(Session session) {
 }
 </pre>
 
+Session Expiry
+--------------
+
+You can set an expiry time on sessions.  The expiry time is checked every time
```
raphaelbauer, line 101: Nice description!

raphaelbauer:
Hey! That looks nice...

Just to clarify what we are doing here... People understand different things under the name "remember - me" functionality.

Before your change we had rolling - potentially infinitely valid - sessions. A session will timeout if there is long enough no request to "refresh" the timestamp. Otherwise it will just be prolonged.This is verified by the browser (Max-Age) and Ninja via the signed session cookie. Securing that via Max-Age is nice, but we don't trust the browser of course, so we do it ourselves.

Now:
We can get the old behaviour of rolling infinite sessions (with Max-Age and Ninja timestamp). But we can now also specify an expiry time that will make the session time out after a fixed time - let's say 30 days. That's also set via Max-Age, but also verified via Ninja itself - because we don't trust the browser...

Did I sum it up correctly?

Cheers,

Raphael

watsonmw:
That's a good summary.  The max-age/expiry is still verified by Ninja in case the cookie expiry is faked by the browser client.  This is done in both the case where the expiry is set in application.conf and explicitly on the session in code.

The expiry time is only refreshed when a Ninja application changes a session variable (by default).  If 'application.session.send_only_if_changed' is set to false, then the expiry time is updated every time a request is made.  Personally I  would like if 'send_only_if_changed' was something you had to turn on, but I guess it's like this because by default sessions don't expire.

ninja-core/src/main/java/ninja/session/Session.java
watsonmw: Ok!  The rest of that file is indented with tabs (below that line) though - didn't reformat it, as it would just confuse the change request.

raphaelbauer: Yea... Session is a very very old class :D

raphaelbauer:
Okay... then we are on the same page... Will check-out the code and merge asap... Very useful contribution Mark!

jjlauer:
@watsonmw PR looks great.  Can you add your change to the changelog.md file, optionally add yourself to the team.md file, and then you'll need to sign the contributor agreement.  Links to the contribute agreement are on this page: http://www.ninjaframework.org/developer/getting_started.html

ninja-core/src/main/java/ninja/session/SessionImpl.java
```diff
@@ -139,13 +135,75 @@ public void init(Context context) {
         }
     }
 
+    protected boolean shouldExpire() {
+        if (sessionExpireTimeInMs != null) {
+
+            // Make sure session contains valid timestamp
+            if (!data.containsKey(TIMESTAMP_KEY)) {
+
```
raphaelbauer: Tiny - but please compact the code by using empty lines only sparsely (I know this may not have been the case before, but it's just more concise :) ). Thanks!

ninja-core/src/test/java/ninja/session/SessionImplTest.java
```diff
@@ -398,7 +381,101 @@ public void testThatCookieUsesContextPath() {
...
+    @Test
+    public void testExpiryTime() {
+
+        // 1. Check that session is still saved when expiry time is set
+
+        Session sessionCookie1 = getNewSession();
+        sessionCookie1.init(context);
+
+        sessionCookie1.put("a", "2");
+
```
raphaelbauer, line 395: Also too many empty lines... Please compact the code a bit.. Thanks!

ninja-core/src/test/java/ninja/session/SessionImplTest.java
```diff
@@ -398,7 +381,101 @@ public void testThatCookieUsesContextPath() {
...
+        when(context.getCookie("NINJA_SESSION")).thenReturn(cookieCaptor.getValue());
+
+        // ... and roundtrip it into an new session
+        Session sessionCookie2 = getNewSession();
+        sessionCookie2.init(context);
+        return sessionCookie2;
+    }
+
+    private Session getNewSession() {
```
raphaelbauer: Would you mind renaming the method to createNewSession()? .. get... always has the association to JavaBeans... which is not the case here :) Thanks!

raphaelbauer:
Very good PR! Nice feature, good tests and documentation. I added some tiny remarks (just style stuff). @jjlauer Feel free to merge as you see fit :) Thanks!

watsonmw:
Updated based on the feedback above and signed the contributor agreement (actually had signed last week I think). Cheers!

jjlauer:
Awesome work @watsonmw ! Feature will be very useful. Mucho appreciated.

danielsawan:
I got an horrible day today in production env :/
First of all i am using the last version of Ninja (5.2.2).
I was using Ninja without any session problem until yesterday.

But yesterday i tried the new feature that let us controle the user session timeout.
my cookie parameters are :

Configuration file :
application.cookie.encryption=true
application.session.send_only_if_changed = false

LoginController : 
session.setExpiryTime(60 \* 60 \* 1000L); // 1 Hour

The idea (as the doc say) is to refresh the session each time a request is made and to restart the 1 hour timeout countdown.

I want to precise that the two major factor of this bug are those 2 configuration :

1) application.session.send_only_if_changed = false (if true there is no bug anymore)
2) session.setExpiryTime(xxxx) (if not expiryTime setted then there is no bug)

When both of those condition are true then you can enter my nightmare.

The problem is that (i don't know how) User A & User B & user C when  coming to their panel was loged as User D

while playing with multiple session with different brownser i manage to do this but i don't know what is doing this and how to reproduce it. But i just disconnect / reconnecter user A then B etc.. and sometimes the bug occurs. Let's say user A disconnect then i just hit refresh and he is connected as user B even without logging in...

Now a second bug :
I don't know if there is the same bug or an other but this bug is easily reproducable.

1) start a fresh brownser
2) Log with user A
3) disconnect user A from his panel.
4) normaly now you are at your frontend main page and not logged in.
5) Just hit F5 to refresh the page
6) Abracadabra you are logged in again

I tried this with Chrome & Firefox same problem.

I think that this is the main bug and the user session switching is a sort of border effect...

jjlauer:
Daniel,

What does "3) disconnect user a from his panel." mean?  Do you mean "log
user out by clearing the session" ?  If so how are you doing that in java
code?

-Joe

watsonmw:
Hi Daniel,

I tried to reproduce using two users in the Ninja blog example, but didn't
see this issue.  Here's the logout code for reference:

https://github.com/ninjaframework/ninja/blob/develop/ninja-servlet-jpa-blog-integration-test/src/main/java/controllers/LoginLogoutController.java

public Result logout(Context context) {
    // remove any user dependent information
    context.getSession().clear();
    context.getFlashScope().success("login.logoutSuccessful");
    return Results.redirect("/");
}

danielsawan:
@jjlauer disconnect mean clearing the session (cookie). And yes i do that i java code when a user click on disconnect bouton.

@watsonmw my logout look like :

```
    @GET
    @Path({"/logout"})
    public Result Logout(Session session) {
        // remove any user dependent information
        session.clear();
        return Results.redirect("/");
    }
```

@watsonmw don't forget that the bug only happen when those 2 conditions are set :

1) application.session.send_only_if_changed = false (if true there is no bug anymore)
2) session.setExpiryTime(xxxx) (if not expiryTime setted then there is no bug)

danielsawan:
I feel like when those two option are set with the "bugy" parameters, Ninja store the cookie in memory.
Because i can't explain otherwise how can Ninja send a valid cookie of another existing user ?

I set the user Id in session (cookie) only when login. I checked that the cookies was clean on disconnecting (chrome dev mod). Then hit F5 and Ninja generate me a valide session ID even without login. That's really like the cookie info was save in memory and not cleaned when disconnecting.

Do you have something like that @watsonmw  in your code ?

jjlauer:
Daniel,

Ninja doesn't store cookies in memory.  They are stored on the client (your
browser) and sent with each request. If you inspect the interaction with
your browser using something like Chrome's developer tools, you'll maybe
see what is happening with the session simply by inspecting the headers
your browser is sending/receiving. I'd suggest tracing it more closely to
see what's happening in your logout method.  Perhaps save the cookie
request header & set-cookie response header for each of your 6 steps you
say is reproducible.  I'd be curious what you track.  Make sure cookie
isn't fully encrypted (that's a ninja option) so we can see the values of
the fields in it for easier debugging.

-Joe

danielsawan:
I think i gonna make a sample app with no filters / extractors etc...
But i don't think this a code i made which generate the bug because then why it is a Ninja option which create / make disapear the bug then ? I a 100% sure i only create session in one place the Login page.

But you are right i should simplify the problem by creating a simple app maybe. But before i wanted to see if someone else manage to generate this bug with your own applications.

watsonmw:
Only things I can think of:
- You clear the cookies in the logout request (request a), while another
  request is being processed (request b), request b never sees the cleared
  cookie, so saves and extends the timeout again. Whatever request b is,
  needs to leave the cookies alone so logout can occur.
- The same SessionImpl is used twice, not sure how this would inadvertently
  happen, the guice injection code, always grabs a new one the way things are
  setup.

danielsawan:
If the user have auto refresh option enabled (a website option), then his browser make many many polling using jquery ajax request and the bug happen. (imagine a sort of live view of Google Analytics)

When i disable auto refresh option on the website then there is no bug when i disconnect. It work as intended.

So what is seems to happen is : 

1) The user log into his web site
2) His browser is now polling many requests (10 req / sec)
3) he click the disconnect button
4) he go back to frontend index page and the website doesn't show the "disconnect button" (which normaly appear when a user is loged in and on the index). But the cookies contains a session. (as if the session was created when the ajax finish and after the call of session.clear() when returning the ajax results)
5) the user click F5 on the index which now detect the session and log him (that's what i was calling autologin)

When those option are configured like this :

1) application.session.send_only_if_changed = false 
2) session.setExpiryTime(xxxx)

The bug always happen then. When there is no setExpiryTime or send_only_if_changed=true there is no bug. Do you know why ?

And i still doesn't understand what happen when User A Become User B. This bug is not reproducable easily because i feel that it occurs when User A and User B are spamming polling requests. Then User A disconnect. Hit refresh and get the session of User B. This is just crazy as if Ninja at a point of execution send back the session of User B To User A... (User A and User B are on different country. It is not the same computer with 2 browser)

watsonmw:
Sounds like you need the AJAX requests not to update the session cookie.
When always save cookie is on, each AJAX request will save the cookie.

danielsawan:
I manage to reproduce the "stay logged in bug" on a new web app very easily. The idea is to make a page with many jquery polling request and it is almost impossible to logout with

1) application.session.send_only_if_changed = false 
2) session.setExpiryTime(xxxx)

It does logout but then just hit a refresh to see that we are stilled logged in.

Now the second bug (session share between users)

After many hours of investigation i manage to find the source of this but don't understand why.
I have a filter which only authorised beta user to access my website.

It contains : 

```
    @Override
    public Result filter(FilterChain chain, Context context) {
        String betaKey = context.getSession().get("betaTester");
        if (betaKey == null || !CryptTool.isValidcrypt(BETAKEY, betaKey)) 
        {
            return Results.text().render("very soon ...");
        } 

        return chain.next(context);
    }
```

I suspect that some times User B become user A because of this line of code :

```
return Results.text().render("very soon ...");
```

I feel like when user A reach the code return Results.text().render("very soon ..."); then the filter has not really finished the filtered request of User A and keep it in memory. Then user B call the same filter and get the non well finished filter request of user A and get User A Session.

danielsawan:
What i am sure about is that.
1) The Story is true : User B become User A (i can make a video to show you)
2) Something stay in Ninja memory and i am not putting anything user related in memory by myself.
3) Both bugs only happen when application.session.send_only_if_changed = false and session.setExpiryTime is set (both options at the same time)

watsonmw:
Regarding the first issue, this is the cookie save race condition I was referring to earlier. The session  cookie is read when the request is made and then saved after the browser receives the response.  Since requests can overlap, it is possible to logout in one request, expiring the cookie, then in another request happening at the same time the cookie timeout is extended.  I think we should update the docs to explain this. Regardless of 'application.session.send_only_if_changed' this can cause issues, but when the cookie is saved every time there's a much higher chance of running into it.

As for what you can do (if you still want to save session expiry client side):
- First, enable send_only_if_changed
- Then, only call setExpiryTime() on the main HTML requests that you want to extend the life of the session on, don't do it on the AJAX requests.  This way the cookie will only be saved for the HTML requests, and the AJAX calls won't resave the session during logout.  The issue can still occur if you have two tabs open and refresh one, while logging out on another, but that's not a big deal usually.  You can do this with a filter.

danielsawan:
I only use setExpiryTime() in the login request and never use it again. There is no setExpiryTime() in any filters if it is what you mean.

danielsawan:
here is some news.

As i told you in my last comment i suspected that the user switching bug was coming from my filters

```
    @Override
    public Result filter(FilterChain chain, Context context) {
        String betaKey = context.getSession().get("betaTester");
        if (betaKey == null || !CryptTool.isValidcrypt(BETAKEY, betaKey)) 
        {
            return Results.text().render("very soon ...");
        } 

        return chain.next(context);
    }
```

And i was right. I replaced the return like this :

```
//          return Results.text().render("very soon ...");
            return Results.unauthorized().html().template("views/system/beta.ftl.html");
```

And there is no user session switching anymore.
I thought return Results.text() was the same thing that return Results.html() only that in the .text() we can render our text directly in the .render (i use .text() alot with ajax request). But seems like not. I feel like the the .text() don't send the cookies. (or maybe it is other things happening here and i am wrong).

I still don't understand what is the real cause of user Switching but it seems to be a very very bad idea to use return Results.text() in Filters...

And the first bug of disconnecting but still connected is solved. It is how browser deal with ajax request interfering with the fact that in Ninja the session is client side. I don't know if there is something we can do about this because i am using application.session.send_only_if_changed = false in order to refresh the session timeout on each request... So i guess if i want to use this option, my users will have to disconnect 2 times. I think a solution is to redirect them when they click on disconnect to a sort a count down page (i saw many bank website doing this and i will always wondering why lol. Now i can understand why i guess)

watsonmw:
I dont think you need an extra page, just make sure the cookie is not saved with the AJAX requests.  Try the following:
- Set application.session.send_only_if_changed = true
- Call setExpiryTime() on every controller method you want to extend the expiry time on, you can do this with a filter.  i.e. don't call setExpiryTime()  on the AJAX requests.

Regarding the user switching thing you are seeing, the ninja framework side of things look correct, at least in standalone/servlet mode.  There's no state saved between requests, you'd have to set custom module bindings on Context or Session to get that to happen.  Otherwise it's a logged in cookie getting reused by clients somehow.

danielsawan:
Thanks for your help @watsonmw  but as i told you i manage to find the bug of user switching : it was from the  return Results.text().render("very soon ..."); in the filter.

Replacing by return Results.unauthorized().html().template("views/system/beta.ftl.html"); fix the bug i get.

I still didn't understand what happen and as it is a multi request env it is not that easy to debug. But there is definitly something strange here.

I spend 2 full days searching where this bug come from and i have no more time to spend on this now that i found a fix but i am afraid that the bug is still here and only occurs under certain circonstance.

I didn't manage to reproduce the user switching bug on a sample project but i easily can reproduce this on my real projet.

If someone want to see this strange problem i can easily share my desktop using Teamviewer (we just have to fix a date ;)).

I trust Ninja is not saving the session in memory in an intended way but i think there is actualy a bug in Ninja while using :

application.session.send_only_if_changed = false
Call setExpiryTime() 
and a filter with  return Results.text().render("blabla");

i feel that the  return Results.text().render("xxx") in filters don't clear the context sometimes and give it back to the following user coming from here...

I really need an expert help to this because it is a dangerous thing (cc @raphaelbauer & @jjlauer )

watsonmw:
Some other questions:
- How do you run/host your ninja application (e.g. Standalone with Jetty) ?
- Is there a reverse proxy in front of the application (for example Nginx)?
- Do you set your own module (Guice) bindings for Ninja's Context or Session?
- Each page issuing 10s of AJAX requests a second sounds like a lot, make sure the browser is not queuing up tons of requests and issuing them after logout (Just use Chrome DevTools Network panel).
- This would be the worst web-cache ever, but your "Content-Type: text" status 200 requests could be getting cached, and if the Set-Cookie header is getting cached that's real bad news.  The "Vary: Cookie" might help but:
- Are you running over SSL?  SSL traffic over the public internet won't be cached along the way, it's the only way to really ensure this doesn't happen.

danielsawan:
. How do you run/host your ninja application (e.g. Standalone with Jetty) ?
Standalone Jetty under dev using maven and as jar on prod env and bug happens on both

. Is there a reverse proxy in front of the application (for example Nginx)?
On prod yes on dev no

. Do you set your own module (Guice) bindings for Ninja's Context or Session?
Nop i am using Ninja default one

. Each page issuing 10s of AJAX requests a second sounds like a lot, make sure the browser is not queuing up tons of requests and issuing them after logout (Just use Chrome DevTools Network panel).
That's why it is hard to logout when we are on that specific page but i don't see what i can really do.

. This would be the worst web-cache ever, but your "Content-Type: text" status 200 requests could be getting cached, and if the Set-Cookie header is getting cached that's real bad news. The "Vary: Cookie" might help but:
Didn't understand :/. But i disabled cached with Chrome dev tool (checkbox option)

. Are you running over SSL? SSL traffic over the public internet won't be cached along the way, it's the only way to really ensure this doesn't happen.
On prod yes and the bug still happen

If need i can open a teamviewer session to show you. I can't share the source or make the project public because it is not mine. I am developping this for a little company.

But when you take a little step back to this problem. How can Session from User A going to User B while i never coded that myself as Ninja did not. This is definitly in the Filters and i got only 2 Filters with 4 line of code each.
So i personnaly bet on a bug but i need to show this to someone because i feel like i am not trusted :p Maybe i am using the Filters the wrong way i don't know. I need someone to review those 8 line of codes especially on my specific project.

danielsawan:
@watsonmw can you check your email Please ?

danielsawan:
After the discussion i had with @watsonmw on Teamspeak i decided to try to understand why this bug only happen on my prod env and not on my sample test app and i think i found.

Can you please test this app and tell me if you the bug happen to you also ?

in order to make the bug appear i had to change the catch all in the Routes.java to what i have in production :

///////////////////////////////////////////////////////////////////////
        // Index / Catchall shows index page
        ///////////////////////////////////////////////////////////////////////
        router.GET().route("/._").with(Results.redirect("/")); // this cause the bug
//        router.GET().route("/._").with(ApplicationController.class, "index"); // this is ok

here is the sample ninja app : https://dl.dropboxusercontent.com/u/23483842/session-bugs.jar

So in order to make the bug appear

0) Use Chrome not Firefox (don't manage to reproduce the bug all the time on firefox)
1) Go to http://localhost:8080/
2) Login as User A or User B
2') wait at least 5 seconds 
3) Call http://localhost:8080/error 
4) Open new chrome private browser session or a different browser (firefox) to http://localhost:8080/
5) normaly you have been auto logued as the user you used in 2)
6) if not autologued try to hit F5 (refresh) in the new private browser 4).
7) if not autologued try to call http://localhost:8080/error in the new private browser 4).

watsonmw:
Ahhh... I think what happens is the ninja.Result for redirects is static, but the cookie data gets mistakenly saved there.  I'm not sure what other static ninja.Results can exist.  I guess the Ninja fix would be don't save cookies to ninja.Result on redirect, and/or don't have static ninja.Results that can be modified.

Phew, this one took a lot of digging to find, thanks for persevering and coming up with the minimal test case!

watsonmw:
Added a pull request to update the docs with recommendations above, and fixed the issue with Routes sending the same Result object instance on every request.

danielsawan:
nice catch ! :)

raphaelbauer:
:+1:

PR #422, opened 2015-11-12
Fix garbled parameters in multipart requests

mallowlabs:
I found that `Context#getParameter` returns garbled texts in multipart requests from v5.2.0.
Because `Streams#asString()` reads stream with the platform encoding.
I fixed this problem.

Environment:

```
$ mvn --version
Apache Maven 3.2.5 (12a6b3acb947671f09b81f49094c53f426d8cea1; 2014-12-15T02:29:23+09:00)
Maven home: C:\usr\apache-maven-3.2.5\bin\..
Java version: 1.8.0_65, vendor: Oracle Corporation
Java home: C:\Program Files\Java\jdk1.8.0_65\jre
Default locale: ja_JP, platform encoding: MS932
OS name: "windows 10", version: "10.0", arch: "amd64", family: "dos"
```

### DIFF
--- a/ninja-core/src/site/markdown/developer/changelog.md
+++ b/ninja-core/src/site/markdown/developer/changelog.md
@@ -1,6 +1,7 @@
 Version X.X.X
 =============
 
+ * 2015-11-13 Fixed garbled parameters in multipart requests (mallowlabs)
  * 2015-10-18 Added Session.setExpiryTime() to control session expiry (watsonmw)
  * 2015-10-18 OptionalBinder for ObjectMapper and XmlMapper, so that users can easily override
               the default types for both of them. (amit2103/jjlauer)
--- a/ninja-servlet/src/main/java/ninja/servlet/NinjaServletContext.java
+++ b/ninja-servlet/src/main/java/ninja/servlet/NinjaServletContext.java
@@ -537,9 +537,17 @@ public class NinjaServletContext extends AbstractContext {
                 FileItemStream item = fileItemIterator.next();
 
                 if (item.isFormField()) {
+
+                    String charset = NinjaConstant.UTF_8;
+
+                    String contentType = item.getContentType();
+
+                    if (contentType != null) {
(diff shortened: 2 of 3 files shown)


raphaelbauer:
Excellent finding and fix! @jjlauer if there are no objections from your side I'll merge until the end of the week...

jjlauer:
Should UTF-8 be hard coded or should it be the content type of the http request?  The other question is whether this should be in ninja-servlet or ninja-core -- ninja-core has the new AbstractContext class which should have most functionality implemented -- unless that functionality relies on the concrete implementation.  I didn't take a close enough look to judge this either way, but would appreciate it if you took a look and evaluated that as well.

mallowlabs:
@jjlauer

> Should UTF-8 be hard coded or should it be the content type of the http request?

This is good point.
I fixed my branch, I added to check content type of parts.

@raphaelbauer
Please review me again.

jjlauer:
@mallowlabs Getting closer.  While I think your impl is most likely correct now, I think it needs to be unit tested a tad more. 

Can you add a unit test that verifies the charset of the multipart is used?  The RFC for it is https://www.ietf.org/rfc/rfc2388.txt  From section 4.5:

Each part of a multipart/form-data is supposed to have a content-type.  In the case where a field element is text, the charset parameter for the text indicates the character encoding used.

For example, a form with a text field in which a user typed `Joe owes <eu>100` where `<eu>` is the Euro symbol might have form data returned as:

```
--AaB03x
content-disposition: form-data; name="field1"
content-type: text/plain; charset=windows-1250
content-transfer-encoding: quoted-printable
Joe owes =80100.
--AaB03x
```

mallowlabs:
Thank you for your good advice.
I added a unit test.
Please check again.

jjlauer:
Looks great.  Nice catch!

raphaelbauer:
Great work - and great comments :+1:

mallowlabs:
Thank you :smile:

PR #432, opened 2015-12-10
Allow DTOs to work with multipart requests.

watsonmw:
I noticed I had some user agents / client's making multipart POST requests instead of x-form POST requests to controller methods I was using Dtos on, so the following wasn't working (dto was null):

```
public class Dto {
  String param1; 
  String param2;
}

public class MyController {
    public void post(Dto dto) {
       // null for multipart requests, not null for application/x-form requests
    }
}
```

This patch allows non multipart and multipart POSTs to work similarly.  You can still handle the multipart body in a streaming fashion, by specifying only a "Context" parameter or nothing at all for the controller method arguments.

I should mention I couldn't find a reason for NoFileItemProvider to exist, it was just causing some early out logic when calling the Context.get_Paramater_() methods with multipart requests, requiring a FileItemProvider annotation for controller methods that didn't even use files.

### DIFF
--- a/ninja-core/src/main/java/ninja/ContentTypes.java
+++ b/ninja-core/src/main/java/ninja/ContentTypes.java
@@ -17,12 +17,11 @@
 package ninja;
 
 public interface ContentTypes {
-
     String TEXT_HTML = "text/html";
     String APPLICATION_JSON = "application/json";
     String APPLICATION_JSONP = "application/javascript";
     String APPLICATION_POST_FORM = "application/x-www-form-urlencoded";
+    String MULTIPART_FORM_DATA = "multipart/form-data";
     String APPLICATION_XML = "application/xml";
     String TEXT_CSS = "text/css";
-
 }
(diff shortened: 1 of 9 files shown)


jjlauer:
@watsonmw Interesting PR. So a couple things pop out to me.  1) The default implementation was NoFileItemProvider --  and you changed the default to the memory version.  That seems like a significant change from the default.  Are you sure there aren't other side effects you didn't consider, especially from a security/accidental usage context?  2) I think of uploaded files as InputStreams and it seems like one could accidentally try to "stream" it twice pretty easily in your examples (e.g. as a param or inside a dto or both).

watsonmw:
1) Ah I think I see the point of NoFileItemProvider, it's because the body is parsed without knowledge of the parameters being requested (even when you specify the params/dto up front).  So all file items in the request will be serialised (up to the limits specified in the application.conf) regardless of what parameters are specified in the controller method.  Will have to think about this some more....

2) You would get the same stream in both in that case, that's fine the user will get an error saying the stream has closed. It seems like a very rare mistake to make, but we could catch it by throwing an exception from getFileParamter() i think (this problem existed before I made the changes above BTW).

jlannoy:
Replaced by PR #630

PR #434, opened 2015-12-21
Basic SSL support for standalone and SuperDevMode

jjlauer:
The web is quickly moving to SSL-only (and for HTTP 2.0 many impls will only support it over an encrypted channel).

This PR adds support for basic SSL (no fancy optimizations, no algo selection, etc.) in either standalone or SuperDevMode. In dev or test modes, a pre-packaged self-signed Ninja certificate is used for the default keystore and truststore.  Thus, enabling SSL during development is as easy as adding `ninja.ssl.port=8443` to your `conf/application.conf` file.


raphaelbauer:
Not sure what to say here... It looks perfect :) Very well done and a very nice feature!

raphaelbauer:
:+1:

PR #435, opened 2015-12-22
Travis tests on OpenJDK 7 and Oracle 8

jjlauer:
Specify jdks to run travis with (reliance on current travis defaults is brittle)


raphaelbauer:
True...

PR #436, opened 2015-12-22
Unit tests improved for compat with non-servlet standalones

jjlauer:
I've been working on undertow support and the ninja-servlet-integration-test module is really the best "ninja-integration-test" module.  I thought about renaming the module, but wanted to keep this PR simple. There were just a couple very minor changes required to make it a great integration test for a non-servlet standalone such as undertow.


raphaelbauer:
Losing one "test" here, but it's covered somewhere else already... No problem...

PR #443, opened 2016-01-15
Bump Jetty to 9.3.6.v20151106

raphaelbauer:


jjlauer:
Looks like Jetty 9.3.6 jar doesn't run on Java 7 per travis ci.

raphaelbauer:
Yea. Hmm. We should switch to Java8 soon... (Cool that the CI setup found the issue)...

PR #447, opened 2016-01-20
freemarker support custom template model

huwei:
for example:

```
public class NinjaExtModule extends AbstractModule {

    protected void configure() {
        MapBinder<String, TemplateModel> mapbinder
                = MapBinder.newMapBinder(binder(), String.class, TemplateModel.class);
        mapbinder.addBinding("xsrfToken").to(TemplateEngineFreemarkerXsrfTokenDirective.class);
        mapbinder.addBinding("xsrfForm").to(TemplateEngineFreemarkerXsrfFormDirective.class);
    }

}
```

### DIFF
--- a/ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
+++ b/ninja-core/src/main/java/ninja/template/TemplateEngineFreemarker.java
@@ -26,22 +26,15 @@ import java.util.Map.Entry;
 
 import javax.inject.Singleton;
 
-import ninja.Context;
-import ninja.Result;
-import ninja.i18n.Lang;
-import ninja.i18n.Messages;
-import ninja.template.directives.TemplateEngineFreemarkerAuthenticityFormDirective;
-import ninja.template.directives.TemplateEngineFreemarkerAuthenticityTokenDirective;
-import ninja.utils.NinjaConstant;
-import ninja.utils.NinjaProperties;
-import ninja.utils.ResponseStreams;
-
 import org.slf4j.Logger;
 
 import com.google.common.base.CaseFormat;
 import com.google.common.base.Optional;
 import com.google.common.collect.Maps;
 import com.google.inject.Inject;
+import com.google.inject.Injector;
+import com.google.inject.Key;
+import com.google.inject.TypeLiteral;
 
 import freemarker.cache.ClassTemplateLoader;
 import freemarker.cache.FileTemplateLoader;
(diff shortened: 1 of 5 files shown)


raphaelbauer:
Hey! 

that's really useful. Do you think we can implement everything without the need to add the injector itself? That usually leads to very bad code and is not a good example to follow.

What about the approach described here: https://github.com/google/guice/wiki/Multibindings

Basically something like a class or interface FreemarkerDirective (or any name you like).

```
class FreemarkerDirective {

   FreemarkerDirective(String name, TemplateModel templateMode) {
        ...
   }
   String name();
   TemplateModel templateModel();
}
```

And that is then injected via a Set<FreemarkerDirective>... into TemplateEngineFreemarker.

What do you think?

Raphael

huwei:
Hey!

your solution is great.

```
public class TemplateEngineFreemarkerModel {

    private String name;
    private TemplateModel templateModel;

    public TemplateEngineFreemarkerModel(String name, TemplateModel templateModel) {
        this.name = name;
        this.templateModel = templateModel;
    }

    public String getName() {
        return name;
    }

    public TemplateModel getTemplateModel() {
        return templateModel;
    }

}
```

change TemplateEngineFreemarker class：

```
    private final Set<TemplateEngineFreemarkerModel> templateEngineFreemarkerModels;

    @Inject
    public TemplateEngineFreemarker(Messages messages,
                                    Lang lang,
                                    ...
                                    Set<TemplateEngineFreemarkerModel> templateEngineFreemarkerModels,
                                    NinjaProperties ninjaProperties) throws Exception {
        this.templateEngineFreemarkerModels = templateEngineFreemarkerModels;
    }
}
```

```
public class NinjaModule extends AbstractModule {

    protected void configure() {
        Multibinder<TemplateEngineFreemarkerModel> modelBinder = Multibinder.newSetBinder(binder(), TemplateEngineFreemarkerModel.class);
        modelBinder.addBinding().toInstance(new TemplateEngineFreemarkerModel("xsrfToken", new TemplateEngineFreemarkerXsrfTokenDirective()));
        modelBinder.addBinding().toInstance(new TemplateEngineFreemarkerModel("xsrfForm", new TemplateEngineFreemarkerXsrfFormDirective()));

    }

}

```

Is it right？

raphaelbauer:
Yes! That looks nice :)

Please don't forget that we also need documentation (so that anyone knows about your great feature) and some tests to make sure the feature does not get broken by accident. And you need to sign the contributor's agreement. More here: http://www.ninjaframework.org/developer/getting_started.html

Looking forward to the feature! Good work :)

huwei:
But that make an error.
By default,  have not custom template, application not inject TemplateEngineFreemarkerModel .
Will throw exception when running the application.

```
com.google.inject.CreationException: Unable to create injector, see the following errors:||1) No implementation for java.util.Set<com.meicanyun.ninja.template.TemplateEngineFreemarkerModel> was bound.
```

So I guess use Injector in TemplateEngineFreemarker is required.

raphaelbauer:
I'd need some code to check out why that happens. If  https://github.com/google/guice/wiki/Multibindings works it our approach will work, too.

Can you try to make TemplateEngineFreemarkerModel an interface and create a class that implements it? That's the only difference I can see now...

raphaelbauer:
Hey @huwei , you are right. I'll have a look. There must be a nice solution :)

raphaelbauer:
Ok. The "solution" to the problem: Guice has to know that there is a Set<TemplateEngineModels> that will be injected somewhere. Otherwise it complains. Therefore we have to call the multibinder at least once in the modules that create the injector.

That means we have to add something like

```
Multibinder.newSetBinder(binder(), TemplateEngineFreemarkerModel.class);
```

in Ninja's Configuration.java. But also in all tests that use an injector but don't use Configuration.java as basis (eg TemplateEngineManagerImplTest).

Then it works for me nicely. Can you confirm that?

huwei:
Yes! 
In addition, I recommend to use MapBinder. If SetBinder, not only implementation TemplateModel, but also implementation TemplateEngineFreemarkerModel. Writing custom templates will become more complex.

PR #448, opened 2016-01-20
Enhanced parsing of input values (post or parameters)

jlannoy:
The goal of this PR is to enhance the parsing of strings to simple types, from the parameters or from the request. 
In details, it allows :
- To avoid duplicated code or differences between ParamParsers and SwissKnife converters (their job are in fact the same)
- To allow the POST body engine to register validation errors on conversion failure
- To avoid registering enumeration at start
- To allow users to register their own ParamParser
- Even overriding default ones, which can be useful to specify a particular DateFormat

Notes : 
- Custom ParamParsers are instancied via an Injector, so that users can make usage of @Inject dependencies
- The convert method for Date from the SwissKnife class have been translated to a ParamParser, based on LocalDateTime (as the original method). I’m thinking about something more useful, with date format from the properties file based on current Locale ; but I didn’t want to make a bigger PR. The fact is that with this PR, each user will be able to write a parser with it’s own format.


jlannoy:
Raphael, can you do a first review of this? If OK, then I'll check what's going on with the branch conflicts (it must be because I started it before last release).

raphaelbauer:
Hey! This looks good (I just roughly looked at the code). Some really good stuff in this PR! Could you make CI pass so that I can run the code? Thanks!

jlannoy:
Voila. Now it's okay ;-)

raphaelbauer:
Checking out the PR now... :)

ninja-core/src/main/java/ninja/utils/SwissKnife.java
```diff
@@ -40,19 +34,6 @@
 
     private static final Logger logger = LoggerFactory.getLogger(SwissKnife.class);
 
-    private static final Map<String, Method> CONVERTERS = new HashMap<>();
```
raphaelbauer, line 43: Cleaning SwissKnife up and having that code only once is genius... :+1:

ninja-core/src/site/markdown/documentation/basic_concepts/controllers.md
```diff
@@ -125,7 +125,10 @@ If your custom object has a date/timestamp field, first make sure that you use t
 <a href="http://docs.oracle.com/javase/7/docs/api/java/util/Date.html">java.util.Date</a> 
 type and when you pass some date to the controller pass in the 
 <a href="http://joda-time.sourceforge.net/api-release/org/joda/time/format/ISODateTimeFormat.html#localDateOptionalTimeParser%28%29">format</a> 
-accepted by Joda Time library.
+accepted by Joda Time library. Or you can register a <code>ninja.params.ParamParser</code> with your own date format parsing in your <code>conf.Module</code>:
+<pre class="prettyprint">
+ParamParsers.registerParamParser(Date.class, MyDateParamParser.class);
```
raphaelbauer, line 130: That makes me think whether it is clever that ParamParsers is static. (Not an issue of this PR). Do you think we should create an issue for that?

jlannoy, line 130: Not sure. I think it's quite the same discussion as the NinjaProperties given to a module as constructor arg... If it is not static, you'll not be able too to inject it in the conf.Module. So where will users do the early job of registering parsers? 

Not that, with the way I did the PR, at least all custom parsers will now be Guice-aware.

ninja-core/src/main/java/ninja/bodyparser/BodyParserEnginePost.java
```diff
@@ -62,21 +75,34 @@
...
-
-                    // convert the values based on the field type,
-                    // supported types are collections, arrays and boxed primities
-
-                    if (Collection.class.isAssignableFrom(fieldType)) {
-                        convertedValue = SwissKnife.convertCollection(values, getGenericType(field));
+                    
+                    if (Collection.class.isAssignableFrom(fieldType) || List.class.isAssignableFrom(fieldType)) {
+                    	
+                    	ListParamParser<?> parser = (ListParamParser<?>) ParamParsers.getListParser(getGenericType(field));
```
raphaelbauer: I guess getListParser should get the injector... (Is that something unit tests should have caught?)

ninja-core/src/main/java/ninja/bodyparser/BodyParserEnginePost.java
```diff
@@ -37,6 +43,13 @@
 
     private final Logger logger = LoggerFactory.getLogger(BodyParserEnginePost.class);
 
+	Injector injector;
```
raphaelbauer: this can be private

ninja-core/src/main/java/ninja/bodyparser/BodyParserEnginePost.java
jlannoy: You are right. I made a unit test for a user custom Date parser that needs a injector, but I forget to test Array and List parsers that would use that custom parser. I'll fix it tomorrow.

ninja-core/src/main/java/ninja/params/ParamParsers.java
```diff
@@ -53,61 +60,95 @@
...
-    private static final Map<Class<? extends Enum<?>>, ParamParser<?>> ENUM_PARSERS = Maps.newHashMap();
-
-    public static ParamParser<?> getParamParser(Class<?> targetType) {
-
+    private static final Map<Class<?>, Class<? extends ParamParser<?>>> PARAM_PARSERS_CLASSES = Maps.newHashMap();
+    
+    /**
+     * ALlows to register a new parser for given Class parameters.
+     */
+    public static <T> void registerParamParser(final Class<T> paramClass, final Class<? extends ParamParser<T>> parserClass) {
```
raphaelbauer: I'd like to have a big warning sign that registerParamParser must only be used at startup of Ninja. All other usages will not work in a multi threaded context with PARAM_PARSERS_CLASSES's HashMap. The same should be added for unregisterParamParsers.

Maybe that is another sign to rewrite that class to a non static class (Singleton). (I guess not part of this PR, just saying).

ninja-core/src/main/java/ninja/params/ParamParsers.java
```diff
@@ -53,61 +60,95 @@
...
                 // return CSV parser
                 return new CsvParamParser(targetType, componentParser);
             }
         }
-
-        if (ENUM_PARSERS.containsKey(targetType)) {
-            return ENUM_PARSERS.get(targetType);
+        
+        if (targetType.isEnum()) {
+        	return new GenericEnumParamParser((Class<Enum>) targetType);
```
raphaelbauer: there's a hidden tab :) should be 4 spaced ;)

raphaelbauer:
Hey @jlannoy , PR looks good. Very thoughtful and well done. The only thing that I have in my mind is rewriting ParamParsers from being static to being a @Singleton.

Injection of ParamParsers can then be done via https://github.com/google/guice/wiki/Multibindings . The additional benefit is that we can make sure that no user violates the contract of register / unregisterParamParser.

The disadvantage is that adding custom Param parsers is a bit more "complex":

```
   Multibinder<ParamParser> uriBinder = Multibinder.newSetBinder(binder(), ParamParser.class);
   uriBinder.addBinding().toInstance(new MyFunkyParamParser);
```

That's not something that you have to do I guess... I got some time and can do it on top of your PR if that is ok with you... Let me know what you think. Maybe I missed something...

ninja-core/src/test/java/ninja/bodyparser/BodyParserEnginePostTest.java
```diff
@@ -143,4 +293,15 @@ public void testBodyParserWithCollectionAndArray() {
 
     }
     
+    public static enum MyEnum {
+    	VALUE_A, VALUE_B, VALUE_C
```
raphaelbauer: there is a nasty tab hidden in the code ;)

ninja-core/src/site/markdown/documentation/basic_concepts/controllers.md
raphaelbauer, line 130: (It's not meant to be a critique of your PR). I added my "solution" to the discussion page. Would be great getting feedback on that...

jlannoy:
No problem, I completely agree on your comments. I didin't had a better solution in mind when I made the PR ; so I copied the static registerEnum() way of working. I'll check tomorrow for the MultiBindings. I'll also fix the other notes you made during your review.

jjlauer:
This is a cool PR.  Looking fwd to it being integrated. +1 on non-static /
multibinding solution if it works.

jlannoy:
Hi @raphaelbauer and @jjlauer.
I finally managed to do it with a _Multibinder<ParamParser>_. I also add tests for the collections and array parsers. 

But... It's not the end. I've discovered that the way I did it presents a conceptual error. Reusing the ParamParsers in the BodyParserEnginePost will allow to get parsing errors in the Validation object but.. Actually, they are treated as "FileViolations", not as "BeanViolations" as it should be, following the Validation object javadoc. So, how to solve this problem?

I can modify the ParamParser interface to add one method or one attribute to the current parsing method, allowing to know if it's for a bean of a field... Or we can rethink a little bit the Validation object. Is it really needed to separate the two? ("Field" is a little bit inappropriate (personal opinion), it should for example be "Parameter".) It's an open question, please give my your advice on the best way to handle that.

Jonathan

jlannoy:
Hi @raphaelbauer and @jjlauer.

It's well done now.
- All your comments have been fixed
- It works with the MultiBinder, and is documented as well
- I had to enhanced the Validation documentation page, to make the differences with the JSR303 validator clear (interpolated messages in bean violation // message keys in field validations)

Work done for me.

jjlauer:
Hey @jlannoy -- that's a pretty large commit.  Its a bit difficult to provide fresh feedback. Wondering if you tried/could rebase and force a fresh push to your branch so we can review your change in a single commit?  Here's one blog explaining the process a bit more: https://robots.thoughtbot.com/keeping-a-github-fork-updated

jlannoy:
I'll see when I can try to do it. Note that since @raphaelbauer's original review, not so much code has change. A lot of commits are due to merges with the develop branch (and conflicts one the changelog).

jlannoy:
(Rewrite modifications in a new branch)

PR #451, opened 2016-01-21
NinjaProperties available in main Module

jlannoy:
Allow the main module to access the NinjaProperties if needed. May be very useful for users that needs to do their own JPA initialization (to access other db properties for example). Backward
compatible.


jjlauer:
@jlannoy Have you looked at doing your initialization in services rather than directly in Module.java?  By using that approach you then get injected NinjaProperties via Guice rather than the method you proposed.  I'd be curious what all you're actually doing directly in conf/Module.java.  Just thinking there may be a cleaner way to do it than what your PR would allow.  Thx!

jlannoy:
Of course. But I think there are some cases where it's not possible ; when you want override some modules initialization done by default in Ninja. So it's not for services initialization. I agree with you that all what could be done via Guice should be done via Guice (especially with the @Name annotation, allowing to get only needed properties).

A concrete example now. Look at the NinjaJpaModule... It automatically takes your db connection properties and puts them... In hibernate specific properties. In a project, I had to import an existing persistence layer made with EclipseLink. It didn't cause any trouble, ninja can perfectly deal with an other actor that Hibernate. Except for the connection properties, that we put in the persistence.xml file at first. But having 4 or 5 environments, it couldn't stay there. I couldn't find an other solution that provide my own JPA module (based on Ninja's one), disabling the PU loaded automatically by ninja (not providing the properties) and doing the JPA initialization in my module. But, as you can figure it out now, to make it work I need there the NinjaProperties. And it cannot be injected at this time via Guice. 

Note that the Ninja JPA module do already take the NinjaProperties as constructor arg ; I didn't invent that need. Some users on the mailing lists have requested to use more than one database in their Ninja applications... Each time they have been answered to do their own JpaModule and to look at GuicePersistMultiModules. So they will encounter the same problem with the properties.

ninja-core/src/test/java/conf/Module.java
```diff
@@ -18,13 +18,26 @@
...
 // Allows to check that custom Ninja module in user's conf directory
 // works properly.
 public class Module extends AbstractModule {
 
-  
+	NinjaProperties ninjaProperties;
+	
+	public Module(NinjaProperties ninjaProperties) {
+		
+		// Will fail any test based on this Module if properties not received
```
raphaelbauer: That comment is not needed...

ninja-core/src/test/java/conf/Module.java
```diff
@@ -18,13 +18,26 @@
...
+		if(ninjaProperties == null) {
+			throw new IllegalArgumentException("null properties");
```
raphaelbauer: Please use Preconditions from Guava (checkNotNull or similar). And please print out a better error message. Just for the very rare case when something goes wrong this will help a lot to find the problem asap... :)

raphaelbauer:
I have no opinion on that... Up to you to decide... :)

ninja-core/src/test/java/conf/Module.java
jlannoy: Guava precondition or a better error message? I did choose the second one ; pay attention it's the Module defined for the unit tests only.

raphaelbauer: Yea - you are right it's only the test. I'd anyway appreciate a better failure message. Something like "ninjaProperties are null - something's really fishy here..." (or similar). At some point that stuff will be refactored - and then good error (or test failure) messages are gold...

raphaelbauer:
@jjlauer @jlannoy it right - JPAModule already gets NinjaProperties manually "injected" (Configuration.java down below). But maybe that's already an anti-pattern here.

The basic problem seems to be: Guice Modules are for configuration. NinjaProperties are for configuration. But because NinjaProperties are also created via Guice (not entirely true) we cannot use them in a clean way in Guice Modules that by default can not be used to inject stuff

One "solution" would be to make NinjaProperties a singleton (via Java enum singleton pattern or similar) so that it can be accessed everywhere in the application (and Guice). Not sure if that is a good idea... But that'd allow us to use NinjaProperties in all modules as well (without injecting them when calling new MyModule(ninjaProperties)). Of course that also means that NinjaProperties can not be mocked (with side-effects on testcases potentially).

I wanted to think a bit more about it. But it seems @jlannoy's solution is pragmatic and works best.

jjlauer:
@raphaelbauer Sounds good.  I think keeping NinjaProperties non-singleton
is the right approach.  And at a minimum you need to get some of those
properties in a Module.

raphaelbauer:
Just added the changelog. Thanks for your contribution :+1:

PR #453, opened 2016-01-24
Clarifieddefault db-config, enhanced document-flow

### LINKED ISSUE #445: Simplify Configuration within the JPA Documentation
http://www.ninjaframework.org/documentation/working_with_relational_dbs/jpa.html

The `Configuration` section mentions the need to configure the database.

But: The blog archetype uses an in-memory db by default (which is very nice so).

So, the minimal change for the documentation would be to add to the `Quickstart` section:

```
the application uses an in-memory db, so you don't need to configure a db. If you like to change to another db, like Postgresql orMySQL , see the `Configuration` section.
```

(Ideally, the `Configuration` section would be moved to the end of the document, thus the `Models` section comes first)

ghost:
closes #445

### DIFF
--- a/ninja-core/src/site/markdown/documentation/working_with_relational_dbs/jpa.md
+++ b/ninja-core/src/site/markdown/documentation/working_with_relational_dbs/jpa.md
@@ -21,128 +21,14 @@ mvn ninja:run
 </pre>
 
 
-Configuration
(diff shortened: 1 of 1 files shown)


raphaelbauer:
Nice :+1:

Issue #454, opened 2016-01-24
Some General Issues Subjecting Documentation


ghost:
I have some suggestions in order to make the documentation a bit more efficient. Please see if those changes are likely to be accepted. Please just take a look and let me know:
- [x] **ADVANCED TOPICS** should get more subsections, in order to shrink (19 entries are really too much)
- [ ] Avoid long menu entries (like **Working with XML**, should become **XML** thus essence is immediately seen)
- [x]  **COMMERCIAL SUPPORT** should move after "GREAT COMMUNITY"
- [ ]  The **Ninja - high level overview** should become graphical (very important, and should be done by the core devs). Additionally rename to **Architecture** or another concise single-word term.
- [ ] The **INTRODUCTION** should start with a **Quick Start** , code & results should be the first thing a visitor sees.
- [ ]  The **Quick-Start** should showcase a **Blog in 15 minutes** (with JPA, and if possible User-Login), demonstrating immediately the "change-something and see it immediately" and other benefits of ninja.
- [ ]  The index page should possibly contain right after **Rock solid, fast, and super productive** something like this: Can't wait? Get a blog app in 20 min <link to Quick Start>

raphaelbauer:
That makes a lot of sense. You are a born marketing genius :)

Commercial support will become a separate site anyway. So there will be support => community support (so, mailing list, github) and commercial support (separate site).

lazaridis-com:
Nah, I just looked at hundreds of sites describing frameworks, and thus I know quite good what works well.

I'm still missing a framework where the website introduces you with a simple app (e.g. blog or issue-tracker) into the framework, and then teaches you how to extend this, whilst introducing further top-level functionality.

Now, anything special I need to know in order to change the navigation of the docs?

lazaridis-com:
(it seems the website gets not updated. But I need to see the changes, in order to take the next steps)

lazaridis-com:
@raphaelbauer , I don't understand. I've asked here prior to making changes, in order to save my time and your time.

Now I saw that you don't merge the change within  #455, which is essential in order to make things better.

"Commercial Support" does **not** belong in a *prominent position* within the menu. It's simply too much, the visitor sees 2 times "Commercial Support", before seeing any code.

You don't need to be a genius to understand this.

Either you delegate some work, or not, but I cannot discuss every tiny (and obvious) change.

raphaelbauer:
[...]

PR #455, opened 2016-01-25
re #454, move "Commercial Support" to end

ghost:


raphaelbauer:
Hey lazaridis... That's something we'll handle separately... Commercial support has to stay in a prominent position...

PR #458, opened 2016-01-27
re #454, reduced/rearranged root menu entries

ghost:
Note that this must not be the valid final result. The overall tree needs to be restructured, which is an incremental process.

### DIFF
--- a/ninja-core/src/site/site.xml
+++ b/ninja-core/src/site/site.xml
@@ -74,6 +74,8 @@
                 <item name="Flash scope" href="./documentation/basic_concepts/flash_scope.html"/>
                 <item name="Error handling" href="./documentation/basic_concepts/error_handling.html"/>
                 <item name="Metrics" href="./documentation/basic_concepts/metrics.html"/>
+                <item name="Logging" href="./documentation/logging.html"/>                
+                <item name="Sending mail" href="./documentation/sending_mail.html"/>
             </item>
 
             <item name="Configuration and modes" href="./documentation/configuration_and_modes.html"/>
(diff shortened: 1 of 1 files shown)


ninja-core/src/site/site.xml
```diff
@@ -89,34 +90,29 @@
...
-            <item name="Sending mail" href="./documentation/sending_mail.html"/>
-
-            <item name="Logging" href="./documentation/logging.html"/>
-
-            <item name="Custom routing" href="./documentation/custom_routing.html"/>
-
-            <item name="Servlet bridge" href="./documentation/servlet_bridge.html"/>
-
+            <item name="Debugging" collapse="true"href="./documentation/debugging.html">
+                <item name="Scheduler" href="./documentation/scheduler.html"/>
```
raphaelbauer, line 109: I guess Debugging should not be a top-section of Scheduler, Lifecycle and so on...

ghost, line 109: As I mentioned "The overall tree needs to be restructured, which is an incremental process."

Next step would most possibly be:
- either "Debugging" menu entry is renamed to "Development", or a new document is introduced as top-level, named "General Development" (or similar).

lazaridis-com:
@raphaelbauer 

Please, have the decency (against users and interessents, not me) to avoid rejecting contributions like a stubborn kid. This is essentially a suggestion, or a question:

> either "Debugging" menu entry is renamed to "Development", or a new document is introduced as top-level, named "General Development" (or similar).

If you get exited by micromanaging people which can fulfill a task better than you, please just decide which way to go (rename label, or introduce new Development text).

Otherwise just check things in, and the followups will come.

#454 contains the minimum steps neccessary to make the docu at least a little bit better.

Hopefully you are aware that in the current condition, the documentation does not allow a decent teach-in to Ninja (that's why I'm trying to enhance it, thus it becomes usable for evaluation).

PR #459, opened 2016-01-29
Session and flash data save to Context instead of Result

jjlauer:
Based on issue #450 it was uncovered that session and flash data was being saved to the Result then copied to the Context then written to the underlying response stream.  This is not a problem except in the case where a user accidentally/intentionally reuses a Result object across requests.  In that case the session and context data had a race condition / security hole.

This PR skips writing the session and flash data to the Result and instead directly saves it to the Context.     It also saves a few bytes of ram by eliminating an unnecessary extra array copy.  The Result has been entirely eliminated from the session and flash interfaces to prevent future changes to it.

### DIFF
--- a/ninja-core/src/main/java/ninja/session/FlashScope.java
+++ b/ninja-core/src/main/java/ninja/session/FlashScope.java
@@ -36,7 +36,7 @@ public interface FlashScope {
 
     void init(Context context);
 
-    void save(Context context, Result result);
+    void save(Context context);
 
     void put(String key, String value);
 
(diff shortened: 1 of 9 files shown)


raphaelbauer:
Nice one. Makes stuff a lot cleaner! How should we proceed regarding #450? I'd like to cut a release with a fix...

jjlauer:
Not sure on #450 what the right move is.  While removing .with() in routes solves one issue, who is to say a user doesn't try to reuse Results in other places in the app?  I was thinking we could make a Result 1-time use or add a "freeze" feature so you can't modify anything in it.  The "freeze" feature means you could keep the .with() method.

raphaelbauer:
That's why I proposed to make Result immutable. It would remove a whole class of problems and workarounds we'd have to apply...

Did I miss something or does your PR fix the security issue anyway? As long as we don't have any global filters and such the reused result is not a problem.

PR #460, opened 2016-01-29
Unit test for session bug with single instance Result

jjlauer:
Unit test for #450 using ninja-standalone (easiest since it requires so much interaction between components)


raphaelbauer:
Perfect.

Issue #462, opened 2016-02-08
Cookie domain is not set when clearing session


cbxp:
Currently, when calling `Session#clear`, cookie domain is not set when it was set previously, thus cookie is not deleted in the client.

Here's how to reproduce it:

1) Configure cookie domain

```
# conf/application.conf
application.cookie.domain=cookie-domain.com
```

2) Create a session

``` java
session.put("foo", "bar");
```

3) Clear session

``` java
session.clear();
```

Expected:
Cookie is deleted in the browser.

Actual:
Cookie is not deleted because cookie domain is not set when "clearing" cookie.

We are using currently version `5.2.2`, but it also seems to exist in master (didn't verify yet): https://github.com/ninjaframework/ninja/blob/develop/ninja-core/src/main/java/ninja/session/SessionImpl.java#L223-L235

raphaelbauer:
Very nice report - that's absolutely true and a bug! We'll fix that asap...

raphaelbauer:
Hey @cbxp - could you check that https://github.com/ninjaframework/ninja/pull/463 fixes the issue? Thanks!

PR #463, opened 2016-02-08
Fix Cookie domain is not set when clearing session #462

raphaelbauer:


jjlauer:
This looks a lot cleaner using a common method to create a cookie.  Looks good to me.

PR #464, opened 2016-02-09
IllegalArgumentException in ninja.utils.AbstractContext.java

matthias-burgfried:
The goal of this pull request is to catch an IllegalArgumentException in  ninja.utils.AbstractContext.java.

An invalid URI (e.g. http://www.example.com/search/berlin|t3843egel372f) causes an uncatchable exception in getPathParameter(String key).

```
ERROR [qtp3447021-53] n.NinjaDefault [NinjaDefault.java:232] Emitting bad request 500. Something really wrong when calling route: /search/berlin|t3843egel372f/26 (class: class 
controllers.SearchController method: public ninja.Result controllers.SearchController.ergebnis(java.lang.String,int))
java.lang.IllegalArgumentException: Illegal character in path at index 6: berlin|t3843egel372f
        at java.net.URI.create(URI.java:852) ~[na:1.8.0_72]
        at ninja.utils.AbstractContext.getPathParameter(AbstractContext.java:191) ~[www-1.0-SNAPSHOT.jar:na]
        at ninja.params.ArgumentExtractors$PathParamExtractor.extract(ArgumentExtractors.java:130) ~[www-1.0-SNAPSHOT.jar:na]
        at ninja.params.ArgumentExtractors$PathParamExtractor.extract(ArgumentExtractors.java:121) ~[www-1.0-SNAPSHOT.jar:na]
        at ninja.params.ControllerMethodInvoker.invoke(ControllerMethodInvoker.java:54) ~[www-1.0-SNAPSHOT.jar:na]
        at ninja.FilterChainEnd.next(FilterChainEnd.java:49) ~[www-1.0-SNAPSHOT.jar:na]
        at ninja.NinjaDefault.onRouteRequest(NinjaDefault.java:102) ~[www-1.0-SNAPSHOT.jar:na]
        at ninja.servlet.NinjaServletDispatcher.service(NinjaServletDispatcher.java:86) [www-1.0-SNAPSHOT.jar:na]
        at com.google.inject.servlet.ServletDefinition.doServiceImpl(ServletDefinition.java:287) [www-1.0-SNAPSHOT.jar:na]
        at com.google.inject.servlet.ServletDefinition.doService(ServletDefinition.java:277) [www-1.0-SNAPSHOT.jar:na]
        at com.google.inject.servlet.ServletDefinition.service(ServletDefinition.java:182) [www-1.0-SNAPSHOT.jar:na]
        at com.google.inject.servlet.ManagedServletPipeline.service(ManagedServletPipeline.java:91) [www-1.0-SNAPSHOT.jar:na]
        at com.google.inject.servlet.FilterChainInvocation.doFilter(FilterChainInvocation.java:85) [www-1.0-SNAPSHOT.jar:na]
        at com.google.inject.servlet.ManagedFilterPipeline.dispatch(ManagedFilterPipeline.java:119) [www-1.0-SNAPSHOT.jar:na]
        at com.google.inject.servlet.GuiceFilter$1.call(GuiceFilter.java:133) [www-1.0-SNAPSHOT.jar:na]
        at com.google.inject.servlet.GuiceFilter$1.call(GuiceFilter.java:130) [www-1.0-SNAPSHOT.jar:na]
        at com.google.inject.servlet.GuiceFilter$Context.call(GuiceFilter.java:203) [www-1.0-SNAPSHOT.jar:na]
        at com.google.inject.servlet.GuiceFilter.doFilter(GuiceFilter.java:130) [www-1.0-SNAPSHOT.jar:na]
        at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1652) [www-1.0-SNAPSHOT.jar:na]
        at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:585) [www-1.0-SNAPSHOT.jar:na]
        at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1127) [www-1.0-SNAPSHOT.jar:na]
        at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:515) [www-1.0-SNAPSHOT.jar:na]
        at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1061) [www-1.0-SNAPSHOT.jar:na]
        at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:141) [www-1.0-SNAPSHOT.jar:na]
        at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:97) [www-1.0-SNAPSHOT.jar:na]
        at org.eclipse.jetty.server.Server.handle(Server.java:497) [www-1.0-SNAPSHOT.jar:na]
        at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:310) [www-1.0-SNAPSHOT.jar:na]
        at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:257) [www-1.0-SNAPSHOT.jar:na]
        at org.eclipse.jetty.io.AbstractConnection$2.run(AbstractConnection.java:540) [www-1.0-SNAPSHOT.jar:na]
        at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:635) [www-1.0-SNAPSHOT.jar:na]
        at org.eclipse.jetty.util.thread.QueuedThreadPool$3.run(QueuedThreadPool.java:555) [www-1.0-SNAPSHOT.jar:na]
        at java.lang.Thread.run(Thread.java:745) [na:1.8.0_72]
Caused by: java.net.URISyntaxException: Illegal character in path at index 6: berlin|t3843egel372f
        at java.net.URI$Parser.fail(URI.java:2848) ~[na:1.8.0_72]
        at java.net.URI$Parser.checkChars(URI.java:3021) ~[na:1.8.0_72]
        at java.net.URI$Parser.parseHierarchical(URI.java:3105) ~[na:1.8.0_72]
        at java.net.URI$Parser.parse(URI.java:3063) ~[na:1.8.0_72]
        at java.net.URI.<init>(URI.java:588) ~[na:1.8.0_72]
        at java.net.URI.create(URI.java:850) ~[na:1.8.0_72]
```

### DIFF
--- a/ninja-core/src/main/java/ninja/utils/AbstractContext.java
+++ b/ninja-core/src/main/java/ninja/utils/AbstractContext.java
@@ -15,8 +15,10 @@
  */
 package ninja.utils;
 
+import java.io.UnsupportedEncodingException;
 import java.net.InetAddress;
 import java.net.URI;
+import java.net.URLEncoder;
 import java.net.UnknownHostException;
 
 import ninja.bodyparser.BodyParserEngine;
@@ -188,7 +190,22 @@ abstract public class AbstractContext implements Context.Impl {
         if (encodedParameter == null) {
             return null;
         } else {
-            return URI.create(encodedParameter).getPath();
+        	try {
+        		return URI.create(encodedParameter).getPath();
+        	} catch (IllegalArgumentException e) {
+        		StringBuilder sb = new StringBuilder();
+        		for (String part : encodedParameter.split("/")) {
+        			try {
+						sb.append(URLEncoder.encode(part, "UTF-8")).append("/");
+					} catch (UnsupportedEncodingException e1) {
+						// TODO: catch
+					}
+        		}
+        		if(sb.length() > 0 && !encodedParameter.substring(encodedParameter.length()-1).equals("/")) {
+        			sb.setLength(sb.length()-1);
+        		}
+        		return sb.toString();
+        	}
         }
     }
 


raphaelbauer:
Hi Matthias,

thanks for the PR and the report. I'd assume Ninja is doing the right thing here. We decode the path parameter according to RFCs. If that fails Ninja logs the error.
In my opinion the sender is responsible for correctly encoding the url (eg when I enter the url given in Chrome I get http://www.example.com/search/berlin%7Ct3843egel372f which is correct).

What is your opinion? - I may be wrong here, because url encoding gets very complex very fast...

Cheers,

Raphael

matthias-burgfried:
You're right, but some bots/senders do not encode all symbols as they should. In particular the pipe seems problematic. The error above is caused by google translate.
`108.162.xxx.xxx - - [09/Feb/2016:00:03:55 +0100] "GET /search/berlin%7Ct38364143egel/13 HTTP/1.1" 500 1550 "https://translate.googleusercontent.com/translate_p?hl=en&prev=search&sl=de&u=http://www.example.com/search/berlin%257Ct38364143egel/seite13.html&depth=1&rurl=translate.google.com.ng&usg=ALkJrhgAAAAAVrpxeQncj1H48y2R_ax9Exyl7TzBfcao" "Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2638.0 Safari/537.36,gzip(gfe)"`

Is therefor a way to catch the exception? I tried `getInternalServerErrorResult()` with no luck.

Best,
Matthias

matthias-burgfried:
Any suggestions how to catch the exception?

Matthias

jfendler:
@matthias-schulz you can provide your own `conf.Ninja` class by overwriting `NinjaDefault`, and then provide your own `getInternalServerErrorResult(Context context, Exception exception)` method to handle the exception as you please.

However, @raphaelbauer, I would support a slightly different handling of this (and similar) situations. In this case, the pattern for the `PathParam` seems to match, only the values cannot be parsed into the necessary type. What would be nice in this case is for the original controller method to still be called (with null values in place of the invalid params), if the developer wants this.

At a first glimpse I would see the following (more or less easy to implement) option:
Catch any exceptions from the call to 
`arguments[i] = argumentExtractors[i].extract(context);`
in `ControllerMethodInvoker.invoke(Object,Context)`
and handle exceptions by a) logging the exception, and b) setting the respective `arguments[i]` to null. Then go on to call the controller method as if the parameter was parsed correctly (passing the null value).
As far as I understand, a controller with (non-primitive) PathParam types is currently guaranteed to receive non-null values. If I am correct with this assumption, that means that this approach would possibly break existing applications, where controllers currently rely on non-null parameters. Such could be addressed by e.g. introducing a new boolean parameter "nullable" to the `PathParam` (and probably `Param`) annotation, defaulting to "false", which would explicitly allow the above described handling if set to "true". By default the existing Ninja behaviour would remain as-is.

This would provide a bit more flexibility when dealing with possibly-erroneous parameters, without the need to handle all the different exceptions for the whole application with `NinjaDefault`.

What do you think? Worth starting some work on a PR?

raphaelbauer:
@jfendler Thanks for all the good comments... I will probably come back to that... I wanted to first understand the root cause of the problem... Encoding is always a bit tricky and we want to get it "right" in the first place...

@matthias-schulz In your example with Google Translate all the encodings are correct (no raw pipe symbols in the GET part...) ... So I am wondering: How can I reproduce the problem in the real world with Google Translate?

```
108.162.xxx.xxx - - [09/Feb/2016:00:03:55 +0100] "GET /search/berlin%7Ct38364143egel/13 HTTP/1.1" 500 1550 "https://translate.googleusercontent.com/translate_p?hl=en&prev=search&sl=de&u=http://www.example.com/search/berlin%257Ct38364143egel/seite13.html&depth=1&rurl=translate.google.com.ng&usg=ALkJrhgAAAAAVrpxeQncj1H48y2R_ax9Exyl7TzBfcao" "Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2638.0 Safari/537.36,gzip(gfe)"
```

jfendler:
@raphaelbauer I don't think Ninja is doing anything wrong with respect to encodings. In fact, I don't think it's doing anything "wrong" as such. The question to me is where exceptions from invalid parameters (as produced on the client side) can usefully be handled. Obviously, the Ninja catch-all approach is necessary, but should be a last resort in my opinion. Yet trying to handle these errors as close to the code which was intended to handle similar, but _correct_ requests (i.e. in the respective controller method Ninja already knows was meant to serve the request) would probably be a good thing to have.

raphaelbauer:
@jfendler @matthias-schulz There is one thing that can cause this behaviour: Ninja is relying on a proper Servlet spec in order for this to work. If the underlaying Servlet engine already decodes the path (which is not according to spec) then we'll run into this problem. And that's the real bug we should then report imho...

@matthias-schulz How do you run your Ninja application? Is it running in the default Jetty we are packaging or any special servlet container as war? (And: Which versions of the software are you using?)

matthias-burgfried:
@raphaelbauer 
I'm running the default jetty and using the current version (5.3.1).
This is a Google Translate URL which causes an error: [Click](https://translate.google.com/translate?sl=en&tl=de&js=y&prev=_t&hl=de&ie=UTF-8&u=www.berlinerfirmen.info%2Fsuchen%2Fberlin|t35f5egel3719%2F104&edit-text=&act=url)
@jfendler 
I will try `getInternalServerErrorResult(Context context, Exception exception)` and report the results here.

matthias-burgfried:
@jfendler 
That is not working either.
I tried

```
package conf;

import ninja.Context;
import ninja.NinjaDefault;
import ninja.Result;
import ninja.Results;

public class Ninja extends NinjaDefault {

    @Override
    public void onRouteRequest(Context.Impl context) {
        System.out.println("onRouteRequest: " + context.getRequestPath());
        super.onRouteRequest(context);
    }

    @Override
    public Result getInternalServerErrorResult(Context context, Exception exception) {

        System.out.println("getInternalServerErrorResult: " + context.getRequestPath());

        Result r = Results.redirect("/");
        return r;

    }

}
```

All I get:
`onRouteRequest: /suchen/berlin|t3843egel372`
`getInternalServerErrorResult(Context context, Exception exception)` is not called.

jfendler:
@matthias-schulz You are right. Overriding `getInternalServerErrorResult(Context context, Exception exception, Result underlyingResult)` instead should do the trick for you.

matthias-burgfried:
@jfendler 
Perfect, everything works properly.

raphaelbauer:
@matthias-schulz thanks for the info... good to know that Jetty does not cause the behavior :D ...
another question: how to you create the url for google translate? do you create the url yourself? or is this a button that automagically transforms the url?

jfendler:
@matthias-schulz the error is strange though. When I use the following on the URL in the address bar to target a String parameter (as copied from your link above) in a controller, everything works as expected:
`berlin%7Ct35f5egel3719`
No problem, no exception I can see.

Of course, Firefox (and most other browsers I would believe) immediately render this as
`berlin|t35f5egel3719` in the addres bar. As long as I just refresh with F5, everything is still ok. But once I copy the URL from the address bar paste it again, the URL encoding is lost and the error pops up (as it should for now).

Edit: you do log already - what do you see in your logs when using the link you posted above?

jfendler:
@raphaelbauer I agree - if a Servlet engine is at fault, this should be fixed. However, Jetty appears to be not at fault, and I've never had any such issues with Tomcat either. I would assume such error would pop up in a very early test case of every engine.
I still believe in Matthias' case the client is at fault - or whatever component is creating the URL. The example link is correctly encoded and works fine, yet the only log output he provided indicates the pipe character NOT being encoded. For me this only happens when the incoming request is already broken.

But apart from that, the question remains if we can (should?) allow null values (following exceptions) for parameters. The controller could then decide if it can recover or not.

@matthias-schulz did you try to put `berlin%7Ct35f5egel3719` manually in the address bar to test?

matthias-burgfried:
@raphaelbauer 
Visit the Google Translate website ([Google Translate](https://translate.google.com/)) put an URL into the text area and click on the link which appears on the right hand. To answer your question: I do not create the URL by myself. Google does the "magic".
@jfendler 
Should I log something for you or is this a general hint for me?

matthias-burgfried:
Mhh, `berlin%7Ct35f5egel3719` causes the application to loop infinitely...
I think, I must rework my patch.

jfendler:
@matthias-schulz I believe whenever the error occurs, Ninja does NOT receive a properly encoded pipe symbol. You can log this to verify, but actually there's no need as it is obvious from the error.
However, in the link you provided, the encoding is correct. Is there anything else between your client (when following this link) and the Ninja application? If we assume Jetty is not the culprit, perhaps a proxy or a reverse proxy? Try and enable logging on each component in between and see where the encoding is messed up.

As to your "application-loop",  this is probably another bug in your application which might have nothing to do with the encoding. ;-) Perhaps just use any simple controller with a single String-type path parameter (instead of the "real thing") for getting to the root of the original problem first..

matthias-burgfried:
@jfendler 
Yes, the "application-loop" is my fault.
The encoding error occurs, even if I hit the ninja application directly. 
I tried `localhost:8080/suchen/berlin|t3843egel372` after running `mvn ninja:run`. Same error.

jfendler:
@matthias-schulz yes, this is the expected behaviour for a NOT-properly encoded parameter. (You are entering the pipe character directly.) Please try it as `%7C` instead. I guess it will work fine then.

matthias-burgfried:
Yes, if I try `%7C` instead of `|` it works as expected. Perhaps it is better to improve the documentation and drop the PR. ;)
@raphaelbauer and @jfendler 
Thank you very much for pointing me to the right direction.

raphaelbauer:
Hey Matthias... if you enter the url in google translate make sure you use the following url: 
www.berlinerfirmen.info/suchen/berlin%7Ct35f5egel3719/104

You have to "double" encode the | symbol... This may well be a problem of Google translate btw... I'd report it...

PR #465, opened 2016-02-15
Graceful handling of Jetty EofException when client closes connection.

jfendler:
- This commit addresses the issue described in https://groups.google.com/forum/#!topic/ninja-framework/z12l8Izh7EM
- EofExceptions are logged at INFO level, but the exception is not wrapped/thrown further.
- No hard dependency to Jetty container (Exception type is evaluated by exception class name). Similar exceptions as may be thrown by other containers could be handled in the same way.

### DIFF
--- a/ninja-core/src/main/java/ninja/Result.java
+++ b/ninja-core/src/main/java/ninja/Result.java
@@ -381,10 +381,20 @@ public class Result {
                     writer.write(string);
                     
                 } catch (IOException ioException) {
-                
-                    logger.error(
-                            "Error rendering raw String via renderRaw(...)", 
-                            ioException);
+                    Throwable cause = ioException.getCause();
+                    if ( cause != null && "org.eclipse.jetty.io.EofException".equals(
+                            cause.getClass().getName())) {
+                        // Jetty indicating that the client closed the stream.
+                        // Let's be nice about this and not throw an exception.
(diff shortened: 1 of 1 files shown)


jjlauer:
@jfendler I sympathize w/ annoying exceptions from the container.  I'm not sure you're approach is elegant.  For example, quite a few lines of code simply copy n' pasted.  While your solution doesn't rely on Jetty for compilation, it is pretty brittle at the same time.  Are you using `ninja-standalone` -- which does compile against Jetty?  Perhaps a more elegant solution is to provide some sort of handling in `ninja-standalone` which can compile against Jetty.  I can't give you much direction on what that could be, but perhaps some sort of `Provider` or `NinjaDefault` that is bound by Guice that can decide if exceptions are worthy of being logged by default.

jfendler:
@jjlauer well, I _am_ quite sure that it's not elegant. :-) At this point it's more of a quick-and-dirty fix rather than a fully thought through implementation. However, I made this adjustment for one of my own projects some time ago, and thought I'd just leave it here.

I like the idea of a guice component which would determine (with some user configuration options) the "severity" of an exception and log/handle it accordingly. I'll think about and might come up with something.

Either way, it's more of a little itch than an issue - I might take a while to get into this..

PR #470, opened 2016-02-19
Add ninja undertow module to docs

jjlauer:
Standalone implementation for the using Undertow.  Ninja-undertow does _not use servlets_ under-the-hood -- therefore it bypasses a significant amount of code that Ninja's default Jetty-based standalone uses.  Ninja-undertow can fully replace Ninja's Jetty-based standalone.

Based on this [benchmark](src/test/java/ninja/undertow/Benchmarker.java) ninja-undertow is 15.5% faster than ninja-jetty for standard GET requests and 7.6% faster than ninja-jetty for POST requests w/ JSON.  Future optimizations and tuning should only improve performance.


raphaelbauer:
Nice!

PR #473, opened 2016-02-26
Added RecycledNinjaServerTester for speedy unit tests

jjlauer:
Added support for new junit testers -- most notable is the RecycledNinjaServerTester which uses a single Ninja server across all unit tests in a class.  If you have any number of unit tests in a class, using this approach can help speed up your unit tests by an order of magnitude.  For example, updated Controller test in servlet-integration runs 4x faster than before.


raphaelbauer:
Pretty cool!

PR #474, opened 2016-02-26
Removed ninja-core dependency on org.mindrot:bcrypt (it was unused)

jjlauer:
This dependency was sneaking its way into ninja-core even though its not used.  It is only used in the servlet integration test -- which had this dependency added.


raphaelbauer:
:+1:

PR #475, opened 2016-02-27
Remove ehcache type use in generic ninja cache interface

jjlauer:
I was trying to remove my dependency of ehcache the other day and found out I couldn't.  Turns out NinjaCache accidentally was using an ehcache CacheException to throw a simple exception.


raphaelbauer:
:+1:

PR #476, opened 2016-02-29
More flexible guice configuration!

jjlauer:
This PR provides much more flexible configuration of guice -- including a new method for applications to entirely skip Ninja's "classic" stack of freemarker, ehcache, etc.

**Overview**
- All changes should be backwards compatible
- Your `conf.Module` can now optionally extend `ninja.conf.FrameworkModule` to skip Ninja’s default guice bindings for its “classic” stack of Freemarker, Jackson, Cache, etc. Example is below.
- New `ninja.conf.NinjaClassicModule` to let power users configure the stack they want.
- Significantly cleaned up `cache.CacheProvider` and `postoffice.PostofficeProvider` and added more unit tests.  Removed hard-coded references to implementations such as Ehcache so that it's dependency could actually be excluded if its not used.
- Added configurable implementation for `migrations.MigrationEngineProvider` similar to cache and postoffice.
- Body parser and template engine managers no longer hard-code themselves to freemarker, json, xml, etc. Those are now bound with guice configuration.

**Advanced application configuration example**

``` java
package conf;

import ninja.conf.FrameworkModule;
import ninja.conf.NinjaClassicModule;
import ninja.utils.NinjaProperties;

public class Module extends FrameworkModule {

    private final NinjaProperties ninjaProperties;

    public Module(NinjaProperties ninjaProperties) {
        this.ninjaProperties = ninjaProperties;
    }

    @Override
    protected void configure() {
        // classic ninja stack but with no freemarker or xml
        install(new NinjaClassicModule(ninjaProperties)
            .freemarker(false)
            .xml(false)
        );
    }

}
```


ninja-core/src/main/java/ninja/conf/NinjaClassicModule.java
```diff
@@ -0,0 +1,165 @@
...
+import ninja.template.TemplateEngineJsonP;
+import ninja.template.TemplateEngineText;
+import ninja.template.TemplateEngineXml;
+import ninja.utils.NinjaProperties;
+
+/**
+ * The classic configuration of the ninja framework (jackson, freemarker, 
+ * postoffice, etc.)
+ */
+public class NinjaClassicModule extends AbstractModule {
```
raphaelbauer, line 49: That's clever...

ninja-core/src/main/java/ninja/cache/CacheProvider.java
```diff
@@ -25,84 +27,51 @@
...
-      
+        });
     }
 
     @Override
     public Cache get() {
-        // only called once => reference cached by guice...
-        return cache;
-        
+        return this.supplier.get();
```
raphaelbauer, line 74: Do we really need a supplier here? I actually thought Guice is managing that for us...

ninja-core/src/main/java/ninja/utils/ImplFromPropertiesFactory.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * An abstract provider that determines which implementation to load based on 
+ * the value of key in {@link NinjaProperties} (aka application.conf).
+ * 
+ * If this variable is set then the implementation class is loaded for subclasses
+ * to use in their own implementation of Provider.get().
+ */
+public class ImplFromPropertiesFactory<T> {
```
raphaelbauer, line 30: That's also nice...

raphaelbauer, line 30: There is no test for this class... afaics...

raphaelbauer:
Very clever way of making the framework more configurable :+1: (Just a few questions, nothing dramatic...)

ninja-core/src/main/java/ninja/cache/CacheProvider.java
jjlauer, line 74: I had thought so as well.  The problem is that Cache, Postoffice, etc. should be Singletons.  Guice does not allow Singletons to be lazily built on-demand.  So let's say you don't use Cache anywhere in your app -- Guice would still build it & start it due to be being bound as a singleton.  Postoffice is worse since it validates you have a config key of 'smpt.host' set, but even if you don't use it in your app, your app won't start.  So the solution was to not bind them as Singleton, and then lazily create a singleton instance with a supplier.  If you then inject a Cache or Postoffice into your controller, it'll fail on start, but if you never do then they are essentially ignored and never built.

ninja-core/src/main/java/ninja/utils/ImplFromPropertiesFactory.java
jjlauer, line 30: Since it's used in 3 places that all have unit tests, I skipped including one.  I'll create a simple one today though just to be safe.

PR #477, opened 2016-02-29
New AutoStandalone class to find a standalone for running and unit tests

jjlauer:
With my work on [Ninja-Undertow](https://github.com/fizzed/ninja-undertow), I've been finding it a bit repetitious to become the default standalone in my projects.  Thus, `AutoStandalone` is a new feature where Ninja will find the standalone to use based on a few rules.  First, the current method of system property 'ninja.standalone.class' still is the first preference.  Then, META-INF/services via java.util.ServiceLoader will be used.  Finally, it'll fallback to jetty.

With this change, to pickup undertow (or any external standalone, its as simple as adding it to the classpath.  The caveat is that multiple META-INF/services on the classpath will mean the first one wins, but that's still workable since the user can explicitly then set a system property.  NinjaJetty does not have a META-INF/services since it's still the default.

### DIFF
--- a/ninja-core/src/main/java/ninja/utils/ForwardingServiceLoader.java
+++ b/ninja-core/src/main/java/ninja/utils/ForwardingServiceLoader.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2016 ninjaframework.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


ninja-core/src/main/java/ninja/utils/ForwardingServiceLoader.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.Iterator;
+import java.util.ServiceLoader;
+
+/**
+ * Forwards method calls to an underlying java.util.ServiceLoader -- mainly to
+ * allow participating and mocking in unit tests.  Designed to be a drop-in
+ * replacement at the source-code level since java.util.ServiceLoader is final.
+ */
+public class ForwardingServiceLoader<T> {
```
raphaelbauer, line 26: :+1:

raphaelbauer:
Nice implementation! Feel free to merge anytime :)

PR #478, opened 2016-02-29
Faster reverse routing; support for new MethodReference type

jjlauer:
This PR implements two main changes:
- Reverse routing now uses a pre-calculated hashmap for O(1) reverse lookups vs. the current O(N) complexity of iterating thru routes
- RouteBuilder and Router accept a new `ninja.utils.MethodReference` type -- which simply includes a Class and method name.  This is not only used for the reverse routing key in the hashmap, but it also will enable [ninja-java8](https://github.com/fizzed/ninja-java8) to supply compile-time checked routes!  Even refactoring controller or method names in your IDE will change them for routing too :-)  Once this PR is accepted, I'll be able to modify my ninja-java8 code to support something like the following.

```
router.GET().route("/index").with(methodRef(Application::Index));
```

methodRef could be a static method that returns a type of `ninja.utils.MethodReference`

### DIFF
--- a/ninja-core/src/main/java/ninja/utils/MethodReference.java
+++ b/ninja-core/src/main/java/ninja/utils/MethodReference.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2016 ninjaframework.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


ninja-core/src/main/java/ninja/utils/MethodReference.java
```diff
@@ -0,0 +1,73 @@
...
+package ninja.utils;
+
+import java.lang.reflect.Method;
+import java.util.Objects;
+
+/**
+ * Compound key of a Class and the name of a method in it. Primarily used for
+ * reverse route lookups.
+ */
+public class MethodReference {
```
raphaelbauer, line 25: Do we need tests for that class?

raphaelbauer:
Looks good to me!

Ultimately I'd like to have something like 

```
router.GET().route("/index").with(Application::Index);
```

jjlauer:
You can get:

router.GET().route("/index").with(Application::Index);

But you'll need to add an interface for every number of arguments a method
could define.  So, you could do that for up to say 10 variables, and if you
need more than that, then use a helper method for us to define more -- or
fallback to class + method name.

Issue #494, opened 2016-04-03
Risky error information in prod mod


danielsawan:
Hi there,

I have an nginx serveur in front of my Ninja app and when i receive some DDOS i noticed that there is some sensible information that are shown : http://puu.sh/o4Lug/7b62ebb3a0.png

I have no controle of that and i think we should be able to configure a template for all non ninja error.
The "hacker" here have the information that i am using JAVA and that i have some problem to connect to my database :-1:

raphaelbauer:
How do you connect to your db - and how is it configured?

danielsawan:
The database is on another server then the Ninja APP.
I am connecting to the database using the "normal" way described in Ninja docs.
The only things is that i am using MySQL compression to compress the trafic between my Ninja App and my MySQL database.

raphaelbauer:
That seems to happen when the error occurs outside of what Ninja can control.

There are two things you can do:
1) Add error-pages to web.xml and define an url that should be used to handle your error (http://www.eclipse.org/jetty/documentation/current/custom-error-pages.html).
2) Add a custom error handler to Jetty's startup, so we can handle those deep errors with Ninja... (https://virgo47.wordpress.com/2015/02/07/jetty-hardening/).

1) should be easy for you to do immediately. 2) Is something we should think about adding to Ninja's Jetty implementation.

danielsawan:
Thanks for your help i will have a look.

raphaelbauer:
Hey @danielsawan - can you check whether https://github.com/ninjaframework/ninja/pull/495 fixes your problem? Thanks!

danielsawan:
I answer on the pr.

raphaelbauer:
Houskeeping here. Please re-open the issue if you think it is still valid.

PR #495, opened 2016-04-04
Fix for #494. Suppress chatty Jetty error messages.

raphaelbauer:
Jetty by default prints out parts
of the stacktrace if an error occurs that is not
under control of Ninja.

In those rare occasions we replace the default error
handler with a silent error handler, that
just logs the problem and informs the user in
a very basic way.

### DIFF
--- a/ninja-core/src/site/markdown/developer/changelog.md
+++ b/ninja-core/src/site/markdown/developer/changelog.md
@@ -1,7 +1,8 @@
 Version 5.X.X
 =============
 
-
+ * 2016-03-06 Added silent error handle to NinjaJetty that does not print out stacktraces.
+              Directory listing by NinjaJetty disabled (ra).
  * 2016-04-05 Improve Ninja usability in Google AppEngine Environment (ra).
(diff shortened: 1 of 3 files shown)


raphaelbauer:
If @danielsawan confirms the fix I'd like to add a testcase to make sure we don't break the feature...

jjlauer:
Looks like a good fix.

danielsawan:
I tried to reproduce the case locally by shutting down my sql database before and after my Ninja App has started  but i didn't manage.
When this case happen it was on my prod under heavy DDOS. It seems like it is more complicated then just shutdown the database and expect to reproduce the case.

Sorry but i can't manage to reproduce this locally :(

rrarrarra:
maybe try to simulate a DOS locally (or to your prod-system) with tools like apacheBench (ab) to reproduce the case.
`
ab -n 500 -c 100 http://localhost/
`

danielsawan:
Unfortunaly a DDOS is a way more complicated than just flood requests. It is more like flooding all UDP port with specific requests etc (i am not a ddos expert). My hosting provide me a DDOS protection so i solved my problem this way.

PR #498, opened 2016-04-06
Improve usability in Google AppEngine environment

raphaelbauer:
This PR removes the nasty logging of a bogus
exception (when logging is configured via jdk logging
as needed eg in AppEngine).

The PR also removed the usage of nio.Path in dev 
mode because it is an AppEngine restricted class 
(Rewritten with File which works just fine).


jjlauer:
Looks good.  Honestly, app engine... java.nio.Path is restricted yet its supposedly Java 7 compliant?

raphaelbauer:
There are a lot if things that are great on AppEngine. And there are a lot of things that are... well... the opposite of great. AppEngine is a diva.

PR #500, opened 2016-04-08
Fix another glitch to improve Ninja usability in Google AppEngine Environment

raphaelbauer:
yea. another usage of path which is one of the prohibited classes in AppEngine. Improves usability. Just a rewrite with other classes - changes no functionality.


jjlauer:
Again just looks good, but sad you can't use one of the more useful classes
in Java 7 IMHO.

Issue #497, opened 2016-04-05
Regex at beginning of a Route's path fails


JesseTG:
I have a `Route` of the form `(/users)?/{user}/?`.  That is, I want a user's page to be accessible at either `/users/name` like Reddit, or `/name` like GitHub.  And I don't want the presence of a trailing slash to matter.

However, this gets me a long-ass stack trace when I try to access that page.  The top of it looks like this:

```
java.lang.IndexOutOfBoundsException: index (1) must be less than size (1)
    at com.google.common.base.Preconditions.checkElementIndex(Preconditions.java:310)
    at com.google.common.base.Preconditions.checkElementIndex(Preconditions.java:292)
    at com.google.common.collect.SingletonImmutableList.get(SingletonImmutableList.java:45)
    at ninja.Route.getPathParametersEncoded(Route.java:128)
        ...
```

This problem does not occur if I change the route to `/users/{user}/?`.

bazi:
I found that using capturing groups in routes causes such issues. I proposed a fix in PR #501.
As a workaround you can use `(?:/users)?/{user}/?` until the fix is applied.

JesseTG:
That worked, thanks!  Even if you didn't have a proposed fix, it's not _that_ big of a deal since I don't need `(?:/users)` to capture.

JesseTG:
Fixed by #501

raphaelbauer:
:+1:

flniu:
Thanks! Solved my problem.

PR #501, opened 2016-04-08
fix issue #497

bazi:
Fixes issue #497 

Capturing groups in routes cause mismatch of variable parts and groups in route regex.
To avoid such cases, all groups in a route are converted to non-capturing groups.

### DIFF
--- a/ninja-core/src/main/java/ninja/Route.java
+++ b/ninja-core/src/main/java/ninja/Route.java
@@ -171,7 +171,12 @@ public class Route {
      */
     protected static String convertRawUriToRegex(String rawUri) {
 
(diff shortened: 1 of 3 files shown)


jjlauer:
What happens to users who already had added a non-capture group regex to their routes?

bazi:
Oops! It was in my mind but missed it in regex. Fixed!
Non-capture groups are left intact. Thanks @jjlauer ;)

JesseTG:
Fix #497

raphaelbauer:
afaics that's a very nice solution :) good work!

raphaelbauer:
@jjlauer any objections against merging?

jjlauer:
Looks good to me.

raphaelbauer:
:+1:

PR #504, opened 2016-04-15
Enhanced parsing of input values (post or parameters)

jlannoy:
Hello.

I already tried to send this [request](https://github.com/ninjaframework/ninja/pull/448) before but your review, the modifications needed and the evolution of the main repository I have not been able to do a rebase. I tried several times but finally it took less pain to copy them on a fresh branch. 

If you remember it, the goal of this PR is to enhance the parsing of strings to simple types, from the parameters or from the request. In details, it allows :
- To avoid duplicated code or differences between ParamParsers and SwissKnife converters (their job are in fact the same)
- To allow the POST body engine to register validation errors on conversion failure
- To avoid registering enumeration at start
- To allow users to register their own ParamParser
- Even overriding default ones, which can be useful to specify a particular DateFormat
- To allow parsing of other objects inside the parsed POST object

Notes :
- Custom ParamParsers are instancied via an Injector, so that users can make usage of @Inject dependencies
- The convert method for Date from the SwissKnife class have been translated to a ParamParser, based on LocalDateTime (as the original method). I’m thinking about something more useful, with date format from the properties file based on current Locale ; but I didn’t want to make a bigger PR. The fact is that with this PR, each user will be able to write a parser with it’s own format.
- It works with the MultiBinder, and is documented as well
- I had to enhanced the Validation documentation page, to make the differences with the JSR303 validator clear (interpolated messages in bean violation // message keys in field validations)

Regards,

Jonathan

### DIFF
--- a/ninja-core/src/main/java/ninja/Bootstrap.java
+++ b/ninja-core/src/main/java/ninja/Bootstrap.java
@@ -209,7 +209,6 @@ public class Bootstrap {
             @Override
             protected void configure() {
                 bind(Ninja.class).to(ninjaClass).in(Singleton.class);
-                Multibinder.newSetBinder(binder(), ParamParser.class);
             }
         });
     }
--- a/ninja-core/src/main/java/ninja/Bootstrap.java
+++ b/ninja-core/src/main/java/ninja/Bootstrap.java
@@ -20,9 +20,8 @@ import java.util.ArrayList;
 import java.util.List;
 
 import ninja.application.ApplicationRoutes;
-import ninja.lifecycle.LifecycleSupport;
 import ninja.logging.LogbackConfigurator;
-import ninja.scheduler.SchedulerSupport;
+import ninja.params.ParamParser;
 import ninja.utils.NinjaConstant;
 import ninja.utils.NinjaProperties;
 import ninja.utils.NinjaPropertiesImpl;
@@ -38,6 +37,8 @@ import com.google.inject.Injector;
 import com.google.inject.Module;
 import com.google.inject.Singleton;
 import com.google.inject.Stage;
+import com.google.inject.multibindings.Multibinder;
+
 import ninja.conf.FrameworkModule;
 import ninja.conf.NinjaBaseModule;
 import ninja.conf.NinjaClassicModule;
@@ -208,6 +209,7 @@ public class Bootstrap {
             @Override
             protected void configure() {
                 bind(Ninja.class).to(ninjaClass).in(Singleton.class);
+                Multibinder.newSetBinder(binder(), ParamParser.class);
             }
         });
     }
--- a/ninja-core/src/main/java/ninja/bodyparser/BodyParserEnginePost.java
+++ b/ninja-core/src/main/java/ninja/bodyparser/BodyParserEnginePost.java
@@ -65,7 +65,7 @@ public class BodyParserEnginePost implements BodyParserEngine {
             return null;
         }
 
-        for(String declaredField : getAllDeclaredFieldsAsStringSet(classOfT)) {
+        for (String declaredField : getAllDeclaredFieldsAsStringSet(classOfT)) {
 
             try {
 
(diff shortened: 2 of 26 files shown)


jlannoy:
Hello @raphaelbauer and @jjlauer ... Don't forget this change. It was quite hard to merge my modifications from the previous pull request, I'd rather not doing it again. Thanks!

raphaelbauer:
Hi @jlannoy thanks for bringing that PR back to our attention! @jjlauer I guess you had a discussion about the PR already - would you like to handle it or should I do it?

jjlauer:
@raphaelbauer I am compressed on time these days so it'd be great if you could handle it!

raphaelbauer:
Cool. Will do. @jlannoy I will handle the PR in the coming days :)

ninja-core/src/main/java/ninja/Bootstrap.java
```diff
@@ -208,6 +209,7 @@ protected void configure() throws Exception {
             @Override
             protected void configure() {
                 bind(Ninja.class).to(ninjaClass).in(Singleton.class);
+                Multibinder.newSetBinder(binder(), ParamParser.class);
```
raphaelbauer: Not sure, but either we add the binder here or in NinjaBaseModule - but not in both... I guess NinjaBaseModule is the right place...

ninja-core/src/main/java/ninja/bodyparser/BodyParserEnginePost.java
```diff
@@ -49,46 +65,67 @@
             return null;
         }
 
-        Set<String> declaredFields = getAllDeclaredFieldsAsStringSet(classOfT);
+        for(String declaredField : getAllDeclaredFieldsAsStringSet(classOfT)) {
```
raphaelbauer: tiny, but there should be a space between for and ( like "for ("...

ninja-core/src/main/java/ninja/bodyparser/BodyParserEnginePost.java
```diff
@@ -49,46 +65,67 @@
...
-                        | IllegalAccessException e) {
+                    }
 
-                    logger.warn(
-                            "Error parsing incoming Post request into class {}. Key {} and value {}.", 
-                            classOfT.getName(), ent.getKey(), ent.getValue(), e);
+                } else {
+                    
+                    // Check if we have a parameter key corresponding to inner attributes of this field
+                    for(String parameter : context.getParameters().keySet()) {
```
raphaelbauer: Again very tiny, but there should be a space between for and ( like "for ("...

ninja-core/src/main/java/ninja/params/ParamParsers.java
```diff
@@ -462,46 +492,68 @@ public Character parseParameter(String field, String parameterValue, Validation
...
-                        if (value.name().equals(parameterValue)) {
-                            return value;
-                        }
-                    } else {
-                        if (value.name().equalsIgnoreCase(parameterValue)) {
-                            return value;
-                        }
+//                try {
+//                    return Enum.valueOf(targetType, parameterValue);
+//                }
```
raphaelbauer: I think you can remove the block that is commented out...

raphaelbauer:
Hey @jlannoy that PR is very well done. Puh. What a great work on so many parts of Ninja :) Congrats!

There are some minor things I annotated. Once these are resolved we'll merge the PR (and make a new Ninja release...)

ninja-core/src/main/java/ninja/Bootstrap.java
jlannoy: I completely agree... I think I did it twice because I started this PR before the NinjaBaseModule class appeared ;-)

jlannoy:
Thank you @raphaelbauer. I fixed the 4 comments, you can check the last commit ;-)

raphaelbauer:
Perfect - Thanks!

PR #508, opened 2016-05-10
Crypto should use getBytes("UTF-8")

lishid:
This is to guarantee consistent behavior across different platforms.

Disclaimer: The reason why this is desirable for me is to get consistent behavior with Play: 
https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/libs/crypto/Crypto.scala#L238

I think a migration from Play framework should work regardless, but having this here is probably for the best.

### DIFF
--- a/ninja-core/src/main/java/ninja/utils/Crypto.java
+++ b/ninja-core/src/main/java/ninja/utils/Crypto.java
@@ -16,6 +16,7 @@
 
 package ninja.utils;
 
+import java.nio.charset.StandardCharsets;
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
 
@@ -57,7 +58,7 @@ public class Crypto {
         try {
 
             // Get an hmac_sha1 key from the raw key bytes
-            byte[] keyBytes = key.getBytes();
+            byte[] keyBytes = key.getBytes(StandardCharsets.UTF_8);
             SecretKeySpec signingKey = new SecretKeySpec(keyBytes, "HmacSHA1");
 
             // Get an hmac_sha1 Mac instance and initialize with the signing key
@@ -65,7 +66,7 @@ public class Crypto {
             mac.init(signingKey);
 
             // Compute the hmac on input data bytes
-            byte[] rawHmac = mac.doFinal(value.getBytes());
+            byte[] rawHmac = mac.doFinal(value.getBytes(StandardCharsets.UTF_8));
 
             // Convert raw bytes to Hex
             byte[] hexBytes = new Hex().encode(rawHmac);
--- a/ninja-core/src/site/markdown/developer/changelog.md
+++ b/ninja-core/src/site/markdown/developer/changelog.md
@@ -1,6 +1,7 @@
 Version 6.x.x
 =============
 
+ * 2016-09-20 Session signatures now explicitly use UTF-8 for String.getBytes (lishid)
  * 2016-09-01 Bump to minimum requirement of Java 8
  * 2016-09-01 jetty from 9.2.10.v20150310 to 9.3.11.v20160721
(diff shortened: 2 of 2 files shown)


jjlauer:
Couple issues.  First, you'll need to add one or more unit tests.  Crypto is a critical class and so we need to be careful w/ what is accepted.  If you examples of Play compatibility then those would be good to potentially include.  Second, I would have thought Ninja users would have had already hit this issue if its actually a bug/platform-specific issue.  Using Play's .scala class is helpful, but not definitive.  Plus, this may actually break existing Ninja sessions -- so while you get compat, it may break  other users.  So its important to understand this issue much deeper, especially with unit tests.

lishid:
Sounds good, I'll see what I can do.

lishid:
Apart from adding unit test, which is fairly straightforward, I'm unsure how to proceed with the rest (and whether it's necessary or not).

It shouldn't be a problem unless someone use anything not ASCII, because almost every modern encoding is an extension of ASCII, and since we generate only alpha-numeric characters for application.secret, it should be equivalent with or without using UTF-8.

So I guess it all comes down to whether anyone would use an application.secret that's outside the common character range, but for backward compatibility it's probably impractical to enforce a strict encoding.

jjlauer:
You're encoding both the key/secret as UTF-8 and also the value too.  The value is what likely has non-ascii characters and what I'm mostly concerned about.  I concur that the secret likely only has ascii.

lishid:
Hmm, that's correct. Would it make sense to have this as a configuration option (by default uses the system default charset)?

jjlauer:
Perhaps it doesn't matter:
http://stackoverflow.com/questions/25665703/encoding-scheme-used-for-cookies

I guess its possible the value is only url-encoded values which should all
fit in ascii.  You mentioned some sort of issue with Play signatures?  Do
you have an example of where Ninja's signing didn't match Play's?  You
could try out some unicode chars in Ninja's session to see if getBytes()
not passing in UTF-8 really is an issue.

-Joe

lishid:
I'll try and see if I can find some cases. I really just quickly glanced over Ninja's code and Play's code for session handling and noticed that they are pretty much equivalent except for the "UTF-8" parameter.

raphaelbauer:
I think it would make sense to pin getBytes to UTF-8. I guess close to nobody runs the same Ninja application on both Windows and a Linux box - therefore we did not yet get any bug reports, because then the default encoding is always the same.

Afaics the worst thing that can happen is that sessions will be deleted on systems that do not use UTF-8 as default encoding. Which should not be a problem - you just have to login again...

bazi:
Just to make a note: `String.getBytes()` uses default charset and in our case it is "utf-8". Check out [here](https://github.com/ninjaframework/ninja/blob/develop/ninja-core/src/main/java/ninja/conf/NinjaBaseModule.java#L44).

raphaelbauer:
Yea... true! 
I tend not to trust the file.encoding property - but it basically means we are already using utf-8 everywhere. Changing String.getBytes() to String.getBytes("utf-8") should not change a bit anywhere...

Just in case the JVM does not respect "file.encoding" we should anyway change to String.getBytes("utf-8") imho...

metacity:
Ninja requires Java 7, so `getBytes(StandardCharsets.UTF_8)`? :)

raphaelbauer:
@jjlauer ok to merge?

jjlauer:
Yeah, with the major change to Ninja 6, nows the best time to make this change.

raphaelbauer:
@lishid Could you please add one line to changelog.md? Then we'll merge it asap... Thanks!

raphaelbauer:
awesome! thanks for your contribution!

PR #512, opened 2016-06-09
Made dropping of DB schema with Flyway migrations configurable.

jfendler:
- Added boolean configuration property "ninja.migration.drop", allowing
  users to enable or disable dropping of the database with flyway.
- Modified decision to let flyway drop the DB schema based on the
  following: boolean value of "ninja.migration.drop" property, and if not
  specified assuming "true" in test mode, and "false" in any other mode.
  This keeps backward compatibility with existing installations.

### DIFF
--- a/ninja-core/src/main/java/ninja/migrations/flyway/MigrationEngineFlyway.java
+++ b/ninja-core/src/main/java/ninja/migrations/flyway/MigrationEngineFlyway.java
@@ -49,7 +49,8 @@ public class MigrationEngineFlyway implements MigrationEngine {
 
         // In testmode we are cleaning the database so that subsequent testcases
         // get a fresh database.
-        if (ninjaProperties.isTest()) {
+        if (ninjaProperties.getBooleanWithDefault(NinjaConstant.NINJA_MIGRATION_DROP_SCHEMA,
+                ninjaProperties.isTest() ? true : false )) {
             flyway.clean();
         }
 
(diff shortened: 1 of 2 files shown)


jfendler:
The motivation behind this PR is to allow me to keep an existing DB for testing (which is quite large) without being forced by ninja to have the schema dropped in test mode.
I'm currently circumventing this issue by running ninja in dev mode during testing, but I'd like to move back to using the proper mode.
By default, the old behaviour of ninja/flyway is retained.

@raphaelbauer / @jjlauer I'm not sure if we should perhaps add another "safety net" to effectively prohibit (or make it rather hard, e.g. by using an additional system property) dropping the DB in production mode ever. Alternatively, we could continue to only run flyway.clean() in test mode, but after checking the additional property.. let me know..

raphaelbauer:
That's pretty useful! Can you think of a test for that? I think that would be important... Documentation and changelog are needed, too. Once that's done we should merge asap 👍

jfendler:
@raphaelbauer I fully agree - test cases are definitely needed for this one. I'll come up with something.. Just wanted to find out if you consider this useful at all (or if there was perhaps some deeper reason for always dropping in test).

jjlauer:
@jfendler Just raising this PR back up to the top.  I know I'd use this feature since I regularly need to drop the database during dev (vs. just for testing).  Any thoughts on a unit test?

jfendler:
@jjlauer I'll see what I can do, but I don't have much time available these days (preparing to relocate from Namibia back to Germany). If you want to lend a hand that'd be highly appreciated.

jjlauer:
@jfendler This change is fairly simple.  I'm going to merge to get it in prior to the switch to Java 8.

PR #513, opened 2016-06-21
Fix directory listing

mallowlabs:
### Steps to reproduce

``` sh
$ mvn archetype:generate -DarchetypeGroupId=org.ninjaframework -DarchetypeArtifactId=ninja-servlet-archetype-simple
$ cd MY_INSTALLED_PROJECT
$ mvn clean package
$ mvn ninja:run -Dninja.mode=prod -Dninja.ssl.port=-1 &
$ curl http://localhost:8080/assets/css/
custom.css
```

I think that `curl`should get "404 not found". But Ninja returns file list in the directory.
This PR will fix this problem.

### DIFF
--- a/ninja-core/src/main/java/ninja/AssetsController.java
+++ b/ninja-core/src/main/java/ninja/AssetsController.java
@@ -127,6 +127,9 @@ public class AssetsController {
         // check if stream exists. if not print a notfound exception
         if (url == null) {
             context.finalizeHeadersWithoutFlashAndSessionCookie(Results.notFound());
+        } else if (url.getProtocol().equals("file") && new File(url.getPath()).isDirectory()) {
+            // Disable listing of directory contents
+            context.finalizeHeadersWithoutFlashAndSessionCookie(Results.notFound());
         } else {
             try {
                 URLConnection urlConnection = url.openConnection();
--- a/ninja-core/src/site/markdown/developer/changelog.md
+++ b/ninja-core/src/site/markdown/developer/changelog.md
@@ -1,3 +1,8 @@
+Version X.X.X
+=============
+
+ * 2016-07-04 Fixed directory listing in AssetsController (mallowlabs)
+
 Version 5.6.0
 =============
 
--- a/ninja-core/src/test/java/ninja/AssetsControllerTest.java
+++ b/ninja-core/src/test/java/ninja/AssetsControllerTest.java
@@ -138,6 +138,21 @@ public class AssetsControllerTest {
     }
 
     @Test
+    public void testServeStaticDirectory() throws Exception {
+        when(contextRenderable.getRequestPath()).thenReturn("/");
+        Result result2 = assetsController.serveStatic();
+
+        Renderable renderable = (Renderable) result2.getRenderable();
+
+        Result result = Results.ok();
+
+        renderable.render(contextRenderable, result);
+
+        verify(contextRenderable).finalizeHeadersWithoutFlashAndSessionCookie(resultCaptor.capture());
+        assertEquals(Results.notFound().getStatusCode(), resultCaptor.getValue().getStatusCode());
+    }
+
+    @Test
     public void testServeStatic304NotModified() throws Exception {
 
         when(contextRenderable.getRequestPath()).thenReturn(


jjlauer:
@mallowlabs I think you found an issue, not sure the fix is correct though.  In production everything would be packaged as a jar and I'm not sure your directory check would work.  Writing unit tests for that particular case is complicated as well w/ maven.  Instead of running your fix with `mvn ninja:run -Dninja.mode=prod -Dninja.ssl.port=-1` run it with `java -Dninja.ssl.port=-1 -jar target/<project>-1.0-SNAPSHOT.jar`.  That archetype will build a shaded jar and using that instead will actually mimic running in production better since the current project's dir on the filesystem won't be on the classpath vs. the way you suggested.

mallowlabs:
@jjlauer Thank you for reviewing my PR.

> In production everything would be packaged as a jar

Ninja supports [War based deployment](http://www.ninjaframework.org/documentation/deployment/war_based_deployment.html).
I am faced with this problem in production with war based deployment.
Tomcat expands the war file to webapps directory, and the assets directory exists on the filesystem.
I think that Ninja should prevent directory traversal problem on the filesystem if it supports war based deployment.

> I'm not sure your directory check would work

As you say, my directory check may not work in the classpath.
But `AssetsController` does not expose the file list if assets directory existed in the classpath.
I guess that it is the behavior of `URLConnection#getInputStream()`.
So I think that the care only about filesystem is enough to prevent directory traversal.

raphaelbauer:
Good catch! There is another (I guess connected) issue when running Ninja in standalone mode via java -Dninja.ssl.port=-1 -jar target/<project>-1.0-SNAPSHOT.jar and calling curl http://localhost:8080/assets/css/.

I guess in those cases the url is not null, but url.openConnection() returns null (I did not check it though...).

`
23:34:35.229 [qtp1720435669-12] ERROR ninja.NinjaDefault - Emitting bad request 500. Something really wrong when calling route: /assets/css (class: class ninja.AssetsController method: public ninja.Result ninja.AssetsController.serveStatic())
java.lang.NullPointerException: null
    at java.io.FilterInputStream.read(FilterInputStream.java:133) ~[na:1.8.0_92]
    at java.io.FilterInputStream.read(FilterInputStream.java:107) ~[na:1.8.0_92]
    at com.google.common.io.ByteStreams.copy(ByteStreams.java:70) ~[trav-1.0-SNAPSHOT.jar:na]
    at ninja.AssetsController.streamOutUrlEntity(AssetsController.java:155) ~[trav-1.0-SNAPSHOT.jar:na]
    at ninja.AssetsController.access$200(AssetsController.java:49) ~[trav-1.0-SNAPSHOT.jar:na]
    at ninja.AssetsController$1.render(AssetsController.java:99) ~[trav-1.0-SNAPSHOT.jar:na]
    at ninja.utils.ResultHandler.handleRenderable(ResultHandler.java:87) ~[trav-1.0-SNAPSHOT.jar:na]
    at ninja.utils.ResultHandler.handleResult(ResultHandler.java:60) ~[trav-1.0-SNAPSHOT.jar:na]
    at ninja.NinjaDefault.onRouteRequest(NinjaDefault.java:104) ~[trav-1.0-SNAPSHOT.jar:na]
    at ninja.servlet.NinjaServletDispatcher.service(NinjaServletDispatcher.java:86) [trav-1.0-SNAPSHOT.jar:na]
    at com.google.inject.servlet.ServletDefinition.doServiceImpl(ServletDefinition.java:287) [trav-1.0-SNAPSHOT.jar:na]
    at com.google.inject.servlet.ServletDefinition.doService(ServletDefinition.java:277) [trav-1.0-SNAPSHOT.jar:na]
    at com.google.inject.servlet.ServletDefinition.service(ServletDefinition.java:182) [trav-1.0-SNAPSHOT.jar:na]
    at com.google.inject.servlet.ManagedServletPipeline.service(ManagedServletPipeline.java:91) [trav-1.0-SNAPSHOT.jar:na]
    at com.google.inject.servlet.FilterChainInvocation.doFilter(FilterChainInvocation.java:85) [trav-1.0-SNAPSHOT.jar:na]
    at com.google.inject.servlet.ManagedFilterPipeline.dispatch(ManagedFilterPipeline.java:119) [trav-1.0-SNAPSHOT.jar:na]
    at com.google.inject.servlet.GuiceFilter$1.call(GuiceFilter.java:133) [trav-1.0-SNAPSHOT.jar:na]
    at com.google.inject.servlet.GuiceFilter$1.call(GuiceFilter.java:130) [trav-1.0-SNAPSHOT.jar:na]
    at com.google.inject.servlet.GuiceFilter$Context.call(GuiceFilter.java:203) [trav-1.0-SNAPSHOT.jar:na]
    at com.google.inject.servlet.GuiceFilter.doFilter(GuiceFilter.java:130) [trav-1.0-SNAPSHOT.jar:na]
    at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1652) [trav-1.0-SNAPSHOT.jar:na]
    at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:585) [trav-1.0-SNAPSHOT.jar:na]
    at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1127) [trav-1.0-SNAPSHOT.jar:na]
    at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:515) [trav-1.0-SNAPSHOT.jar:na]
    at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1061) [trav-1.0-SNAPSHOT.jar:na]
    at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:141) [trav-1.0-SNAPSHOT.jar:na]
    at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:97) [trav-1.0-SNAPSHOT.jar:na]
    at org.eclipse.jetty.server.Server.handle(Server.java:497) [trav-1.0-SNAPSHOT.jar:na]
    at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:310) [trav-1.0-SNAPSHOT.jar:na]
    at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:257) [trav-1.0-SNAPSHOT.jar:na]
    at org.eclipse.jetty.io.AbstractConnection$2.run(AbstractConnection.java:540) [trav-1.0-SNAPSHOT.jar:na]
    at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:635) [trav-1.0-SNAPSHOT.jar:na]
    at org.eclipse.jetty.util.thread.QueuedThreadPool$3.run(QueuedThreadPool.java:555) [trav-1.0-SNAPSHOT.jar:na]
    at java.lang.Thread.run(Thread.java:745) [na:1.8.0_92]
    Suppressed: java.lang.NullPointerException: null
        at java.io.FilterInputStream.close(FilterInputStream.java:181) ~[na:1.8.0_92]
        at sun.net.www.protocol.jar.JarURLConnection$JarURLInputStream.close(JarURLConnection.java:108) ~[na:1.8.0_92]
        at ninja.AssetsController.streamOutUrlEntity(AssetsController.java:156) ~[trav-1.0-SNAPSHOT.jar:na]
        ... 30 common frames omitted
`

mallowlabs:
@raphaelbauer Thank you for your comment.

But I cannot reproduce `java.lang.NullPointerException` with `jar` command and `curl` access.

My environment is:
- Oracle JDK 7 / 8 (I checked both)
- Windows 10 / OS X El Capitan (I checked both)
- Ninja 5.5.0 / 5.6.1-SNAPSHOT (my patched version)

What is your environment?

raphaelbauer:
I can reproduce it. The Nullpointer is logged by Ninja, so you don't see it in the frontend...

raphaelbauer:
Another question: What tomcat do you use to deploy which causes that issue for you?

raphaelbauer:
Just had one more look into Ninja's codebase... One thing makes me suspicious - that's why I asked about the tomcat to reproduce the issue. There is the following piece of code (getStaticFileFromAssetsDir):

```
        String finalNameWithoutLeadingSlash = assetsControllerHelper.normalizePathWithoutLeadingSlash(fileName, true);
        url = this.getClass().getClassLoader()
                .getResource(ASSETS_DIR
                             + "/"
                             + finalNameWithoutLeadingSlash);
```

which is ALWAYS (afaics) used to load assets in production. So it is simply not possible to encounter file: urls in production.

The question is why that happens to you.

Are you sure you started the application in production mode?

mallowlabs:
> I can reproduce it. The Nullpointer is logged by Ninja, so you don't see it in the frontend...

I could reproduce it!!!
`curl http://localhost:8080/assets/css/` returns 200.
But `curl http://localhost:8080/assets/css` returns 500 with NullPointerException.
Should I add null-check to my PR?

> Another question: What tomcat do you use to deploy which causes that issue for you?

Tomcat 7.0.70 / 8.0.36 (I checked both)

Both `curl http://localhost:8080/bar-1.0-SNAPSHOT/assets/css/` and `curl http://localhost:8080/bar-1.0-SNAPSHOT/assets/css` exposes the file list.

> So it is simply not possible to encounter file: urls in production.

No.
In my debugger, `getStaticFileFromAssetsDir` returns `file:/private/tmp/apache-tomcat-8.0.36/webapps/bar-1.0-SNAPSHOT/WEB-INF/classes/assets/css/`.
Of course, I just deployed `bar-1.0-SNAPSHOT.war` to the webapps directory.
I guess that the class loader of Tomcat uses the file system.

I attach a screen shot of my debugger:

debugger

You can see the `url` has `file:` protocol, and the `mode` is `prod`.

> Are you sure you started the application in production mode?

Yes.
[Ninja document](http://www.ninjaframework.org/documentation/configuration_and_modes.html) says 

> That means if you deploy Ninja via a war file to an arbitrary servlet container Ninja will use the prod mode.

When I added `ninja.mode=prod` to `application.conf`, I got the same result too.

raphaelbauer:
Ninja logs its mode - can you check what mode Ninja logs when it starts up inside your Tomcat and paste that line?

mallowlabs:
```
$ grep "Ninja is running" logs/catalina.out
22:19:57.198 [localhost-startStop-1] INFO  ninja.utils.NinjaModeHelper - Ninja is running in mode prod
```

raphaelbauer:
Okay. Thanks for the debug information. Tomcat really delivers file: resources from its classloader. I actually thought that would not happen - I was wrong.

That means we have to fix that thing a lot more wholistic way. Let me check some cases and ideas first...

raphaelbauer:
Okay. After some investigation I would propose the following (based more or less on your PR anyway):

1) Make sure we do not render directories when we have files (that is relevant for the tomcat issue, but won't fix the 500 issue described above)

```
    private void streamOutUrlEntity(URL url, Context context, Result result) {
        // check if stream exists. if not print a notfound exception

        if (url == null) {
            context.finalizeHeadersWithoutFlashAndSessionCookie(Results.notFound());
        } else if (url.getProtocol().equals("file") && new File(url.getPath()).isDirectory()) { 
            context.finalizeHeadersWithoutFlashAndSessionCookie(Results.notFound());
        } else {
```

2) Another important point: We have to make sure we don't have any problems with hackers that use directory traversal and such. The good news is that this is already implemented and in place via normalizePathWithoutLeadingSlash on both file dev resources AND resources loaded by the classloader (eg tomcat file: and classpath: etc pp stuff). So this can't happen. But I'd would be important to get your feedback on that...

What do you think?

mallowlabs:
> 1)

Your code is simpler than my PR!
I have updated my PR.

> 2)

I'm sorry... I notice that I was confused with the meaning of "directory traversal".
I think that anyone can NOT traverse directories via `normalizePathWithoutLeadingSlash`, as you say.
My problem is not "directory traversal" but "directory listing".

ninja-core/src/main/java/ninja/AssetsController.java
```diff
@@ -127,6 +127,8 @@ private void streamOutUrlEntity(URL url, Context context, Result result) {
         // check if stream exists. if not print a notfound exception
         if (url == null) {
             context.finalizeHeadersWithoutFlashAndSessionCookie(Results.notFound());
+        } else if (url.getProtocol().equals("file") && new File(url.getPath()).isDirectory()) {
```
raphaelbauer, line 130: can you add a simple comment like
// Disable listing of directory contents

raphaelbauer:
Could you squash the commits into one. And create good commit message? Imagine someone looks up that change - the commit message should tell exactly why we did that...

Otherwise it looks perfect. I'll merge asap once these tiny things are fixed. Thanks for your contribution!

mallowlabs:
I have pushed the new commit.
Please check it!

raphaelbauer:
Looks perfect! @jjlauer fine with you, or did we miss something?

jjlauer:
Looks good.

mallowlabs:
Thank you!

mallowlabs:
I tried Ninja v5.7.0, I noticed this PR does not work in some cases...

When the path of Ninja or Tomcat contains white spaces, `new File(url.getPath()).isDirectory()` always returns false.
So `AssetsController` exposes the file list of assets directory.

You can reproduce this problem with below command:

```
$ mkdir /tmp/ninja\ files/ && cd /tmp/ninja\ files/
$ git clone --depth 1 git@github.com:ninjaframework/ninja.git
$ cd ninja
$ mvn test
...
Tests in error:
  testServeStaticDirectory(ninja.AssetsControllerTest)

Tests run: 518, Failures: 0, Errors: 1, Skipped: 0
...
```

I have thought that `new File(url.getPath()).isDirectory()` equals `new File(url.toURL()).isDirectory()`.
But if the path contains special characters (ex: whitespace, #, etc.), `new File(url.getPath()).isDirectory()` does not work correctly.
So we should use `new File(url.toURL()).isDirectory()`.

I'm sorry, I overlooked this problem.
Can I create another PR to fix this problem?

raphaelbauer:
Sure. That's also why @jjlauer said it's a hard problem. I am also not sure why Tomcat maps the classloader.getResouce to a Filesyetem-based implementation. That does not sound clever...

One thing you should check: Is this a bug or a feature of Tomcat? I'd almost bet that there is a setting in Tomcat, where you can enable / disable that listing functionality... (I may be wrong though...)

jfendler:
@raphaelbauer is correct. Directory listings in Tomcat are provided by the DefaultServlet, which has some options to configure (e.g. disable listings): [see e.g. here (not sure about other versions)](https://tomcat.apache.org/tomcat-7.0-doc/default-servlet.html)

mallowlabs:
@raphaelbauer 
Thanks!

> One thing you should check: Is this a bug or a feature of Tomcat?

I think that it is not a bug of Tomcat but a **Java feature**.

Please check below short code:

``` java
File file = new File("/tmp/ninja files");
System.out.println(file.isDirectory()); // -> true
URL url = file.toURI().toURL();
System.out.println(url); // -> file:/C:/tmp/ninja%20files/
System.out.println(new File(url.toURI()).isDirectory()); // -> true
System.out.println(new File(url.getPath()).isDirectory()); // -> false (!)

List<String> lines = IOUtils.readLines(url.openStream());
for (String line : lines) {
    System.out.println(line); // -> ninja (!)
}
```

It explains two important Java features.
1. `new File(url.getPath()).isDirectory()` cannot handle paths contain white spaces.
2.  `url.openStream()` exposes the file list of directories.

It is unrelated to Tomcat.

@jfendler 
I checked my web.xml, it is default of Ninja.
It means `listings` property is `false`.
If it is a Tomcat problem,  failure of `mvn test` is strange.

PR #516, opened 2016-07-16
Fix directory listing (#513) is not enough

mallowlabs:
The PR #513 is not enough because a problem is remaining.

If the server running path contains whitespaces (or other special characters like #), the directory check does not work.
### Steps to reproduce

```
$ mkdir /tmp/ninja\ files/ && cd /tmp/ninja\ files/
$ git clone --depth 1 git@github.com:ninjaframework/ninja.git
$ cd ninja
$ mvn test
...
Tests in error:
  testServeStaticDirectory(ninja.AssetsControllerTest)

Tests run: 518, Failures: 0, Errors: 1, Skipped: 0
...
```

This PR fixes these problems.

### DIFF
--- a/ninja-core/src/main/java/ninja/AssetsController.java
+++ b/ninja-core/src/main/java/ninja/AssetsController.java
@@ -16,9 +16,7 @@
 
 package ninja;
 
-import com.google.common.annotations.VisibleForTesting;
 import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -31,14 +29,12 @@ import ninja.utils.MimeTypes;
 import ninja.utils.NinjaProperties;
 import ninja.utils.ResponseStreams;
 
-import org.apache.commons.io.FilenameUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.io.ByteStreams;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
-import org.apache.commons.lang.StringUtils;
 
 /**
  * This controller serves public resources under /assets.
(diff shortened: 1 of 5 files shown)


ninja-core/src/main/java/ninja/AssetsController.java
```diff
@@ -127,7 +123,7 @@ private void streamOutUrlEntity(URL url, Context context, Result result) {
         // check if stream exists. if not print a notfound exception
         if (url == null) {
             context.finalizeHeadersWithoutFlashAndSessionCookie(Results.notFound());
-        } else if (url.getProtocol().equals("file") && new File(url.getPath()).isDirectory()) {
+        } else if (assetsControllerHelper.isDirectoryURL(url)) {
```
raphaelbauer, line 126: It would be great if you could add a mockito.verify to AssetsController Test that makes sure assetsControllerHelper.isDirectoryURL has been called properly...

raphaelbauer:
Looks good! I added one comment to make sure we are really calling the correct method of the injected AssetsControllerHelper... Do you think you can add that test?

mallowlabs:
@raphaelbauer 
Thank you for your review. I updated my PR!
(CI is failing, but I think it is not caused by my PR.)

ninja-core/src/test/java/ninja/AssetsControllerTest.java
```diff
@@ -148,6 +156,7 @@ public void testServeStaticDirectory() throws Exception {
 
         renderable.render(contextRenderable, result);
 
+        verify(assetsControllerHelper).isDirectoryURL(this.getClass().getResource("/assets/"));
```
raphaelbauer, line 159: yea!

raphaelbauer:
Thanks! Merged 👍

mallowlabs:
Thank you!!

PR #517, opened 2016-07-19
Fixing inner objects instantiation

jlannoy:
After the PR refactoring and enhancing the parsing of request body, I discovered one not-logical behavior. An inner objet of the main object being parsed will be instantiated as soon as a parameter with the right name is present. But as the request body comes from HTLM inputs (for example), the parameter is always presents in the request even if the field is empty. So inner objects are always instantiated even when not needed, which is not relevant.

This PR fixes it ; parameters are checked before inner objects are instantiated. Note that there is not such a problem if the parameter is invalid : the object will be instantiated, the value in it will stay empty, but an error message will be registered in the context validation object.

### DIFF
--- a/ninja-core/src/main/java/ninja/bodyparser/BodyParserEnginePost.java
+++ b/ninja-core/src/main/java/ninja/bodyparser/BodyParserEnginePost.java
@@ -20,6 +20,7 @@ import java.lang.reflect.Field;
 import java.lang.reflect.ParameterizedType;
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import ninja.ContentTypes;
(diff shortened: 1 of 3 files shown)


jjlauer:
Having created a Context for Ninja using Undertow, methods such as Context.getParameters() or Context.getParameter(String) are actually not that efficient when you look how they are implemented.

https://github.com/fizzed/ninja-undertow/blob/master/ninja-undertow/src/main/java/ninja/undertow/NinjaUndertowContext.java#L232

I'd suggest you save the parameters to a local var as you do the loop to check them.  That will save quite a bit of work under-the-hood (avoiding unnecessary new maps) as a form is bound and processed.

jjlauer:
Looks good to me. @raphaelbauer what do you think?

jlannoy:
I thought the context.getParameters() was a map... It will be more efficient now ;)

raphaelbauer:
Good fix :) It would be great if you could squash the commits into one so we got a cleaner history... 
Feel free to merge anytime...

PR #518, opened 2016-07-24
Experience with "Clever Cloud" added

jlannoy:
I was looking to an other solution than Heroku, and recently tested CleverCloud. It's quite simple out there, especially when using GitHub. I don't know if it can be interesting in the official documentation but as you request that we share it...


raphaelbauer:
Nice! Thanks!

PR #520, opened 2016-07-24
Camunda Module and Structure in Modules List

jfendler:
- Added reference to new "ninja-camunda" module (Camunda BPMN integration) to list of available modules.
- Introduced different sections for modules (please re-organise if you feel like it)
- Added self to list of team members.


raphaelbauer:
Thanks! Very useful!

PR #524, opened 2016-08-26
Bump to latest dependency versions; prep for v5.8.0 release

jjlauer:
@raphaelbauer Carefully bumped versions of dependencies after reviewing their changelogs.  Most notable is Jetty to latest 9.2.x (last one targeting Java 7).  I think we should release as 5.8.0 since so many dependency versions were changed.  I'd suggest next dev to be 6.0.0-SNAPSHOT.

### DIFF
--- a/ninja-core/src/site/markdown/developer/changelog.md
+++ b/ninja-core/src/site/markdown/developer/changelog.md
@@ -1,6 +1,17 @@
-Version X.X.X
+Version 5.8.0
 =============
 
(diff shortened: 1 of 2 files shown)


raphaelbauer:
Well done! Will create a new release...

raphaelbauer:
new version is released. Pushing of new website does currently not work (not sure if this is a temporary github.com issue, or something else...). Will try that tomorrow again... And also send out release notes the coming days...

PR #529, opened 2016-10-04
Remove ninja-async-machine-beta module

jjlauer:
Module was never used and complicated some testing w/ ninja-undertow.


raphaelbauer:
Nice cleanup... :) Less code less problems...

Issue #532, opened 2016-12-15
javax.crypto.BadPaddingException when switching between instances


jjlauer:
When switching between instances (e.g. same app but configured w/ different cookie encryption keys), we are sometimes getting 500 errors rather than the expected cleaning of the Ninja session.

Example session cookie: 28afa10d7322bc2afe741755b194f92d2cbcd935-7tu7louhOSy9K78MxpbcCjzZzqkgLCt-c-zbzYZ9oo3HIKpvl0Q-5CarMlGX2NenFHQvixZyDsrLwF-ZZIVzqhVyD8aGl0DsXMN_iIwStyM

Error

```
java.lang.RuntimeException: javax.crypto.BadPaddingException: Given final block not properly padded
	at ninja.utils.CookieEncryption.decrypt(CookieEncryption.java:139) ~[org.ninjaframework.ninja-core-5.8.0.jar:na]
	at ninja.session.SessionImpl.init(SessionImpl.java:117) ~[org.ninjaframework.ninja-core-5.8.0.jar:na]
	at ninja.utils.AbstractContext.init(AbstractContext.java:95) ~[org.ninjaframework.ninja-core-5.8.0.jar:na]
	at ninja.undertow.NinjaUndertowContext.init(NinjaUndertowContext.java:110) ~[com.fizzed.ninja-undertow-5.7.0.undertow2.jar:5.7.0.undertow2]
	at ninja.undertow.NinjaUndertowHandler.handleRequest(NinjaUndertowHandler.java:53) ~[com.fizzed.ninja-undertow-5.7.0.undertow2.jar:5.7.0.undertow2]
	at io.undertow.server.handlers.form.EagerFormParsingHandler.handleRequest(EagerFormParsingHandler.java:55) ~[io.undertow.undertow-core-1.4.0.Final.jar:1.4.0.Final]
	at io.undertow.server.Connectors.executeRootHandler(Connectors.java:202) ~[io.undertow.undertow-core-1.4.0.Final.jar:1.4.0.Final]
	at io.undertow.server.HttpServerExchange$1.run(HttpServerExchange.java:805) ~[io.undertow.undertow-core-1.4.0.Final.jar:1.4.0.Final]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) ~[na:1.8.0_102]
[...]

PR #534, opened 2016-12-23
fix to issue #532

bazi:
As I am more or less familiar with session encryption classes, here I propose a *minimalistic* PR that fixes issue #532.
Returning empty string upon `GeneralSecurityException` (super class of `BadPaddingException` which causes the issue) would result in a clean session (session w/o any data). You can see it [here](https://github.com/ninjaframework/ninja/blob/develop/ninja-core/src/main/java/ninja/session/SessionImpl.java#L117-L118).
We already have logging in [CookieEncryption](https://github.com/ninjaframework/ninja/blob/develop/ninja-core/src/main/java/ninja/utils/CookieEncryption.java#L138) class.

### DIFF
--- a/ninja-core/src/main/java/ninja/utils/CookieEncryption.java
+++ b/ninja-core/src/main/java/ninja/utils/CookieEncryption.java
@@ -102,7 +102,7 @@ public class CookieEncryption {
             throw new RuntimeException(ex);
         } catch (GeneralSecurityException ex) {
             logger.error("Failed to encrypt data.", ex);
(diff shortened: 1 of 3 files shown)


jjlauer:
I think this looks good -- is there a way to have at least 1 unit test to verify it?

bazi:
added tests and updated changelog

jjlauer:
Nice work!  Hit this issue again yesterday so I'm looking forward to getting it shipped in the next beta.

Issue #538, opened 2016-12-23
Kill Guava Optionals and convert to Java 8 Optional


raphaelbauer:
Complexity: Small

Replace all Guava Optionals with Java 8 Optional. Should be straightforward.

bazi:
Done!

raphaelbauer:
Merged. Thanks for your contribution!

PR #550, opened 2016-12-25
Convert Guava `Optional` to standard Java 8 `Optional`

bazi:
PR for issue #538.
Please note that Guava dependency is removed from `ninja-core` module but it still lives in parent module (in case we need it future releases). Yeah, only `Optional` was being used from Guava.

### DIFF
--- a/ninja-core/pom.xml
+++ b/ninja-core/pom.xml
@@ -123,12 +123,6 @@
             <artifactId>freemarker-gae</artifactId>
         </dependency>
 
-        <!-- guava is just cool -->
-        <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-        </dependency>
-
         <!-- Needed for Hex transforms (security, cookies) and more... -->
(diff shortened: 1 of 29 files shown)


raphaelbauer:
Very nice! I am currently out of office and will merge first week of January... 👍

ninja-core/pom.xml
```diff
@@ -123,12 +123,6 @@
             <artifactId>freemarker-gae</artifactId>
         </dependency>
 
-        <!-- guava is just cool -->
-        <dependency>
-            <groupId>com.google.guava</groupId>
```
raphaelbauer, line 128: interesting that we don't need guava any more...

raphaelbauer: This may break stuff for some users. On the other hand. Less libraries less problems. Removing guava is ok. We can re-add it if there are grave issues...

raphaelbauer:
Added changelog.md. Thanks for your contribution!

PR #553, opened 2017-01-03
File Upload Clarification

drowe:
Added clarification about using the new way, in the event people (like myself) miss the bit about the default being 'NoFileItemProvider'.


raphaelbauer:
Nice! Thanks for your contribution!

PR #554, opened 2017-01-05
541 First sketch of Optional support in controller methods.

raphaelbauer:
That's a first sketch. It allows to use Optional in controller methods. Feedback would be great.

### DIFF
--- a/ninja-core/src/main/java/ninja/params/ControllerMethodInvoker.java
+++ b/ninja-core/src/main/java/ninja/params/ControllerMethodInvoker.java
@@ -348,7 +348,7 @@ public class ControllerMethodInvoker {
      * 
      * It investigates the type parameter and allows to remember whether a type
      * was wrapped in an Optional or not. It stores the "real" type of the parameter
-     * that the extractor should extract (String and not optional in example above).
+     * that the extractor should extract (String and not Optional in example above).
      */
     private static class MethodParameter {
         public boolean isOptional;
@@ -394,7 +394,7 @@ public class ControllerMethodInvoker {
             } else {
                 throw new RuntimeException(
                         "Oops. That's a strange internal Ninja error.\n"
-                        + "Seems somone tried to convert a type into a class that is not a real class. ( " + type.getTypeName() + ")");
+                        + "Seems someone tried to convert a type into a class that is not a real class. ( " + type.getTypeName() + ")");
             }
         }
     
--- a/ninja-core/src/main/java/ninja/params/ControllerMethodInvoker.java
+++ b/ninja-core/src/main/java/ninja/params/ControllerMethodInvoker.java
@@ -34,6 +34,11 @@ import com.google.inject.AbstractModule;
 import com.google.inject.Injector;
 import java.lang.reflect.Type;
 import java.util.Optional;
+import ninja.exceptions.BadRequestException;
+import ninja.utils.NinjaConstant;
+import ninja.utils.NinjaProperties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Invokes methods on the controller, extracting arguments out
@@ -41,12 +46,19 @@ import java.util.Optional;
  */
 public class ControllerMethodInvoker {
     
+    private final static Logger logger = LoggerFactory.getLogger(ControllerMethodInvoker.class);
+    
     private final Method method;
     private final ArgumentExtractor<?>[] argumentExtractors;
+    private final boolean useStrictArgumentExtractors;
 
-    ControllerMethodInvoker(Method method, ArgumentExtractor<?>[] argumentExtractors) {
+    ControllerMethodInvoker(
+            Method method, 
+            ArgumentExtractor<?>[] argumentExtractors,
+            boolean useStrictArgumentExtractors) {
         this.method = method;
         this.argumentExtractors = argumentExtractors;
+        this.useStrictArgumentExtractors = useStrictArgumentExtractors;
     }
 
     public Object invoke(Object controller, Context context) {
@@ -55,6 +67,9 @@ public class ControllerMethodInvoker {
         for (int i = 0; i < argumentExtractors.length; i++) {
             arguments[i] = argumentExtractors[i].extract(context);
         }
+        
+        checkNullArgumentsAndThrowBadRequestExceptionIfConfigured(arguments);
+        
         try {
             return method.invoke(controller, arguments);
         } catch (IllegalAccessException | IllegalArgumentException e) {
@@ -64,6 +79,17 @@ public class ControllerMethodInvoker {
                 throw (RuntimeException) e.getCause();
             } else {
                 throw new RuntimeException(e.getCause());
+            }
+        }
+    }
+    
+    private void checkNullArgumentsAndThrowBadRequestExceptionIfConfigured(Object[] arguments) {
+        if (!useStrictArgumentExtractors) {
+            return;
+        }
+        for (Object object : arguments) {
+            if (object == null) {
+                throw new BadRequestException();
             }
         }
     }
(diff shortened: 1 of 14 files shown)


jjlauer:
This looks awesome! This looks pretty well implemented but you called it a sketch?  What more is left? I had a question around type conversion vs. empty parameters.  If someone supplies a value, but it wasn't able to convert to a type, should we change how that's handled (e.g. throwing a BadRequestException).

raphaelbauer:
That's an open question. If an extractor extracts null - should we then automatically redirect to a BadRequest? We could simply do that in 

    public Object invoke(Object controller, Context context) {
        // Extract arguments
        Object[] arguments = new Object[argumentExtractors.length];
        for (int i = 0; i < argumentExtractors.length; i++) {
            arguments[i] = argumentExtractors[i].extract(context);
        }

and throw a BadRequestException when any argument is null. Very strict. And breaking. But nice because it really kills null in Controller methods.

Any suggestions welcome!

ninja-core/src/main/java/ninja/params/OptionalArgumentExtractor.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public Optional<T> extract(Context context) {
+        return Optional.ofNullable(wrapped.extract(context));
+    }
+
+    @Override
+    public Class<Optional<T>> getExtractedType() {
+         // FIXME => how can I create the Class for it during runtime?
```
raphaelbauer: Any suggestions welcome...

jjlauer:
I would rather not break apps, so I was thinking a ninja properties "strict" mode would be useful -- something like `ninja.strict_argument_extractors = true`.  Maybe argument extractors can be set to strict and they are then aggressive at throwing exceptions.  If you wrap something in an `Optional<>` that's the way to swallow the exception even if you're in strict mode.

raphaelbauer:
Sounds good. 

We could set it by default to false and print a warning sign when Ninja starts up "Using non-strict method injection ... blabla... Please upgrade your application to ninja.strict_argument_extractors = true asap as it will be the default from Ninja 7.0.0 onwards...".

Not really necessary, but I don't really like maintaining "two" ways of handling things... 

What do you think?

raphaelbauer:
Pinging @jlannoy as well. Afaicr you also did a lot of work in the past in the classes I touched. Any feedback really welcome :) Many thanks in advance!

jjlauer:
Plan on strict sounds perfect.

ninja-core/src/main/java/ninja/params/ControllerMethodInvoker.java
```diff
@@ -67,6 +83,17 @@ public Object invoke(Object controller, Context context) {
         }
     }
     
+    private void checkNullArgumentsAndThrowBadRequestExceptionIfConfigured(Object[] arguments) {
+        if (!useStrictArgumentExtractors) {
+            return;
+        }
+        for (Object object : arguments) {
+            if (object == null) {
+                throw new BadRequestException();
```
raphaelbauer, line 94: Here we redirect to BadRequest page...

raphaelbauer:
ok... that's the version with strict mode that throws BadRequestException. It can be configured via ninja.strict_argument_extractors.

There are tests missing for eg the BadRequestException. I'll add that later - I first wanted to make sure that this is what we need...

Any feedback welcome. Thanks!

raphaelbauer:
We also got now tests to make sure the strict mode works properly... Feedback appreciated... Thanks!

jjlauer:
@raphaelbauer looks great! Strange error with Travis CI though on the commit.

raphaelbauer:
closed / reopened to make Travis CI (hopefully) pick up the correct commit...

jlannoy:
Hi.

Sorry I didn't had time last 3 days. It seems ok for me. But two side notes : 
- The ninja @Required for parameters validation become deprecated
- Tests should be added to check if other parameters validation annotations works with optional

I can handle these two inside the validation issue. I already started bu I'll need to merge with your current issue. BTW... Are anyone of you using these annotations? It don't think it is officially documented anywhere. ;)

raphaelbauer:
Hey @jlannoy , these are good points! 
Required should be deprecated...

And you are right - there are some tests missing. Did you already do something regarding the validation tests? Otherwise I'll simply add the tests (not sure if I'll rewrite the testcase, it's a bit repetitive right now.)...

jlannoy:
Yes, I've done the main part of the issue we discussed (refactor Validation class + all impacts/tests). I was about to start the small other changes proposed in this issue, so I can manage these points on validation annotations myself.

raphaelbauer:
I added a lot of more tests making sure that Validation is up to the task
Required is now also deprecated.

@jjlauer @jlannoy  Would be great if we could merge that asap if that's okay with you... I'd like to try it asap in production :)

jlannoy:
Seems okay for me. If you merge it before friday, I can make a proposal on saturday for the Validation enhancement issue... And I'll stay available next days for your reviews. So maybe you can include both on the next beta version?

jjlauer:
@raphaelbauer I'm already using beta1 in production and I'm pretty interested in using beta2 as well.  We have a couple good PRs that I think are worth pushing out in a beta2.

raphaelbauer:
Should we release beta2 now? or are there any things we want to merge...?

jjlauer:
I can't think of anything else I know of for a beta2.  I'm a thumbs up on
it!

raphaelbauer:
ninja-6.0.0-beta2 has been released... I'll write an announcement to the mailing list soon...

jjlauer:
hey @raphaelbauer was using beta2 and had a couple questions.  I turned strict mode on and my app gets a bad request exception (as expected) if the parameter is not provided.  Booleans don't work like i thought though.  Seems like if the value doesn't match "true" then its just false.  I was wondering if we should be more strict and actually require it to be true or false otherwise null?  Integers and Longs do work as expected, btw.  If you try to set it to a string, you trigger a bad request exception.

Only other feedback is that it'd be nice to know on the console/log file what parameter failed to convert and triggered the bad request exception.  Right nows its a tad hard to know what param triggered it.

Optionals are awesome to use and make a lot of sense.  Thank you for cranking this out!

raphaelbauer:
just to understand that...

given:
- strict mode is true
- controller method looks like public Result booleanParamWithOptional(@Param("param1") Optional<Boolean> param1);
- in the request param1 is not available (aka null)

then
- you expect a badRequest Exception - right?

raphaelbauer:
Ah sorry... I think I misread that...
You mean that we always get "false" as boolean value as long as it is not null... right?

This comes from:
Boolean's
```
    public static boolean parseBoolean(String s) {
        return ((s != null) && s.equalsIgnoreCase("true"));
    }
```

which we use here:

```
public static class BooleanParamParser implements ParamParser<Boolean> {
        @Override
        public Boolean parseParameter(String field, String parameterValue, Validation validation) {
            if (parameterValue == null || parameterValue.isEmpty() || validation.hasFieldViolation(field)) {
                return null;
            } else {
                return Boolean.parseBoolean(parameterValue);
            }
        }

        @Override
        public Class<Boolean> getParsedType() {
            return Boolean.class;
        }
    }
```

That looks like a bug. And it is not really related to Optional, but to the way we handle Booleans in controllers... I think that is wrong...

I'd propose to fix that in the BooleanParamParser (and also PrimitiveBooleanParamParser)...  (but it may be a breaking change for some).

jjlauer:
Yes, looks like you found it.  I think just Boolean behaves that way.
You're suggested fix sounds right.  I'd expect a case insenstive "true" and
"false" to work, but something like "blah" to be null (rather than false).

Do you think it'd be possible to include a little bit of info about which
parameter caused the BadRequestException as well?  It would help narrow
down what happened IMHO.

raphaelbauer:
Moving discussion to https://github.com/ninjaframework/ninja/issues/557

PR #558, opened 2017-01-19
557 Improve parsing of Boolean

raphaelbauer:


jjlauer:
Looks great :1st_place_medal:

PR #559, opened 2017-01-20
547 First sketch of global filters

raphaelbauer:
Feedback appreciated!

### DIFF
--- a/ninja-core/src/main/java/ninja/Bootstrap.java
+++ b/ninja-core/src/main/java/ninja/Bootstrap.java
@@ -40,6 +40,8 @@ import com.google.inject.Stage;
 import ninja.conf.FrameworkModule;
 import ninja.conf.NinjaBaseModule;
 import ninja.conf.NinjaClassicModule;
+import ninja.utils.NinjaBaseDirectoryResolver;
+import ninja.utils.SwissKnife;
 
 /**
  * Bootstrap for a Ninja application.  Assists with initializing logging,
@@ -59,6 +61,7 @@ public class Bootstrap {
     static public final String NINJA_CONVENTION_LOCATION = "conf.Ninja";
 
     private final NinjaPropertiesImpl ninjaProperties;
+    protected final NinjaBaseDirectoryResolver ninjaBaseDirectoryResolver;
     private final List<Module> modulesToLoad;
     private final Optional<String> applicationModulesBasePackage;
     private Injector injector = null;
@@ -73,6 +76,7 @@ public class Bootstrap {
                 = Optional.ofNullable(ninjaProperties.get(
                         NinjaConstant.APPLICATION_MODULES_BASE_PACKAGE));
         
+        this.ninjaBaseDirectoryResolver = new NinjaBaseDirectoryResolver(ninjaProperties); 
     }
 
     public synchronized void boot() {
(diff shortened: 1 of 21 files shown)


jjlauer:
hey @raphaelbauer one issue jumps out -- how would I disable/exclude a global filter for a particular route? If we wanted to switch session handling to be implemented with a global filter, we'd want to be able to exclude it by default with assets.  I was thinking sessions would be a good use-case to verify the design, but maybe I'm overthinking it.  What do you think?

raphaelbauer:
Thanks for the feedback... I have to think about that I guess... You can of course do that inside a global filter. Inside the filter you can do something like "if !context.path.startsWith("/assets")". assets should be typesafe via reverse routing of course... But that's just an idea..

I also have to think about how to do the binding with the GlobalFilters class... So... All in flux... Let's continue the discussion...

How did you imagine the routing? Doing it in the router and specifying filters on the routes? How would your Dream-Api look like?

raphaelbauer:
Hey @jjlauer any strong opinions on the feature? I'd need that feature in one of my projects, so it's important to me. Otherwise I'd continue with the PR. Thanks!

jjlauer:
Hey @raphaelbauer - what do you think of something like this?  We could allow your current impl to setup the GlobalFilters, but then you could override it per route with something else.  `globalFilters` could accept a list of filter classes, or a List of classes, etc.  You could then freely setup whatever filter pipelines you want and explicitly set them on a route.

I suspect for my assets, I'd want to supply a null/empty list so session handling is skipped.

```java
router.GET().route("/").globalFilters(...).with(ApplicationController::index);
```

Not sure this is very elegant though and it could be confusing.

momiji:
Hi, we've made a #385 for something like this 18 months ago, and are using something quite similar today in production.

Our use case:
- define global filters for all routes, meaning that if not overriden somewhere, it just applies by default.
- replace global filters with specific filters at route definition or with the class annotation.  When the list starts with GlobalFilter.class, it just adds the other items to the default global filter list.

We made all this by overriding the Ninja.class, so we stick to the framework without modifying it. It should be easier to be able to define the filters in Route.java, but well the PR didn't pass :)

Having them in Route.java is definitely an improvement, as it allows to define all filters in a single place, without having to look at all controllers to review them. Another point is it allows to create controllers in common projects, when used by mutliple applications, and define the filters at project level.

Hope it helps you find the best way for doing it...

Regards.

raphaelbauer:
Interesting. I never had that usecase...

Regarding globalFilters

    router.GET().route("/").globalFilters(...).with(ApplicationController::index);

Would that be something like this?

    globalFilters(MyFilter.class, MyOtherFilter.class);

raphaelbauer:
Added support for globalFilters(...) and filters(...).
No tests and documentation yet.

What do you think?

( List<Filter> getGlobalFilters() will likely change to  List<Class<? extends Filter> getGlobalFilters();)

jjlauer:
That looks pretty awesome and would solve what I was thinking.  My only other idea is that we'd potentially move session handling to a default GlobalFilter instance.  We could use the Guice optional stuff.  Users then would need to add it back in for their own implementation of global filters.

Any harm in accepting an Iterable<> or List<> as well as the overloaded "..." method of classes?

raphaelbauer:
Good idea! Testcases and meaningful documentation is coming the next days...

raphaelbauer:
Implementation + Tests are done. Documentation is coming...

Feedback appreciated! Thanks!

raphaelbauer:
Documentation added...

jjlauer:
@raphaelbauer This looks awesome.  I'm good w/ merging.

raphaelbauer:
👍

Issue #565, opened 2017-02-15
Routes w/ no path params trigger NPE if ninja.Route.getPathParametersEncoded called


jjlauer:
Looks like the routing changes for latest Ninja break something that used to work.

A route like this used to work:

```java
router.GET().route("/robots.txt").with(AssetsController::serveStatic);
```

With Ninja 6 an NPE is triggered:

```
java.lang.NullPointerException: null
	at ninja.Route.getPathParametersEncoded(Route.java:128) ~[ninja-core-6.0.0-beta2.jar:na]
	at ninja.utils.AbstractContext.getPathParameter(AbstractContext.java:189) ~[ninja-core-6.0.0-beta2.jar:na]
	at ninja.AssetsController.getFileNameFromPathOrReturnRequestPath(AssetsController.java:225) ~[ninja-core-6.0.0-beta2.jar:na]
	at ninja.AssetsController.access$000(AssetsController.java:45) ~[ninja-core-6.0.0-beta2.jar:na]
	at ninja.AssetsController$1.render(AssetsController.java:93) ~[ninja-core-6.0.0-beta2.jar:na]
	at ninja.utils.ResultHandler.handleRenderable(ResultHandler.java:87) ~[ninja-core-6.0.0-beta2.jar:na]
	at ninja.utils.ResultHandler.handleResult(ResultHandler.java:60) ~[ninja-core-6.0.0-beta2.jar:na]
	at ninja.NinjaDefault.onRouteRequest(NinjaDefault.java:104) ~[ninja-core-6.0.0-beta2.jar:na]
	at ninja.undertow.NinjaUndertowHandler.handleRequest(NinjaUndertowHandler.java:56) [ninja-undertow-6.0.0-undertow1.jar:6.0.0-undertow1]
	at io.undertow.server.handlers.form.EagerFormParsingHandler.handleRequest(EagerFormParsingHandler.java:55) [undertow-core-1.4.8.Final.jar:1.4.8.Final]
	at io.undertow.server.Connectors.executeRootHandler(Connectors.java:210) [undertow-core-1.4.8.Final.jar:1.4.8.Final]
	at io.undertow.server.HttpServerExchange$1.run(HttpServerExchange.java:809) [undertow-core-1.4.8.Final.jar:1.4.8.Final]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [na:1.8.0_102]
[...]

PR #566, opened 2017-02-24
Fix issue #565

jjlauer:
Quick fix for NPE in very small number of routing cases.  Unit tests added.

### DIFF
--- a/ninja-core/src/main/java/ninja/Route.java
+++ b/ninja-core/src/main/java/ninja/Route.java
@@ -45,7 +45,7 @@ public class Route {
     private final String uri;
     private final Method controllerMethod;
     private final FilterChain filterChain;
-    private final LinkedHashMap<String,RouteParameter> parameters;
+    private final Map<String,RouteParameter> parameters;
     private final Pattern regex;
 
     public Route(String httpMethod,
@@ -87,7 +87,7 @@ public class Route {
         return controllerMethod;
     }
 
-    public LinkedHashMap<String,RouteParameter> getParameters() {
+    public Map<String,RouteParameter> getParameters() {
         return parameters;
     }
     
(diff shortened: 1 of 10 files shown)


ninja-core/src/main/java/ninja/Route.java
```diff
@@ -125,10 +123,12 @@ public boolean matches(String httpMethod, String uri) {
         Matcher m = regex.matcher(uri);
 
         if (m.matches()) {
-            Iterator<String> it = this.parameters.keySet().iterator();
-            for (int i = 1; i < m.groupCount() + 1; i++) {
-                String parameterName = it.next();
-                map.put(parameterName, m.group(i));
+            if (this.parameters != null) {
```
raphaelbauer: Another idea. What about returning an empty LinkedList when RouteParameter.parse(uri) instead returning null? Then we don't need any null checks... The test is still cool to make sure stuff works of course...

Not 100% sure but RouteParameter.parse(uri) is only called once when building the Route, so there should not be a performance impact of that...

ninja-core/src/site/markdown/team.md
```diff
@@ -10,10 +10,17 @@ new releases. And driving the platform as a whole.
 That leadership role is currently performed by:
 
  * The team of [FinalFrontierLabs](http://www.finalfrontierlabs.com) 
-   and [Raphael](http://raphaelbauer.com) as lead architect. 
+   and [Raphael](http://raphaelbauer.com) as lead architect.
+
  * The team of [Fizzed, Inc.](http://fizzed.com) and [Joe Lauer](http://lauer.bz)
    as lead architect.
 
+Ninja is proudly sponsored by <a href="https://www.greenback.com">Greenback</a>
```
raphaelbauer, line 18: That's 100% cool - it's great you and your team is part of the project! All the very very best for greenback.com!

raphaelbauer:
Tests don't succeed locally on my machine either (did not do any further debugging...)

jjlauer:
They ran locally for me, but I'll take a look.

jjlauer:
@raphaelbauer Went with your suggestion of just returning an empty map. Travis CI is happy now.

raphaelbauer:
Nice fix!

PR #573, opened 2017-04-30
Handle empty string param parsing

jlannoy:
For each parameter that is not a string, the parameter is checked against null and then against an empty string. In both case, the returned value for that parameter is null.

There is one exception, it's for the String parameters... An empty string (input) results in a empty string (parameter value). It could seems legit but I don't think it is. When an HTML request POST form fields that have not been filled, they appear in the request with... an empty value. Meaning that with the current version of Ninja, you have to check if your string is empty to know if user has entered a value in it. All my projects and my colleagues projects, are overriding the StringParamParser to avoid that.

There could be an side effect. Let's say we have... ?var=&var2=b 
With what I said just before, it means that the var param would be null. So you would not be able to make a difference between ?var=&var2=b and ?var2=b. But is it an issue? Does it make sense to explicitly want to pass an empty string to a parameter? I don't see any case where it could be a good use case.

I think it makes even more sense since with have Optional parameters for cleaner solutions. What do you think?

### DIFF
--- a/ninja-core/src/main/java/ninja/params/ParamParsers.java
+++ b/ninja-core/src/main/java/ninja/params/ParamParsers.java
@@ -371,7 +371,7 @@ public class ParamParsers {
     public static class StringParamParser implements ParamParser<String> {
         @Override
         public String parseParameter(String field, String parameterValue, Validation validation) {
-            if (parameterValue == null || validation.hasViolation(field)) {
+            if (parameterValue == null || parameterValue.isEmpty() || validation.hasViolation(field)) {
                 return null;
             } else {
                 return parameterValue;
--- a/ninja-core/src/site/markdown/developer/changelog.md
+++ b/ninja-core/src/site/markdown/developer/changelog.md
@@ -1,3 +1,5 @@
+ * 2017-04-30 Handling empty string param parsing (jlannoy)
+
 Version 6.0.0
 =============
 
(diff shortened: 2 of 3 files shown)


raphaelbauer:
Thanks for the PR :)

I'll cross post it on the mailing list - to make sure that people can comment. I don't yet have an opinion (and did not run into that issue). Give me some time to check some things out...

@jjlauer do your apps run into that problem?

raphaelbauer:
I am undecided. I've seen that other frameworks do it differently (and accept "empty" parameters as empty strings).

But I also don't care and my apps are not affected by that change. @jjlauer also got a valid point. So we can do it.

Before we can merge the change please do the following:
- Add your change to changelog.md and upgrade_guide.md.
- Update the documentation to highlight the parameter handling (Maybe a box with some notes).

Thanks for the nice contribution!

raphaelbauer:
For the record. That's @jjlauer 's comment from the mailing list:

> I agree that empty strings are as good as null strings from a
> parameter perspective.  I'm an advocate for making parameter parsing
> pretty strict.  It makes the new Optional concept better too since you
> can supply a default if the user didn't really provide one.

jlannoy:
@raphaelbauer And what about a configuration property to let users choose the way it work?

raphaelbauer:
Theoretically yes - but that always takes maintenance effort in the long run. So I don't like it too much. And nobody objected, so I guess it's safe to change the behaviour...

raphaelbauer:
Thanks @jlannoy for your contribution! Will be released asap :)

PR #575, opened 2017-05-22
explain duration parameter

toyg:
Format for the duration is not documented in javadocs, I had to read the source to figure it out. Hopefully this note makes it easy to understand.


raphaelbauer:
Very nice! Thanks!

PR #578, opened 2017-08-22
System properties (as well as prefixed system properties) can override config keys

jjlauer:
While some properties (e.g. ninja.mode or ninja.port) have always allowed system properties to be the ultimate override a value set in configuration files, others have not.  I was trying to override a value the other day on the command-line and realized it wasn't supported in ninja.  Many other web frameworks support this concept and I think it'd be a useful addition to ninja.


raphaelbauer:
Pretty cool!

PR #582, opened 2017-09-01
WebSockets!

jjlauer:
Long awaited WebSockets support for Ninja.  Includes Ninja's own handshake mechanism to get around JSR-356s design flaws and headaches (see stackoverflow :-)).  Works in standalone as well as tested against Jetty, Tomcat, Wildfly, and WebLogic.

Also includes auto servlet configuration (no more web.xml needed) and the reverse router can now build absolute urls (so you can build a nice reverse route to a websocket endpoint).


raphaelbauer:
this looks great...! Give me some days to read through the code and try some thing :)

ninja-core/src/main/java/ninja/RouterImpl.java
```diff
@@ -185,30 +188,44 @@ public void compileRoutes() {
...
-                if (this.reverseRoutes.containsKey(controllerMethodRef)) {
-                    // the first one wins with reverse routing so we ignore this route
-                } else {
-                    this.reverseRoutes.put(controllerMethodRef, route);
+        this.routes.forEach(route -> {
+            Class<?> methodClass = route.getControllerClass();
+            
+            String methodName = Optional.ofNullable(route.getControllerMethod())
+                .map(Method::getName)
+                .orElse("");
```
raphaelbauer: not sure if this is super-clean... returning ""... I guess it simply not be added to the this.reverseRoutes afaics... (I may be wrong though...)

ninja-core/src/main/java/ninja/RouterImpl.java
```diff
@@ -185,30 +188,44 @@ public void compileRoutes() {
         }
         List<Route> routesLocal = new ArrayList<>();
         
-        for (RouteBuilderImpl routeBuilder : allRouteBuilders) {
+        allRouteBuilders.forEach(routeBuilder -> {
```
raphaelbauer, line 191: Love for Java8 is in the air ;)

ninja-core/src/main/java/ninja/websockets/WebSocketUtils.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Utilities for implementing WebSockets.
+ * 
+ * @author jjlauer
+ */
+public class WebSocketUtils {
+    
+    // Attribute key for context storage of the endpoint instance to use
+    // for continuation of a websocket handshake
+    static public final String ATTRIBUTE_ENDPOINT = "ninja.websocket.endpoint";
```
raphaelbauer, line 32: I guess it should be public static (also for the other methods / variables here)...

ninja-core/src/main/java/ninja/websockets/WebSocketUtils.java
```diff
@@ -0,0 +1,58 @@
...
+    static public final String ATTRIBUTE_HANDSHAKE = "ninja.websocket.handshake";
+    
+    /**
+     * Parses header `Sec-WebSocket-Protocol: chat, superchat` into a list
+     * such as `chat`, and `superchat`.
+     * @param value The header value
+     * @return A set of values or null if value was null or empty
+     */
+    static public Set<String> parseProtocolRequestHeader(String value) {
```
raphaelbauer, line 41: Just an observation... might it be more cool to return Collections.emptySet()?... if we have the option to reduce null usage we should do so... imho helps in the long run... but - I may be wrong...

ninja-core/src/site/markdown/documentation/websockets.md
```diff
@@ -0,0 +1,111 @@
...
+
+You will want to make sure you do not mark your WebSocket class as a `@Singleton`.
+Unless you know what you are doing, its better to create a new instance for each
+WebSocket session that is created.
+
+Your controller/endpoint will be created using Guice -- so the `@Inject` annotation
+will be honored when instantiating your endpoint.
+
+## WebSocket handshake
+
```
raphaelbauer, line 69: that is pretty cool... very good explanation...!

ninja-servlet-integration-test/Dockerfile.jetty
```diff
@@ -0,0 +1,4 @@
+# 9.3.15, 9.4
```
raphaelbauer, line 1: (not sure if this comes later) but it may be beneficial having a README saying exactly for what these Dockerfiles can be used. I know it now... But I know myself as well an in 1 year from now I'll have completely forgotten about them and will wonder why these exist...

raphaelbauer:
Tiny: The file copyright headers should be uniform across the project. That should be fixed before merging into master (there is a profile in the main pom.xml to facilitate that..)

raphaelbauer:
Ok.. Some final remarks. I am really looking forward to using your ws implementation in one of my next projects. Pretty awesome work Joe. Clever upgrading and integration into Ninja's flow :) Good tests and well documented. Wow!

I added some tiny comments, but feel free to merge at any time :)

raphaelbauer:
👍

ninja-core/src/main/java/ninja/RouterImpl.java
jjlauer: This is partially a legacy of accepting static Results in old versions.  Those are now wrapped in functional lambdas -- and we can require a method name now.  Will clean this up.

ninja-core/src/main/java/ninja/websockets/WebSocketUtils.java
jjlauer: Yeah great point.  Will do that.

ninja-servlet-integration-test/Dockerfile.jetty
jjlauer: I debated with myself on not even including these on the PR.  However, it was pretty useful to test our .war against a few servlet impls. For example, I discovered Ninja's cookie handling using an equals char isn't supported by Tomcat & Wildfly.  I'll add a README to explain why we have them.  IMHO, I think it may be useful to rework this module to allow for testing it against a number of impls down the road.

jjlauer:
Will shoot to get this merged sometime this week.

jjlauer:
Oops accidental close.

jjlauer:
Travis CI is currently broken.  They are printing out "Picking up _JAVA_OPTIONS" to stdout on new processes and its breaking how we test the maven plugin.  Doesn't happen locally.  Will merge anyway.

raphaelbauer:
👍

Issue #595, opened 2017-12-05
Duplicate count for timed metric


christiangroth:
Due to the call of stop() and close() the timer is calculated twice as close() calls stop() internally. If you create a simple @timed annotated controller method you can see that a single request results in count=2 for this timer in corresponding mbean.
  
https://github.com/ninjaframework/ninja/blob/3a9f02bd32a3963f2dd8b0996e951253797a5456/ninja-metrics/src/main/java/ninja/metrics/TimedInterceptor.java#L56-L62

raphaelbauer:
Nice finding! Would you mind creating a PR for that?

christiangroth:
There you go: https://github.com/ninjaframework/ninja/pull/596 :)

PR #596, opened 2017-12-05
Ticket #595

christiangroth:
https://github.com/ninjaframework/ninja/issues/595

removed call to close as this also calls stop internally and leads to duplicate metric updates

### DIFF
--- a/ninja-core/src/site/markdown/developer/changelog.md
+++ b/ninja-core/src/site/markdown/developer/changelog.md
@@ -1,6 +1,7 @@
 Version 6.2.0
 =============
 
+ * 2017-12-13 Fixed duplicate count for timed metrics (christiangroth)
  * 2017-08-30 WebSockets support across Jetty standalone and any JSR-356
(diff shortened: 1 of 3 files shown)


raphaelbauer:
The tiny problem is that we might run into regressions later on. Imagine someone just adding that line accidentally again.
Therefore it would be great having a mocked test that tests that close is called only once (and stop never maybe).

Afaics there is no test yet for this class (which is ugly anyway). So you'd be the first and could greatly increase the stability :)

What I imagine is really just a test that uses mocks and makes sure close is called once (and stop never)...

Would you mind adding that? Let me know if you need any hints.. Thanks!

christiangroth:
Done! :)

christiangroth:
@raphaelbauer Any feedback on this one?

christiangroth:
Also updated changelog here

christiangroth:
@raphaelbauer: Any comments on this one? Would be great to have this also merged!

raphaelbauer:
Thanks for pinging! 👍

PR #600, opened 2018-01-08
Fix issue where servlet container without websocket support crashes.

raphaelbauer:
App Engine's Jetty for instance crashed when we do not first check
via Class.forName(...) whether websocket classes exist on the classpath.


raphaelbauer:
@jjlauer does that look ok? Had to add an additional test for servlet environments that do not support the websocket spec directly...

jjlauer:
Looks good.

raphaelbauer:
👍 thanks!

PR #603, opened 2018-01-29
Support for running a Ninja console (server-free) application

jjlauer:
In most applications I generally need to do something administrative -- and reuse my Ninja application code to do it.  For now I've always hacked it by firing up Ninja in a web server in a custom main() method.  Doesn't work so well if you already have a web server running.  This PR makes launching a console-only Ninja application part of the framework itself.  Open to naming it something other than `console` if there a better name for it.

### DIFF
--- a/ninja-core/src/main/java/ninja/standalone/NinjaConsole.java
+++ b/ninja-core/src/main/java/ninja/standalone/NinjaConsole.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2018 ninjaframework.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ninja.standalone;
+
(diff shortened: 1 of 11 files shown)


raphaelbauer:
That's cool! Give me some days to check out the code :)

ninja-core/src/main/java/ninja/standalone/NinjaConsole.java
```diff
@@ -0,0 +1,77 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ninja.standalone;
+
+import com.google.inject.Injector;
+import ninja.Bootstrap;
+import ninja.utils.NinjaPropertiesImpl;
+
+public class NinjaConsole extends AbstractConsole<NinjaConsole> {
```
raphaelbauer, line 22: There is no test for that class... Up to you if you want to add one...

raphaelbauer:
Do you have some real usecases for that? I cannot think of anything concrete right now...
Thanks!
Otherwise: Very nice :) feel free to merge anytime :)

👍

ninja-core/src/main/java/ninja/standalone/NinjaConsole.java
jjlauer: I did have a test, but named it NinjaStandaloneTest rather than what it was testing.  So I renamed it NinjaConsoleTest.

jjlauer:
@raphaelbauer - One real world use case is our app that indexes some of our data into elasticsearch.  We periodically rebuild that index for various reasons and need to do that in a console app.  It's much easier to leverage the same configuration & services that start up in our Ninja app to do the rebuild outside the context of a server.  We've hacked it our solution for now by starting up a Jetty server, but bind it to a different port, but honestly a NinjaConsole version of it makes way more sense.

PR #607, opened 2018-03-17
Make UUID a first class citizen

jjlauer:
Make a UUID a first class citizen for parameter parsing.  Invalid UUIDs will be rejected.


raphaelbauer:
nice 👍

PR #611, opened 2018-04-25
JPA and migrations are now a separate module.

raphaelbauer:
For people that use it not much changes. All others will have a lot
less dependencies when building Ninja apps.


raphaelbauer:
@jjlauer feedback appreciated. That's something I wanted to do for a long time - make the core of Ninja a bit smaller so that users can configure what they need... I am not using the JPA module anyway... so it reduces the waste for me :)

jjlauer:
@raphaelbauer - i am a massive fan of making ninja-core smaller and breaking things out into modules! In fact, I have a repo with used maven bill of materials for this "ninja-atom" and "ninja-lite" concepts we use in various projects.  Anything excluded in https://github.com/fizzed/ninja-plus/blob/master/ninja-bom-atom/pom.xml is probably a candidate to be moved into its own module (maybe).

Here are a few thoughts:

1. I use migrations but do not use jpa.  I think if we're going to do this, I would have put migrations-flyway into its own module and jpa into a different one.  Having a module correlate to the underlying dependency is perhaps a good rule of thumb IMHO.

2. I came up with the "ninja-classic" concept as I thought an eventual "ninja-classic" maven module would simply be a way for a user to get all the other modules easily.  So ninja-classic would pull in the classic ninja stack: ninja-core, ninja-migration-flyway, ninja-jpa, etc.  With that in mind, I would have called what you did simply "ninja-jpa" or "ninja-migration-flyway".  The `NinjaClassicModule` configuration class would only be available in ninja-classic, and it in turn would simply use guice modules that each other module owns.

raphaelbauer:
Thanks for the feedback!

Regarding 1)
I'd like to replace the "classic" approach (migrations+guice-jpa as unit) with https://github.com/ninjaframework/ninja-db.

This is cool because:
- It separates flyway from the rest (and you can use flyway right away)
- It allows to use more than database (and run eg migrations on more than one database)
- It allows to plug in any jdbc access library (ebean, jdbi, jpa etc pp) easily

Is this something you could live with? In your case we could simply add ebean as library to ninja-db and you are ready to go (flyway is already there...)

Regarding 2) ... maybe I'll wait first for your feedback regarding flyway. In general my feeling is that it would be easier to have some small breaking changes for the users (they'd have to add dependencies + install modules) - than us providing the classic module. And if I get it right then your apps also don't need the classic module...

jjlauer:
@raphaelbauer  I've looked at ninja-db a few times and i think it is definitely a more refined approach.  Yes, I could live with it, but last time I looked I think there were a few enhancements I needed before using it (can't remember what they are now). Would you want to keep that as a separate project, or move it in with this framework?

jjlauer:
Btw, caching is one of the only other big dependencies that would be perfect to extract out.  Maybe something like ninja-cache-memcached and ninja-cache-ehcache?

raphaelbauer:
True - ninja-db needs dome more polishing. Should we move it into the ninja project - or keep it separate - what's your opinion? When keeping it separate we can evolve it without increasing the ninja version. On the other hand it might be disconnected from the main framework easily... 

ninja-cache-memcached and ninja-cache-ehcache sounds like a plan... good candidates...

other candidate is the freemarker templating engine...

jjlauer:
@raphaelbauer With how critical datasources are in general to applications, I'd vote for including ninja-db in the mainline project -- along with key implementations like you did with Hikari, JDBI (maybe), and Flyway. If we did then I think you could fold the existing ninja-db-core into ninja-core since it only relies on jdbc packages.

Extracting out freemarker into its own module would be awesome.  I could finally merge ninja-rocker in a bit more easily :-)

raphaelbauer:
@jjlauer would you mind trying out ninja-db? I just released a new version to maven central (1.0.0-beta-7). There's now some documentation: https://github.com/ninjaframework/ninja-db

Just to be sure that it fits your usecase (Flyway)... Ideally I see the ebeans module also living there... Thanks!

Flyway (standalone) should work a config similar to:

```
application.datasource.default.driver=org.h2.Driver
application.datasource.default.url=jdbc:h2:mem:test
application.datasource.default.username=sa
application.datasource.default.password=

application.datasource.default.migration.enabled=true

```

Migrations should live in a directory called "migrations/default" then...

PR #614, opened 2018-04-26
Postoffice is now a separate dependency.

raphaelbauer:
Ninja-core got a more lean footprint - again.


raphaelbauer:
@jjlauer and a small refactoring again... feedback appreciated :)

jjlauer:
@raphaelbauer This looks pretty good.  Maybe a few files are questionable if really needed (in tests), but it may not be worth your time cleaning them up right now.  For example, the Filters example, much of the config data in application.conf, etc.  However, the main source of the module looks great.  I'm good with the PR if you are.

PR #625, opened 2018-08-23
Libraries upgrade + test classes fixed

jlannoy:
I'm about to release one or two modules for Ninja, and I noticed some libraries were rather old... I needed to locally update FluentLenium, Jackson, Guava... But I thought it was the right time to make a global update the framework.

So here is a consistant upgrade, with 2 tests classes fixed. Tests are green (in my environment). I tested the snapshot version on some project too, seems ok.

### DIFF
--- a/ninja-test-utilities/pom.xml
+++ b/ninja-test-utilities/pom.xml
@@ -67,12 +67,6 @@
             <scope>compile</scope>
         </dependency>
 
-        <dependency>
-            <groupId>com.google.code.gson</groupId>
-            <artifactId>gson</artifactId>
-            <scope>compile</scope>
-        </dependency>
-
         <!-- A very nice abstraction for making selenum tests: -->
         <dependency>
             <groupId>org.fluentlenium</groupId>
@@ -115,10 +109,10 @@
             <artifactId>doctester-core</artifactId>
             <scope>compile</scope>
             <exclusions>
-            	<exclusion>
-        			<groupId>com.google.guava</groupId>
-            		<artifactId>guava</artifactId>
-            	</exclusion>
+                <exclusion>
+                    <groupId>com.google.guava</groupId>
+                    <artifactId>guava</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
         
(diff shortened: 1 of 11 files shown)


jjlauer:
This PR looks good to me.  Let's let @raphaelbauer weigh in too.

ninja-test-utilities/pom.xml
```diff
@@ -108,6 +114,12 @@
             <groupId>org.doctester</groupId>
             <artifactId>doctester-core</artifactId>
             <scope>compile</scope>
+            <exclusions>
+            	<exclusion>
+        			<groupId>com.google.guava</groupId>
```
raphaelbauer: tiny.. but here are tabs..

ninja-test-utilities/src/main/java/ninja/NinjaFluentLeniumTest.java
```diff
@@ -20,15 +20,20 @@
...
 public abstract class NinjaFluentLeniumTest extends FluentTest {
 
-	public WebDriver webDriver = new HtmlUnitDriver();
+  /**
+   * @deprecated The driver should be access via {@link #getDriver()}.
+   * Null variable left for deprecation notice only.
+   * @see #getDriver()
+   */
+  @Deprecated
+	public WebDriver webDriver;
```
raphaelbauer: Already a mistake in the codebase... but would you mind fixing the tabs to space? thanks!

pom.xml
```diff
@@ -849,6 +849,13 @@
                 <scope>test</scope>
             </dependency>
 
+            <dependency>
+                <groupId>com.google.code.gson</groupId>
+                <artifactId>gson</artifactId>
```
raphaelbauer: why do we need gson now?

raphaelbauer:
Really cool! I added tiny remarks... otherwise we can merge asap :) 👍

jlannoy:
I fixed tabs to spaces. 

For Gson, it is used under the hood by doctester, fluentlenium and the htmlunitdriver too (this is why it was scoped as test). There was a problem with the version imported by doctester, and my exclusion tag was not working. But after a second check :  it was a really stupid mistake (I inverted groupId and artifactId). ... I was able to remove the Gson mention from the master pom ;)

raphaelbauer:
Awesome! Good work!

PR #627, opened 2018-09-01
Error handling : ForbiddenRequestException + i18n message

jlannoy:
Here is some work on the error management part.

- I added the **ForbiddenRequestException** type, along with the BadRequest and InternalError ; so that it's possible to fire a 403 error at any time, potentially with a root cause. Without this it's only possible to *return* forbidden results, which sometimes could lead to a strange code structure (when doing access check at lower levels).

- I added an error attribute in the Message POJO (used to display errors) that will handle the exception message, so that it can be displayed on the error page. (But not the default behavior.)

- Furthermore, this exception message can be a message key that will be taken from the messages.properties files. You can now throw an exception with an message key as message, to find it back translated in your error page, along with the generic error page text.

Doc updated to represent these changes. I don't see any backward incompatibility, as all parts are additions. Except for method signatures changed in the main Ninja interface, but it's not a big deal I think.

What do you think? :-)

### DIFF
--- a/ninja-core/src/main/java/ninja/Ninja.java
+++ b/ninja-core/src/main/java/ninja/Ninja.java
@@ -16,6 +16,10 @@
 
 package ninja;
 
+import ninja.exceptions.BadRequestException;
+import ninja.exceptions.ForbiddenRequestException;
+import ninja.exceptions.RenderingException;
+
 public interface Ninja {
 
 	/**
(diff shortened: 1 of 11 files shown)


raphaelbauer:
That's pretty cool! nice documentation and tests 👍 
@jjlauer any objections to merge it?

jjlauer:
I love this PR -- wanted this feature for awhile.  I've always wanted a NotFoundException in many cases too, but I'm okay with the PR as-is as well.

jlannoy:
@jjlauer you're right, it would be a best practice to use such an exception at some points.
So I added a RequestNotFoundException, exactly the same way I added the Forbidden one ;-)
Doc and tests updated.

raphaelbauer:
@jjlauer ok with you? Then I'd merge it and also cut a new release... Thanks!

PR #629, opened 2018-09-05
Empty String parser "shortcut"

jlannoy:
As it has been discussed twice (once in the PR, once in an open issue), I thought it was better to document (on the main site) the fact that empty string parameters would give null values by default... And providing a quick shortcut (2 lines now) to change that behavior.


raphaelbauer:
👍

PR #630, opened 2018-09-05
BodyParserEngine for multipart/form-data

jlannoy:
Quick addition that allow to parse a posted object (like the original BodyParserEnginePost, which I extended) but with uploaded files in the same request.

### DIFF
--- a/ninja-servlet-integration-test/src/test/java/controllers/UploadControllerAutoTest.java
+++ b/ninja-servlet-integration-test/src/test/java/controllers/UploadControllerAutoTest.java
@@ -17,34 +17,42 @@
 package controllers;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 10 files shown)


ninja-servlet-integration-test/src/test/java/controllers/UploadControllerAutoTest.java
```diff
@@ -17,34 +17,42 @@
...
-		String result = ninjaTestBrowser
-				.uploadFiles(getServerAddress() + "uploadFinishAuto",
-				        new String[] {"file","file","file2"},
-				        new File[] { file, file2, file2 } );
-
-		// compute excepted result
+
+        // Let's upload a simple txt file...
+        String result = ninjaTestBrowser.uploadFiles(
+                getServerAddress() + "uploadFinishAuto",
```
raphaelbauer, line 49: spaces. Thanks!

PR #631, opened 2018-09-08
Ninja's main site update (via markdown processor)

jlannoy:
By upgrading to the version 1.5 of the markdown processor (for the maven site plugin), it will allow to have dynamic page titles (better for search/index/tabs) along with an anchor for each subtitle.

This update was requiring some fixes, as the new version is more strict with the html special characters not in _code_ tags. Btw, I generated the site locally to check that it wasn't broking any existing layout or content.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -194,9 +194,9 @@
 
                     <dependencies>
                         <dependency>
-                            <groupId>net.ju-n.maven.doxia</groupId>
+                            <groupId>org.apache.maven.doxia</groupId>
                             <artifactId>doxia-module-markdown</artifactId>
-                            <version>1.0.0</version>
+                            <version>1.5</version>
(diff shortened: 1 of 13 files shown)


raphaelbauer:
This is extremely cool! It was very odd that the site did not create dynamic page titles :)

pom.xml
```diff
@@ -194,9 +194,9 @@
 
                     <dependencies>
                         <dependency>
-                            <groupId>net.ju-n.maven.doxia</groupId>
+                            <groupId>org.apache.maven.doxia</groupId>
                             <artifactId>doxia-module-markdown</artifactId>
-                            <version>1.0.0</version>
+                            <version>1.5</version>
```
raphaelbauer, line 199: pretty cool!

PR #632, opened 2018-09-08
Fix postoffice interface exceptions

jlannoy:
Replace PR #345, a quick type fix that have been forgotten since PR #290.


raphaelbauer:
This is something we did not want to do, because it would tie us to the implementation itself. So I'll decline it for now... An alternative would be to introduce something like a NinjaPostofficeException... but I don't think there's the need for it right now...

PR #634, opened 2018-09-09
Fix jetty context path config

jlannoy:
According to the documentation, the contextPath was misplaced, and anyone trying to change it would have no effect. https://www.eclipse.org/jetty/documentation/9.4.x/jetty-maven-plugin.html
Replace PR #276


raphaelbauer:
👍 really cool!

PR #647, opened 2019-01-19
Fixing JaxyRoutes Reflections usage

jlannoy:
I discovered in my last project, using JaxyRoutes, that the scan for annotated route was going through the whole application (including all classes from all libraries exploded in the shaded jar). Causing a slower boot time and also reflection warnings.

After looking down to the Reflections library, I found [that PR](https://github.com/ronmamo/reflections/issues/178) explaining exactly the problem while using `ClasspathHelper.forPackage()` method. This pull request applies the indicated work around to ninja-jaxy-routes.

To test it right, one Controller was added **not** in the `application.modules.package` but outside it, then checking that the route doesn't exist.

### DIFF
--- a/ninja-jaxy-routes/src/main/java/ninja/jaxy/JaxyRoutes.java
+++ b/ninja-jaxy-routes/src/main/java/ninja/jaxy/JaxyRoutes.java
@@ -219,9 +219,9 @@ public class JaxyRoutes implements ApplicationRoutes {
      * Configures the set of packages to scan for annotated controller methods.
      */
     private void configureReflections() {
-        Optional<String> basePackage = Optional.of(ninjaProperties.get(NinjaConstant.APPLICATION_MODULES_BASE_PACKAGE));
+        Optional<String> basePackage = Optional.ofNullable(ninjaProperties.get(NinjaConstant.APPLICATION_MODULES_BASE_PACKAGE));
         
-        if(basePackage.isPresent()) {
+        if (basePackage.isPresent()) {
             reflections = new Reflections(
                     basePackage.get() + "." + NinjaConstant.CONTROLLERS_DIR, 
                     new MethodAnnotationsScanner());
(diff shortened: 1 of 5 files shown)


ninja-jaxy-routes/src/main/java/ninja/jaxy/JaxyRoutes.java
```diff
@@ -217,14 +215,21 @@ public int compare(Method m1, Method m2) {
...
+     */
     private void configureReflections() {
-        ConfigurationBuilder builder = new ConfigurationBuilder();
-
-        Set<URL> packagesToScan = getPackagesToScanForRoutes();
-        builder.addUrls(packagesToScan);
-
-        builder.addScanners(new MethodAnnotationsScanner());
-        reflections = new Reflections(builder);
+        Optional<String> basePackage = Optional.of(ninjaProperties.get(NinjaConstant.APPLICATION_MODULES_BASE_PACKAGE));
```
raphaelbauer: I think it should be ofNullable... otherwise we'll get an exception if the property is not defined...

ninja-jaxy-routes/src/main/java/ninja/jaxy/JaxyRoutes.java
```diff
@@ -217,14 +215,21 @@ public int compare(Method m1, Method m2) {
...
+        
+        if(basePackage.isPresent()) {
```
raphaelbauer: tiny... space thing...  if(baseP => if (baseP

raphaelbauer:
just two very tiny comments... very nice and well tested!

raphaelbauer:
(I have to admit that I don't use JaxyRoutes personally though...)

jlannoy:
Fixed. It's the first time I use JaxyRoutes in production, on a smaller project (in terms of routes) ; this is how I discovered it. It produces about 30 warnings and a certain delay, trying to do reflections on the whole shaded JAR ;)

raphaelbauer:
Beautiful! 👍

PR #648, opened 2019-02-14
Optional custom http methods

jlannoy:
Following my last PR #647 ; restricting scan for methods was not sufficient. 

Jaxy Routes module also has a scanner used to look at custom HTTP methods annotations (if anyone needs more than the standard ones). But has there is no predicted package for theses annotations, the scanner looks through the whole application. It means that on the shaded JAR version of the application, the scanner go through a lot of classes, producing some warnings and time lost at boot time.

I think that the usage of custom http methods annotations is really a special use case, not needed in most case. It's even more true that this part is not covered by unit tests... And can not be ; as the doctester client can only execute requests with standard http methods.

To improve boot performance and avoid unnecessary warnings, I propose to set this Jaxy Route feature as optional, covered by a boolean configuration property.

### DIFF
--- a/ninja-core/src/site/markdown/documentation/basic_concepts/routing.md
+++ b/ninja-core/src/site/markdown/documentation/basic_concepts/routing.md
@@ -511,7 +511,7 @@ The following common HTTP method annotations are available:
 
 If the built-in methods are insufficient :
 
-1. Sets the key `ninja.jaxy.custom_http_methods` to `true` in your `conf/application.conf`.
+1. Set the key `ninja.jaxy.custom_http_methods` to `true` in your `conf/application.conf`.
 2. Then implement your own custom HTTP method:
 
     @Target(ElementType.METHOD)
--- a/ninja-core/src/site/markdown/documentation/basic_concepts/routing.md
+++ b/ninja-core/src/site/markdown/documentation/basic_concepts/routing.md
@@ -509,7 +509,10 @@ The following common HTTP method annotations are available:
 - `@POST`
 - `@PUT`
 
-If the built-in methods are insufficient, you may implement your own custom HTTP methods:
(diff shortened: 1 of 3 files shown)


ninja-jaxy-routes/src/main/java/ninja/jaxy/JaxyRoutes.java
```diff
@@ -55,6 +55,8 @@
 @Singleton
 public class JaxyRoutes implements ApplicationRoutes {
 
+    String NINJA_CUSTOM_HTTP_METHODS = "ninja.jaxy.custom_http_methods";
```
raphaelbauer: very minor... could be "final static"

ninja-core/src/site/markdown/documentation/basic_concepts/routing.md
```diff
@@ -509,7 +509,10 @@ The following common HTTP method annotations are available:
 - `@POST`
 - `@PUT`
 
-If the built-in methods are insufficient, you may implement your own custom HTTP methods:
```
raphaelbauer, line 512: Good documentation!

ninja-core/src/site/markdown/documentation/basic_concepts/routing.md
```diff
@@ -509,7 +509,10 @@ The following common HTTP method annotations are available:
...
+If the built-in methods are insufficient :
+
+1. Sets the key `ninja.jaxy.custom_http_methods` to `true` in your `conf/application.conf`.
```
raphaelbauer: very minor: "Sets" => "Set"

raphaelbauer (approved):
Very nice! Just some minor things you might want to fix...

jlannoy:
Fixed.

PR #650, opened 2019-02-15
Automatic CDN usage option

jlannoy:
Here is an option added to the webJarsAt implicit function allowing to use JSDeliver CDN in place of webjar dependency with only one boolean property.

All **classic** webjars (not bower ones) are hosted on public CDN js delivery with the same referring URL. I'm already using it on my projects, but it's quite simple to provide this as a base feature of Ninja (thanks to the webJarsAt implicit function).

https://www.webjars.org/documentation

### DIFF
--- a/ninja-core/src/main/java/ninja/template/TemplateEngineFreemarkerWebJarsAtMethod.java
+++ b/ninja-core/src/main/java/ninja/template/TemplateEngineFreemarkerWebJarsAtMethod.java
@@ -21,10 +21,13 @@ import java.util.List;
 
 import ninja.AssetsController;
 import ninja.Router;
+import ninja.utils.NinjaProperties;
 
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 
+import freemarker.template.SimpleScalar;
+import freemarker.template.TemplateBooleanModel;
 import freemarker.template.TemplateMethodModelEx;
 import freemarker.template.TemplateModel;
 import freemarker.template.TemplateModelException;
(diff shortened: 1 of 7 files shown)


ninja-core/src/main/java/ninja/template/TemplateEngineFreemarkerWebJarsAtMethod.java
```diff
@@ -33,28 +36,50 @@
...
-       argsWithControllerAndMethod.add(AssetsController.class.getName());
-       argsWithControllerAndMethod.add("serveWebJars");
-       argsWithControllerAndMethod.add("fileName");
-       argsWithControllerAndMethod.addAll(args);
+        if (args.size() == 0) {
+            throw new TemplateModelException("Error. You must specify a webjars asset URL.");
+        }
+
+        // Allow a boolean as second argument to force webjars usage
+        if (useCDN && (args.size() == 1 || !((TemplateBooleanModel) args.get(1)).getAsBoolean())) {
```
raphaelbauer, line 66: Very stupid question. But why does this work?
if I got ${webJarsAt("bootstrap/3.3.4/css/bootstrap.css", true)}

Then 
!((TemplateBooleanModel) args.get(1)).getAsBoolean()))

should evaluate to false - right?

jlannoy: Exactly. But this boolean means "force my local webjar even if configured for CDN". The contrary makes no sense using `webJarsAt`, as "force one CDN if configured with local webjar" will be achieved by not using `webJarsAt`.

So I wrote in the documentation :
_You can always pass <code>true</code> as a second argument of <code>webJarsAt</code> to force usage of a local webjar content._

raphaelbauer: I am actually having a hard time to get what true means. And when I imagine myself using that feature or reading a templating using it - then I know that I will likely have to look up the manual.

What about the following: Do we really need that feature? Or do you need that feature? Otherwise let's just remove it and add it again (maybe differently) when we really need it.

Having the option to globally define whether libraries are loaded from a CDN or locally seems already great to me.

What do you think?

jlannoy: I think it's a common need. In fact there is a problem with some "bower webjars" not available through CDN. That's some webjars automatically created from bower libraries (other webjars being "classic webjars"). They are useful for some little JS libs... So depending on what you use, you cannot switch **all** your webjars to their CDN version.

I completely agree with you on readability but I wasn't able to find a better way way to express it. Don't forget it's a specific case.
- An other option would be to add a new freemarker method for this feature, kind of "webJarCdnAt()", and only this one would do the trick. But I thought it was a better idea to keep only one method.
- A last option would be to document to force **not to** use webJatAt() when you want to always force the local webjar address. But as using a boolean, understanding why in a template one js or css inclusion is not done via webJarAt() would also require to look up the manual.

raphaelbauer: Ok. Got it. I also can't come up with a better idea for now. Let's use it that way. It's documented properly. We can also rewrite it later if it does not work.

Thanks for the explanation!

PR #652, opened 2019-03-01
Update ninja_fluentlenium_test.md

jlannoy:
I noticed the test example is not up to date with FluentLenium available methods from 3.X. Here is the right version.


raphaelbauer:
top!

PR #659, opened 2019-05-01
JPMS compatibility

zmeggyesi:
Related to issue #628 , I added JPMS-compatibility to Ninja. This should open up possibilities related to the module system such as pared-down runtime creation and more robust dependency management.

PR #661, opened 2019-05-03
Java 11 support

raphaelbauer:


raphaelbauer:
@jjlauer @jlannoy Support for Java 11 (minimum JDK version still is 1.8). Feedback appreciated... Thanks!

jjlauer:
@raphaelbauer Looks pretty good so far.  I know #659 attempts to add modules support, but I think your approach of just starting with Java 11 is good.

Issue #671, opened 2019-08-07
XSS in example blog application


fxb-cocacoding:
If you run the example blog application via

`mvn archetype:generate -DarchetypeGroupId=org.ninjaframework -DarchetypeArtifactId=ninja-servlet-jpa-blog-archetype`

then the example application has a persistent cross site scripting vulnerability.
You can reproduce this behavior if you create a new blog entry with

`<script>alert('xss')</script>`

raphaelbauer:
Wonderful finding! Would you mind fixing the archetype?

llenrup:
No HTML is being escaped after update to the latest version of freemarker as it changed various things . See my pull request for a way around this.

PR #672, opened 2019-09-11
Set default output format to HTML and remove escaping loader.

llenrup:
The way escaping is handled in Freemarker has changed.
This change removes a XSS vulnerability because HTML was not
being escaped.

This should fix issue XSS in example blog application #671


llenrup:
Added new configuration boolean **freemarker.oldversion** which defaults to false.
If false will use freemarker 2.3.28 which has incompatible changes.
You may wish to change around the default.
I have changed the jpa archetype to generate using new freemarker ( ie. uses ?no_esc rather than <#noescape> etc. )
I also added dependency to OWASp sanitiser in archetype then use it to santise the posted article,

raphaelbauer:
Hi @llenrup Thanks for the nice PR! I'll have a look now to merge it asap!

raphaelbauer:
Merged. Thanks for your contribution!

PR #675, opened 2019-09-24
Make splash logo customizable and/or hide it

### LINKED ISSUE #674: Option to disable splash screen
It would be cool to have an option to disable the splash screen. Yes it's cool and everything, but it looks a bit unprofessional when you are distributing an app.

Having something like `ninja.splashLogo=false` (with default value being `true`) would be nice to have. If you really don't like the idea, you could leave it undocumented; this way nobody would know it's there, but people wouldn't have to fork the whole repo just to remove that bit.

toyg:
Fixes #674

### DIFF
--- a/ninja-core/src/main/java/ninja/NinjaDefault.java
+++ b/ninja-core/src/main/java/ninja/NinjaDefault.java
@@ -145,9 +145,9 @@ public class NinjaDefault implements Ninja {
     
     @Override
     public void onFrameworkStart() {
-
-        showSplashScreenViaLogger();
-                
+        if (ninjaProperties.getBooleanWithDefault(NinjaConstant.NINJA_SPLASH_DISPLAY, true)) {
+            showSplashScreenViaLogger();
+        }
         lifecycleService.start();
     }
 
--- a/ninja-core/src/main/java/ninja/NinjaDefault.java
+++ b/ninja-core/src/main/java/ninja/NinjaDefault.java
@@ -16,12 +16,17 @@
 
 package ninja;
 
+import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


ninja-core/src/main/java/ninja/utils/NinjaConstant.java
```diff
@@ -127,6 +127,11 @@
      */
     final String serverName = "application.server.name";
 
+    /**
+     * Whether to display logo at startup. Defaults to true if missing.
+     */
+    final String NINJA_SPLASH_DISPLAY = "ninja.splash.display";
```
raphaelbauer, line 133: Very nice! One thing we need is documentation - otherwise nobody would know about your great feature... I think configuration_and_modes.md would be a good candidate to add some lines regarding ninja.splash.display and ninja/logo.txt

ninja-core/src/main/java/ninja/NinjaDefault.java
```diff
@@ -477,9 +482,16 @@ private final String readNinjaVersion() {
     private final void showSplashScreenViaLogger() {
         
         String ninjaVersion = readNinjaVersion();
-        
-        // log Ninja splash screen
-        logger.info(NINJA_LOGO, ninjaVersion);
+
+        // log Ninja splash screen, from resources if available (so it can be overridden)
+        String NINJA_LOGO_LOCATION = "ninja/logo.txt";
+        try (InputStream is = this.getClass().getClassLoader().getResourceAsStream(NINJA_LOGO_LOCATION);) {
```
raphaelbauer, line 488: Do you see a simple way to test it? To make sure the stuff keeps on working after we eg do some refactoring?

raphaelbauer:
And one tiny thing is to add one line to changelog.md. Thanks!

raphaelbauer:
Nice! Thanks for your contribution!

PR #692, opened 2021-01-30
Include query string parameters as part of multi-part form upload request

jjlauer:
Multi-part upload handling did not mirror how standard POST entities were handled for "Query String" parameters.  If you send a form post or json post, along with query string parameters, context.getParameters() would always return a merged view of any of those parameters.  On multi-part file uploads, however, any query string parameters were missing.  This patch fixes that.


raphaelbauer:
Really cool fix. 👍

PR #694, opened 2021-01-30
Rendering InputStreams directly

jjlauer:
Rendering InputStreams directly to a Result currently requires a Renderable. This is a helper method to make that process a bit simpler.

### DIFF
--- a/ninja-core/src/main/java/ninja/Result.java
+++ b/ninja-core/src/main/java/ninja/Result.java
@@ -38,6 +38,8 @@ import org.slf4j.LoggerFactory;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import com.google.common.io.ByteStreams;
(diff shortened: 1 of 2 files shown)


ninja-core/src/main/java/ninja/Result.java
```diff
@@ -442,6 +444,45 @@ public void render(Context context, Result result) {
...
+     * Will set the content type to application/octet-stream, unless the content-type
+     * was already set. If any exception is thrown during the render, and internal
+     * server error exception will be thrown.
+     * 
+     * @param input The input stream to read from
+     * @param close If true this method will close the input stream after copying
+     *      all the bytes, otherwise that is up to the caller.
+     * @return This result
+     */
+    public Result render(final InputStream input, boolean close) {
```
raphaelbauer, line 459: cool

raphaelbauer:
Actually I forgot one thing: changelog is not updated...

PR #698, opened 2021-02-08
Support for ninja-db / creating ninja-db-classic for legacy support

raphaelbauer:
work in progress


jjlauer:
I'm not as familiar with JPA.  Does flyway actually need JPA?  Just curious if there should be a flyway module and then a jpa module?

raphaelbauer:
Yea... No it does not need it. Completely separate things. Flyway is for migrations. JPA for the database stuff.
I just kept it together because it was like this before.

The future should be https://github.com/ninjaframework/ninja-db where everything is separated nicely.

raphaelbauer:
A first step to de-bloat ninja-core... Feels better...

PR #707, opened 2021-03-18
forbid passing null argument to TimeUtil.parseDuration()

asolntsev:
it was a cause of errors: developer forgot to add configuration property with cache lifetime, and objects were accidentally cached for 30 days!

I know this PR does change the behaviour of the framework, but I believe this change can save users from accidental errors. 

it's always safer to explicitly forbid null arguments.


raphaelbauer:
👍

PR #709, opened 2021-03-19
Cleanup tests

asolntsev:


raphaelbauer:
Nice! Could you add some remarks to changelog.md then we can merge it!

asolntsev:
@raphaelbauer Do we really need to describe these changes in changelog? It's just an internal refactoring which will not affect users of Ninja framework.

PR #711, opened 2021-04-01
Upgrade to hibernate 5

asolntsev:
based on another PR https://github.com/ninjaframework/ninja/pull/708 (branch `upgrade-dependencies`).

### DIFF
--- a/ninja-core/pom.xml
+++ b/ninja-core/pom.xml
@@ -1,3 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
 <!--
 
     Copyright (C) 2012-2020 the original author or authors.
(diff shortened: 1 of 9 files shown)


raphaelbauer:
Just fyi: Mid-Term Hibernate should no longer be part of Ninja, but of Ninja-Db. Ninja-Db living here: https://github.com/ninjaframework/ninja-db . It already got support for jdbi, and you could use that as boilerplate to add Hibernate...

asolntsev:
@raphaelbauer Wow, I didn't know that, thanks. 
But shouldn't we remove Hibernate dependency from https://github.com/ninjaframework/ninja then?

raphaelbauer:
That was roughly the plan for Ninja 7. Making it a lot smaller. We could then also remote Flyway and all db related dependencies from the core...

raphaelbauer:
Let me know if we can help in any way :)

PR #717, opened 2021-07-29
Use latest Flyway

llenrup:
Change to flyway 7.11 this allows later postgres databases.
It is configured differently so made changes.


ninja-db-classic/src/main/java/ninja/migrations/MigrationClassicModule.java
```diff
@@ -24,6 +26,7 @@
     @Override
     protected void configure() {
         bind(MigrationEngine.class).toProvider(MigrationEngineProvider.class);
+        //bind(FluentConfiguration.class).toProvider(FluentConfigurationProvider.class);
```
jjlauer: Probably just delete this line

jjlauer (changes requested):
Looks pretty good, useful contrib! I didn't check but are there any other classes that are now defunct and should be deleted from the repo if there's a new configuration method?

ninja-db-classic/src/main/java/ninja/migrations/MigrationClassicModule.java
llenrup: Done  ( I am used  to bitbucket rather than github, so when I check in on my branch does this fix this pull request?)

thibaultmeyer: Hello @llenrup, do you think you have time to finish the last corrections requested? If you don't have time, can I take over your merge request to finish it?

@jjlauer

llenrup: Yes sorry go ahead,
Things been busy this year

thibaultmeyer: PR is open, is it possible to close this one?

@jjlauer @llenrup

jjlauer:
Upgraded flyway in a different PR.

PR #722, opened 2021-12-17
Update Flyway dependency to 8.2.2

thibaultmeyer:


jjlauer:
This looks pretty good to me. Just one question since I haven't done a lot of research. Are there any significant upgrade issues for applications that are using Flyway 5.x to get to 8.x?  For example, they had planned on deprecating some support for various things in mysql for flway after 5.x.  If there are links we could point people to if there are, perhaps we make the CHANGELOG more descriptive or include a link to that info.

thibaultmeyer:
Locally, I worked mostly with H2 and CockroachDB (PostgreSQL compatible). With this update of Flyway it is now possible to work with H2 version 1.x and 2.x.

The problem with MySQL seems to be in the licensing.  Since version 8.2.0, Flyway has removed the MySQL driver and replaced it with the MariaDB driver. To continue using MySQL, developer have to add a new dependency as specified in this page: https://flywaydb.org/documentation/database/mysql#java-usage.

thibaultmeyer:
Just updated `upgrade_guide.md` file. I added it in a 6.8.2 section, but maybe you want to introduce this change in a 6.9.x version instead?

PR #726, opened 2021-12-28
Fix "log.debug()" call (bad number of arguments)

thibaultmeyer:
This format call refers to 0 argument(s) but supplies 2 argument(s)


ninja-maven-plugin/src/main/java/ninja/build/WatchAndRestartMachine.java
```diff
@@ -219,7 +219,7 @@ public void handleNewOrModifiedFile(String newOrMod, Path path) {
         log.debug(" matched rule: type={}, pattern={}, proceed={}", match.type, match.pattern, match.proceed);
         
         if (match.proceed) {
-            log.debug(" will trigger restart", newOrMod, f);
```
thibaultmeyer, line 222: `newOrMod` and `f` are already logged in line 215 (`log.debug("{} file detected: {}", newOrMod, f);`).

jjlauer (approved):

PR #730, opened 2022-01-01
"assets" folder can be located either in "resources" or "src"

### LINKED ISSUE #712: make assets dir configurable
Currently assets dir is hardcoded: `src/main/java/assets`. Though, a best practice in Java project is storing such resources in `src/main/resources`. 

I suggest at least to make it configurable.

thibaultmeyer:
Resolves: #712

### DIFF
--- a/ninja-core/src/main/java/ninja/AssetsController.java
+++ b/ninja-core/src/main/java/ninja/AssetsController.java
@@ -23,6 +23,10 @@ import java.io.OutputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.nio.file.Files;
+import java.nio.file.LinkOption;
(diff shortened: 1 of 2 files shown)


asolntsev (approved):

raphaelbauer:
Nice one!

Could you think of a way to test this?
Otherwise someone might remove that functionality - and nobody notices that it breaks something for users...

thibaultmeyer:
I'm not sure how to unit test this change, it's a private method that uses hardcoded value "src/main" to find assets folder.

I see that at the time, you had the same problem with the existing test  `AssetsControllerTest::testStaticDirectoryIsFileSystemInDevMode`. It only tests the case where the requested file does not exist.

@raphaelbauer

raphaelbauer:
make sense..

PR #734, opened 2022-01-15
Raw access to the query parameter string

thibaultmeyer:
When dealing with external software, sometime we have to get the raw parameter string (query string) to work with.

This PR allow developer to get this data easily.

### DIFF
--- a/ninja-core/src/main/java/ninja/Context.java
+++ b/ninja-core/src/main/java/ninja/Context.java
@@ -108,6 +108,17 @@ public interface Context {
     String getRequestUri();
 
     /**
+     * Returns the parameter string that is contained in the request URL after the path.
+     * <br/>
+     * <b>Example:</b>
+     * For the URL <em>"http://domain.local/?key=private&hash=%93%af%f2%3b"</em>, this method
+     * will return <em>"key=private&hash=%93%af%f2%3b"</em>
+     *
+     * @return The parameter string
+     */
+    String getRequestParameterString();
+
+    /**
      * Returns the hostname as seen by the server.
      *
      * http://example.com/index would return "example.com".
--- a/ninja-servlet/src/main/java/ninja/servlet/NinjaServletContext.java
+++ b/ninja-servlet/src/main/java/ninja/servlet/NinjaServletContext.java
@@ -141,6 +141,11 @@ public class NinjaServletContext extends AbstractContext {
     }
 
     @Override
+    public String getRequestParameterString() {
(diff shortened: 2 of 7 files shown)


asolntsev (approved):

ninja-servlet/src/main/java/ninja/servlet/NinjaServletContext.java
```diff
@@ -140,6 +140,11 @@ public String getRequestUri() {
         return httpServletRequest.getRequestURI();
     }
 
+    @Override
+    public String getRequestParameterString() {
+        return httpServletRequest.getQueryString();
```
raphaelbauer, line 145: I think we should write a test that makes sure we do not break the functionality in the future... I am sure there is a way to test this :)

ninja-core/src/main/java/ninja/Context.java
```diff
@@ -107,6 +107,13 @@ private HTTP_STATUS(int code) {
     @Deprecated
     String getRequestUri();
 
+    /**
+     * Returns the parameter string that is contained in the request URL after the path.
+     *
+     * @return The parameter string
+     */
+    String getRequestParameterString();
```
raphaelbauer, line 119: Could you give an example in the Javadoc? Then it is even more straight-forward for everyone what this method does :)

raphaelbauer:
Looks good :) Just some tiny remarks :)

ninja-core/src/main/java/ninja/Context.java
thibaultmeyer: Done

ninja-servlet/src/main/java/ninja/servlet/NinjaServletContext.java
thibaultmeyer: Done

ninja-core/src/main/java/ninja/Context.java
```diff
@@ -107,6 +107,17 @@ private HTTP_STATUS(int code) {
     @Deprecated
     String getRequestUri();
 
+    /**
+     * Returns the parameter string that is contained in the request URL after the path.
+     * <br/>
+     * <b>Example:</b>
+     * For the URL <em>"http://domain.local/?key=private&hash=%93%af%f2%3b"</em>, this method
+     * will return <em>"key=private&hash=%93%af%f2%3b"</em>
```
raphaelbauer, line 115: 👍

raphaelbauer:
Nice!

